PR #616, opened 2025-12-09
feat: add machine QR codes for reporting

timothyfroehlich:
## Summary
- add server-side QR generation for machine reporting links using qrcode
- preselect machine on public report page via machineId and show QR source note
- surface downloadable QR card on machine detail with target URL

## Testing
- npm test -- src/app/report/default-machine.test.ts src/lib/machines/report-url.test.ts src/lib/machines/qr.test.ts
- npm run check
- npm run preflight


timothyfroehlich:
image
ugh

PR #662, opened 2025-12-25
chore: exclude dependabot from release notes check

timothyfroehlich:
This PR modifies the release notes check to skip for Dependabot PRs. It also includes minor formatting fixes across the repo to satisfy the pre-commit hooks.


timothyfroehlich:
@copilot modify this, and make it so that the release notes check only applies if there are changes to src files in the PR. Are there other files that should be included in release notes? Also address the prettier formatting issue.

PR #886, opened 2026-01-30
feat(notifications): Notify users of their own actions by default

### LINKED ISSUE #849: UX: Ensure users receive notifications for their own actions
A user requested that they receive email notifications for all changes they make to issues (comments, status changes, assignments, etc.) to serve as a record/reminder in their inbox. Currently, this is inconsistent across the application.

### Goal
Ensure that the user who performs an action (the "actor") is included in the notification recipient list by default.

### Root Cause
In `src/services/issues.ts`, the `addIssueComment` function is missing the `includeActor: true` flag when calling `createNotification`. Other services include it, but the default behavior of the notification system is to exclude the actor unless explicitly told otherwise.

### Proposed Fix
1. **Consistency**: Add `includeActor: true` to the `addIssueComment` service call.
2. **System-wide Default**: Consider modifying `createNotification` in `src/lib/notifications.ts` to include the actor by default, rather than requiring an explicit flag in every service call. This ensures that any new actions added in the future will automatically follow this behavior.

### Relevant Files
- `src/services/issues.ts`: Locations where `createNotification` is called.
- `src/lib/notifications.ts`: The logic that handles recipient filtering.

timothyfroehlich:
This change ensures that users receive email notifications for all changes they make to issues, providing a consistent record of their actions in their inbox. The `createNotification` function in `src/lib/notifications.ts` has been updated to include the actor in the recipient list by default. The redundant `includeActor: true` flags have been removed from the calls to `createNotification` in `src/services/issues.ts`, and the corresponding tests have been updated.

Fixes #849

---
*PR created automatically by Jules for task [6309756278702416327](https://jules.google.com/task/6309756278702416327) started by @timothyfroehlich*


timothyfroehlich:
### Code review - Correction

My earlier review incorrectly flagged the `new_comment` notification behavior as a bug. After reviewing issue #849 more carefully, I understand this is **intentional behavior** - users requested notifications for their own actions as a personal record/reminder.

The PR correctly:
- Changes `includeActor` default to `true` so users get notified of their own actions
- Removes redundant `includeActor: true` flags from call sites

**Test fixes needed:** The integration tests were written for the old behavior (exclude actor by default). I'm pushing a fix that adds explicit `includeActor: false` to tests that specifically test actor exclusion or recipient preference behavior, preserving their original test intent.

Also fixed: `database-queries.test.ts` had an invalid UUID (`"some-other-user-id"`) that was exposed by the new default behavior.

PR #909, opened 2026-02-02
feat(notifications): notify users of their own actions by default

### LINKED ISSUE #849: UX: Ensure users receive notifications for their own actions

timothyfroehlich:
## Summary

Users requested notifications for their own actions as a personal record/reminder in their inbox. This changes the default behavior of `createNotification` to include the actor in the recipient list.

- Change `includeActor` default from `undefined` to `true`
- Remove redundant `includeActor: true` flags from call sites
- Update integration tests to explicitly test actor exclusion scenarios
- Fix pre-existing invalid UUID bug in database-queries.test.ts

Fixes #849

## Test plan

- [x] Unit tests pass
- [x] Integration tests pass
- [ ] CI passes
- [ ] Manual verification: Create a comment and verify you receive a notification for it

### DIFF
--- a/src/test/integration/notifications.test.ts
+++ b/src/test/integration/notifications.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
-import { eq } from "drizzle-orm";
+import { eq, sql } from "drizzle-orm";
 import { createNotification } from "~/lib/notifications";
 import { sendEmail } from "~/lib/email/client";
 import { getTestDb, setupTestDb } from "~/test/setup/pglite";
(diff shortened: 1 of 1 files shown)


timothyfroehlich:
Closing - implementation needs rework

PR #1161, opened 2026-04-08
fix: lazy-import db in test-data cleanup route to prevent build-time crash

timothyfroehlich:
`/api/test-data/cleanup` imported `db` at module scope, causing `next build` to throw when `POSTGRES_URL` is unset. The `dynamic = "force-dynamic"` export added in a prior commit prevented static rendering but didn't stop the module from being loaded — `src/server/db/index.ts` throws during initialization regardless.

## Changes

- **Lazy `db` import** — replaced top-level `import { db }` with `const { db } = await import("~/server/db")` inside `POST()`, placed after the `NODE_ENV` production guard. Mirrors the pattern in `test-setup/route.ts`.
- **Import order fix** — moved `export const dynamic = "force-dynamic"` below the import block to match sibling routes (`client-logs`, `cron/cleanup-blobs`) and avoid lint violations.

```ts
export async function POST(request: Request): Promise<Response> {
  if (process.env.NODE_ENV === "production") {
    return NextResponse.json({ error: "Forbidden" }, { status: 403 });
  }

  // Lazy import to avoid loading db module during build
  const { db } = await import("~/server/db");
  // ...
}
```


timothyfroehlich:
@copilot apply changes based on the comments in [this thread](https://github.com/timothyfroehlich/PinPoint/pull/1161#pullrequestreview-4072518422)

PR #1399, opened 2026-05-21
fix(machines): replace hover-only tooltip with visible disabled reason text (PP-kqbk.7)

timothyfroehlich:
## Summary

- Removed the Radix `<Tooltip>` + `<span tabIndex={0}>` wrapper from `EditButtonWithTooltip` — Radix tooltips fire only on hover/focus, not touch, so mobile users had no way to discover why the Edit Machine button was disabled (CORE-A11Y-005, WCAG 1.4.13).
- Replaced with a visible muted text line (`<p className="text-xs text-muted-foreground">`) below the disabled button, surfacing the permission-denied reason directly. The reason strings from `getPermissionDeniedReason` are short (~30–45 chars) and well-suited for inline display.
- Linked the button to the reason paragraph via `aria-describedby` / `id` for assistive technology support. Dropped the `"use client"` directive since no client-only APIs remain.

## Test Plan

- [ ] On a touch device (or Chrome DevTools mobile emulation), load `/m/[initials]` as a member user who does not own the machine — confirm "Only the owner can perform this action" is visible below the disabled Edit Machine button without needing to tap/hover.
- [ ] Keyboard: tab to the disabled Edit button — confirm screen reader announces both the button label and the linked reason text via `aria-describedby`.
- [ ] As a guest user, confirm "Members can perform this action" displays.
- [ ] As a machine owner or admin, confirm no disabled button or reason text is shown (the enabled `EditMachineDialog` renders instead).
- [ ] `pnpm run check` passes.

## Related Issues

Closes PP-kqbk.7

—Antigravity-DisabledTooltip


timothyfroehlich:
Eval result: PR #1411 (—Claude-Reveal) selected as the .7 winner. Both PRs converged on the same architectural choice — drop Tooltip, inline visible reason text + aria-describedby — which validates that the approach is correct. Differentiators:

- #1411 adds a JSDoc explainer block clarifying why the file is named "edit-button-tooltip.tsx" but no longer uses a tooltip. This is load-bearing context for future readers (without it, the filename + missing Tooltip is confusing).
- #1411 centers the reason text under the full-width button (text-center), which matches the visual rhythm better.
- #1399 adds explicit aria-disabled="true" on the Button, which is good practice but redundant when the native disabled attribute is already present (shadcn Button uses native disabled, which sets implicit aria-disabled).

Close call — quality work on both. The JSDoc was the meaningful differentiator. Thanks for the duel; this confirms the approach. —Claude-MWG

PR #1633, opened 2026-07-10
fix(ci): pin supabase CLI version and fix setup-cli token input (PP-d5zn)

timothyfroehlich:
## What

The `Verify Migrations` job — and every other `supabase/setup-cli` call site (9 total) — used `version: latest`, which resolves the newest release via a per-run GitHub API call. That call flakes non-deterministically and false-reds `CI Gate`. It hit **main** twice recently (run `29058831201`; PR #1598 run `28714316725`, which passed on a no-change re-run).

## Fix (two vectors, one fix each)

1. **`version: latest` → pinned `2.109.1`** (current stable, published 2026-07-07). A concrete version makes `supabase/setup-cli` skip `resolveLatestVersion()` entirely and build the download URL directly (`releases/download/v${version}/...`), removing the per-run API resolution. Since `latest` already resolves to `2.109.1` today, this is a no-behavior-change pin. Documented inline (`# Pinned Supabase CLI — bump deliberately`) so future bumps are deliberate.
2. **Bogus token input renamed** — `ci.yml` passed `SUPABASE_CLI_GITHUB_TOKEN`, which `setup-cli@v2.1.1` does **not** accept (valid inputs: `version`, `github-token`). CI logged `Unexpected input(s)` and resolved unauthenticated (60 req/hr/IP). Renamed to `github-token: ${{ github.token }}`.

## Scope

- **`ci.yml`** (6 sites): version pin + token rename.
- **`preview-control.yaml` / `preview-sync.yaml` / `preview-reaper.yaml`** (1 each): version pin only — these never had the bogus token input.
- Retry-wrapper and CLI caching are **explicitly deferred** per PP-d5zn (token-rename + version-pin only).

## Verification

`pnpm run check` (actionlint + yamllint) passes. This is a CI-workflow-only change — no app code.

Bead: PP-d5zn

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -497,8 +497,9 @@ jobs:
       - name: Setup Supabase CLI
         uses: supabase/setup-cli@3c2f5e2ae34c34e428e8e206e2c4d21fa2d20fbf  # ratchet:supabase/setup-cli@v2.1.1
         with:
-          version: latest
-          SUPABASE_CLI_GITHUB_TOKEN: ${{ github.token }}
+          # Pinned Supabase CLI — bump deliberately (was: latest). PP-d5zn.
+          version: 2.109.1
+          github-token: ${{ github.token }}
 
       - name: Setup Supabase
         uses: ./.github/actions/setup-supabase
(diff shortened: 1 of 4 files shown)


timothyfroehlich:
Claude review of head b1078e2 — Self-reviewed workflow-only diff: pinned supabase/setup-cli to 2.109.1 (current stable, no behavior change vs latest) and renamed bogus SUPABASE_CLI_GITHUB_TOKEN -> github-token across 9 call sites. Verified against action.yml/main.ts: concrete version skips resolveLatestVersion API call; github-token is the real input. check (actionlint/yamllint) green. Retry+caching deferred per bead.

PR #1652, opened 2026-07-12
fix(lint): wire eslint-plugin-react-hooks, drop dead eslint-config-next (PP-k6jp)

timothyfroehlich:
## PP-k6jp — eslint-plugin-react-hooks not wired

`eslint-config-next` was installed as a devDependency but **never loaded** in `eslint.config.mjs`, and the lint script is a bare `eslint src/ --quiet`. Result: `react-hooks/exhaustive-deps` and `react-hooks/rules-of-hooks` ran **nowhere** — stale-closure bugs, missing effect deps, and conditional-hook-call bugs in client components went completely unlinted.

## What changed

- **Wired `eslint-plugin-react-hooks` directly** into the flat config (its own plugin block, **not** via the legacy next config), so we own the severities.
  - Scoped to app source under `src/` — both `.ts` custom hooks (9 exist, e.g. `src/hooks/*.ts`, `src/components/machines/settings/use-*.ts`) and `.tsx` client components. Tests/e2e/`src/test/**` excluded (hook-render helpers there produce noise with no user impact).
  - `react-hooks/rules-of-hooks`: **error**
  - `react-hooks/exhaustive-deps`: **error** (hard gate)
  - Branched from current `origin/main` (post-#1648), so the block sits alongside the `projectService: true` parserOptions — no per-block `parserOptions.project` reintroduced.
- **Removed `eslint-config-next` from devDependencies.** Verified nothing else referenced it: `rg -l eslint-config-next` (excluding node_modules/lockfile) showed only `package.json`; no `.eslintrc*` files exist; `next.config.ts` has no eslint reference. It shipped dead weight (jsx-a11y/import plugins we never load). jsx-a11y will be added directly in a separate future bead. The lockfile shrinks accordingly (its transitive plugin tree drops out).

## Violation strategy

Per triage guidance, this PR's job is to **enable the rules cleanly**, not fix the backlog. No dependency-array edits (those change runtime behavior).

- **rules-of-hooks**: **0 violations** across all of `src/`. Nothing to fix or grandfather.
- **exhaustive-deps**: **2 violations, both on a single `useEffect`** in the vendored shadcn calendar (`src/components/ui/calendar.tsx:193` — `[modifiers["focused"]]` complex-expression + missing `modifiers` dep). **Grandfathered** inline with `// eslint-disable-next-line react-hooks/exhaustive-deps -- PP-k6jp: pre-existing, triage separately` rather than touching the dep array.

**Grandfather count: 1 disable comment** (covering the 2 co-located errors).
Burndown dashboard: `rg -c "eslint-disable.*react-hooks" src/`

Because the beads backend is frozen, the grandfather comment uses the literal `PP-k6jp` pointer (no new bead created).

## Verification

- Confirmed the rule genuinely matches broadly (not silently skipping files): temporary probe files in both `src/hooks/*.ts` and `src/components/*.tsx` with deliberate missing deps were both flagged, then removed. No parse errors across `src/`.
- `pnpm run check` green (typecheck, lint, tests 1457 passed, format, linters, pytest).

## Net effect

Config wiring + `eslint-config-next` removal + 1 grandfather comment. No behavioral changes.


timothyfroehlich:
Claude review of head c1c6e62 — Reviewed the full diff: react-hooks wired directly (rules-of-hooks + exhaustive-deps at error, scoped src/**/*.{ts,tsx}, tests excluded), single pre-existing violation grandfathered in calendar.tsx with -- PP-k6jp pointer, no dependency-array edits, eslint-config-next→eslint-plugin-react-hooks (−1916 dead lockfile lines). Head c1c6e627 net diff vs main is only these 4 files; clean main-sync merge, CI green. LGTM. —Claude-TS7Footing

PR #1679, opened 2026-07-16
feat(pbm): shared foundation — pinballmap_state + store-and-heal lmx (PP-o355.16)

timothyfroehlich:
## What

Shared DB + read-path foundation both **PP-o355.11** (inbound cron sync) and **PP-o355.12** (outbound list/unlist) build on. Landing it first, as its own PR, lets those two proceed **in parallel adding zero schema** — no migration collision.

Ports the `pinballmap_state` singleton + `syncLocationSnapshot` from the unmerged reference commit `b6eb7dca`, **drops** all cron/status/action/UI, and **adds** net-new store-and-heal lmx columns that `b6eb7dca` didn't have.

## Changes

- **Migration 0052** (fresh — chain was at 0051; never re-applies `b6eb7dca`'s 0046):
  - `pinballmap_state` singleton table (id-singleton CHECK, `snapshot_json` jsonb typed `LocationSnapshot`, sync health `unknown|ok|error` CHECK, outbound creds refs, mirrors the `discord_integration_config` pattern).
  - `machines.pinballmap_lmx_id` (nullable) — the durable captured PBM listing handle.
  - Two machines CHECKs: lmx ⇒ catalog link; lmx ⇒ `pinballmap_listed`.
  - Partial `UNIQUE(pinballmap_machine_id) WHERE pinballmap_listed` — one PBM lister per catalog title (PbmApiAudit finding #1).
- **`src/lib/pinballmap/state.ts`** — `getPinballMapState()` + `syncLocationSnapshot()`. Fetch runs **outside any transaction** (CORE-ARCH-011); never throws on PBM failure (records health, returns `{ ok: false }`).
- **Type export** `PinballmapState` / `NewPinballmapState`.
- **Comment fix** in `types.ts` — the stale "ephemeral lmx, never cache" `PbmLmx` doc contradicted the locked store-and-heal model; rewritten.
- **Integration test** (PGlite) — upsert + persist + error path, PBM pinned to the mock at the seam (CORE-TEST-006).

## Explicitly out (→ other beads)

- **PP-o355.11:** cron route, hourly schedule, `status.ts` derivation, desync UI.
- **PP-o355.12:** `pinballmap.sync` permission, "Sync now" action, vault token write, connect/link/list/unlist actions, form UI.
- **PP-h059:** backfilling existing machines' lmx from the snapshot.

## Testing

- `pnpm run preflight:unlocked` — typecheck, lint, unit, build, and **both integration suites (PGlite + Supabase) green**.
- Local E2E `smoke` hit the known fresh-worktree degraded-auth-stack issue (`auth.setup` can't log in — documented 2026-07-13, not a code regression; this PR touches no auth/middleware/UI). CI runs the authoritative E2E.

Spec: `docs/superpowers/specs/2026-07-16-pbm-foundation-design.md`
Plan: `docs/superpowers/plans/2026-07-16-pbm-foundation-PP-o355.16.md`

### DIFF
--- a/src/lib/pinballmap/state.ts
+++ b/src/lib/pinballmap/state.ts
@@ -1,0 +1,102 @@
+import "server-only";
+import { eq } from "drizzle-orm";
+import { db } from "~/server/db";
+import { pinballmapState } from "~/server/db/schema";
+import { getPinballMapClient } from "./client";
+import { APC_LOCATION_ID } from "./config";
+import type { PinballmapState } from "~/lib/types/database";
+
+/**
+ * PinballMap location-snapshot read path (foundation — PP-o355.16).
+ *
+ * The integration keeps one `pinballmap_state` singleton row. A sync fetches our
+ * location's full JSON through the client seam and stores the WHOLE snapshot, so
+ * every downstream surface (status card, desync view, link/verify) reads the
+ * stored snapshot rather than hitting PBM per request — PBM's "one call per hour"
+ * conduct (CORE-PBM-001). The fetch is a side effect performed OUTSIDE any
+ * transaction (CORE-ARCH-011); we persist the result after it returns.
+ *
+ * PP-o355.11 schedules `syncLocationSnapshot` on a cron; PP-o355.12 reuses the
+ * persisted snapshot to resolve/verify lmx handles.
+ */
+
+const SINGLETON_ID = "singleton";
+
+/** Read the integration-state singleton (null when never initialized). */
+export async function getPinballMapState(): Promise<PinballmapState | null> {
+  const [row] = await db
+    .select()
+    .from(pinballmapState)
+    .where(eq(pinballmapState.id, SINGLETON_ID))
+    .limit(1);
+  return row ?? null;
+}
+
+/** Outcome of a sync attempt. */
+export type SyncResult =
+  | { ok: true; machineCount: number; syncedAt: Date }
+  | { ok: false; error: string };
+
+/**
+ * Fetch the configured location's snapshot from PBM and store it whole, updating
+ * sync health. Never throws on a PBM/network failure: it records the error on the
+ * singleton and returns `{ ok: false }` so callers (cron, "Sync now") can surface
+ * it without a 500.
+ */
+export async function syncLocationSnapshot(opts?: {
+  updatedBy?: string;
+}): Promise<SyncResult> {
+  const state = await getPinballMapState();
+  const locationId = state?.locationId ?? APC_LOCATION_ID;
+  const syncedAt = new Date();
+
+  try {
+    const snapshot = await getPinballMapClient().fetchLocation(locationId);
+    await db
+      .insert(pinballmapState)
+      .values({
+        id: SINGLETON_ID,
+        locationId,
+        snapshotJson: snapshot,
+        lastSyncedAt: syncedAt,
+        lastSyncStatus: "ok",
+        lastSyncError: null,
+        updatedAt: syncedAt,
+        ...(opts?.updatedBy ? { updatedBy: opts.updatedBy } : {}),
+      })
+      .onConflictDoUpdate({
+        target: pinballmapState.id,
+        set: {
+          snapshotJson: snapshot,
+          lastSyncedAt: syncedAt,
+          lastSyncStatus: "ok",
+          lastSyncError: null,
+          updatedAt: syncedAt,
+          ...(opts?.updatedBy ? { updatedBy: opts.updatedBy } : {}),
+        },
+      });
+    return { ok: true, machineCount: snapshot.machineCount, syncedAt };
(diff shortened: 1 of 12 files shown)


timothyfroehlich:
Claude review of head 28f3e3a — Addressed 3 review comments (lastSyncedAt=last-successful-sync + regression test; upsertState helper; documented enabled-gate ownership). Synced main. Copilot quota-blocked → Claude fallback. —Claude-PbmFoundation

src/lib/pinballmap/state.ts
```diff
@@ -0,0 +1,102 @@
...
+        locationId,
+        lastSyncedAt: syncedAt,
+        lastSyncStatus: "error",
+        lastSyncError: message,
+        updatedAt: syncedAt,
+      })
+      .onConflictDoUpdate({
+        target: pinballmapState.id,
+        set: {
+          lastSyncedAt: syncedAt,
```
timothyfroehlich: **`lastSyncedAt` is bumped on failed syncs, decoupling it from the stored snapshot.** On the error path this sets `lastSyncedAt = syncedAt` (the failed-attempt time) while `snapshotJson` is intentionally left stale. Anything downstream that computes snapshot freshness as `now - lastSyncedAt` (the PP-o355.11 status card / desync badge) will report the data as fresh right after a failed fetch, masking that the snapshot is actually hours old and the last real sync errored.

If `lastSyncedAt` is meant to be "last **successful** sync", drop it from the error-path `values`/`set` (leave it untouched on failure). If it is genuinely "last attempt", the .11 consumers must also read `lastSyncStatus` — worth a one-line comment pinning the intended semantic. —Claude-PbmPM

src/lib/pinballmap/state.ts
```diff
@@ -0,0 +1,102 @@
...
+        id: SINGLETON_ID,
+        locationId,
+        snapshotJson: snapshot,
+        lastSyncedAt: syncedAt,
+        lastSyncStatus: "ok",
+        lastSyncError: null,
+        updatedAt: syncedAt,
+        ...(opts?.updatedBy ? { updatedBy: opts.updatedBy } : {}),
+      })
+      .onConflictDoUpdate({
```
timothyfroehlich: **Minor (cleanup):** the success and error branches duplicate the whole `insert(...).onConflictDoUpdate(...)` scaffold plus the `...(opts?.updatedBy ? { updatedBy } : {})` spread. A single local `upsertState(fields)` helper taking the differing health fields would halve this and keep the two paths from drifting. Not blocking. —Claude-PbmPM

src/lib/pinballmap/state.ts
```diff
@@ -0,0 +1,102 @@
...
+ * singleton and returns `{ ok: false }` so callers (cron, "Sync now") can surface
+ * it without a 500.
+ */
+export async function syncLocationSnapshot(opts?: {
+  updatedBy?: string;
+}): Promise<SyncResult> {
+  const state = await getPinballMapState();
+  const locationId = state?.locationId ?? APC_LOCATION_ID;
+  const syncedAt = new Date();
+
```
timothyfroehlich, line 76: **Altitude:** `syncLocationSnapshot` reads `state.enabled` but never checks it — it fetches PBM unconditionally. Fine if every caller gates on `enabled`, but the .11 cron will be the first caller and this shared entrypoint is where a "disabled ⇒ skip the network call" guard belongs (CORE-PBM-001). Consider short-circuiting when `state?.enabled === false`, or documenting that the caller owns the gate. —Claude-PbmPM

src/lib/pinballmap/state.ts
timothyfroehlich: Fixed. `lastSyncedAt` is now written only on the ok path (removed from the error path), so it means "last successful sync" and the .11 freshness math can't be fooled by a failed attempt over a stale snapshot. Documented the semantic on `syncLocationSnapshot` and added a regression test (failed sync after a success preserves `lastSyncedAt` + snapshot; `lastSyncStatus` flips to error). —Claude-PbmFoundation

src/lib/pinballmap/state.ts
timothyfroehlich: Done — extracted a local `upsertState(fields)` helper; both paths now call it with only their differing health fields, so they can't drift and the `updatedBy` spread lives in one place. —Claude-PbmFoundation

src/lib/pinballmap/state.ts
timothyfroehlich: Documented rather than guarded: this bead scopes cron/enable gating to PP-o355.11, and short-circuiting on `enabled===false` here would make the foundation refuse to sync by default (the singleton defaults `enabled=false`). Added a doc note on `syncLocationSnapshot` that it's the pure mechanism and the caller (the .11 cron) owns the enabled/connection gate. —Claude-PbmFoundation

PR #1680, opened 2026-07-16
fix(issues): filter chips to own row on mobile + real overflow detector

timothyfroehlich:
## Summary

- **Bug fix:** active-filter chips were absolutely positioned over the search input and spilled off-screen on narrow viewports. Moved onto their own `flex-wrap` row below the input at all viewports, with an always-visible remove (✕) button (was hover-only → unusable on touch). Reported by Tim on a collection issues page.
- **Latent a11y fix:** chip labels used the severity/priority accent color as text over the chip background (WCAG contrast fail, e.g. light-purple "High"). Accent now lives on the icon only; label is neutral `text-foreground`. Surfaced by rendering active chips through axe for the first time.
- **Overflow detector rewritten:** the old `document.scrollWidth` check was vacuous — the app shell clips at `<body>` (`overflow-x:hidden`), so the document can never exceed the viewport. It now walks the DOM for visible content a non-scrollable ancestor clips off-screen (32px threshold). Added filters-active route variants to `responsive-overflow.spec.ts` (empty-state routes never rendered chips) and strengthened **CORE-RESP-004** to require loaded-state coverage.

## Test Plan

- [ ] CI green (full E2E suite — this is the clean run the local box couldn't produce under memory pressure)
- [ ] `responsive-overflow.spec.ts` exercises the filters-active state and passes
- [ ] Chips render below the search input, wrap, and the ✕ is tappable on mobile widths

## Related Issues

PP-e6ky. Follow-up filed: PP-bu9e (pre-existing timeline timestamp clip).

### DIFF
--- a/docs/NON_NEGOTIABLES.md
+++ b/docs/NON_NEGOTIABLES.md
@@ -449,9 +449,9 @@ trigger: always_on
 **CORE-RESP-004:** Overflow testing required
 
 - **Severity:** High
-- **Why:** Horizontal overflow is invisible to Playwright visibility assertions but breaks the user experience
-- **Do:** Add route to `e2e/smoke/responsive-overflow.spec.ts` when creating a new page
-- **Don't:** Ship a new page without overflow coverage at both mobile and desktop viewports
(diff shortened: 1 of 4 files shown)


timothyfroehlich:
Claude review of head 025574b — Claude fallback review (Copilot quota-limited): clean PR, 2 low-severity non-blocking notes — redundant visibleBadges alias; overflow-detector false-positive risk for future off-screen-transform elements (skip-link/off-canvas). No correctness bugs; full E2E green incl. new loaded-state chip coverage.

PR #1684, opened 2026-07-17
feat(collections): Wave 0b — view-link sharing (view_token) (PP-wqit.2)

timothyfroehlich:
## Summary

- Adds **"anyone with the link can view"** sharing to collections — the tournament-shareable slice of the epic (PP-wqit.2). A collection owner flips a toggle, gets a link, and anyone (no sign-in) can see the read-only Overview/Issues/Timeline.
- **Edit sharing is intentionally split out** to account-based collaborators (PP-wqit.7) — there is no edit token. Rationale below.

## Design decisions

- **View link is a rotatable `view_token`, not the collection id.** Unlike Google Docs (which uses the doc id itself as the capability + a server-side flag), we mint a separate nullable/unique token so the id stays a non-secret internal identifier and "reset link" gives real revocation. A single view tier means one token, not two.
- **Token in the path, never the query.** `/c/<token>` keeps the secret out of access logs and CSP violation reports (per `modern-web-guidance` security §2.3), and it's the only shape the layout access-gate can read (Next.js layouts don't receive `searchParams`). The `[id]` segment resolves as a *handle*: a uuid (owner/admin only — a uuid is **not** a capability) or a view token (anonymous read). base64url tokens never collide with the uuid format.
- **404, never 403.** A missing/invalid token → `notFound()`; a private collection's existence is never revealed. `referrer: no-referrer` on token views.
- **Middleware:** `/c/collection/` is public so a view link opens without a login redirect; `/c/collections` (the owner's private list) stays protected (trailing slash disambiguates).

## Why edit-via-accounts instead of an edit link (PP-wqit.7)

Real attribution (`added_by` is a named person), per-person revocation, no world-editable-link footgun, no anonymous writes. This ships view sharing complete rather than a big half-feature.

## Test Plan

- [x] Integration: token resolver (valid/anon → read, unknown/empty → null, disabled column unreachable); enable mints / disable nulls / enable idempotent; reset rotates; owner-only gating on both actions.
- [x] RTL: Share dialog toggle/link/reset state + error surface.
- [x] Smoke E2E: owner creates + adds a machine + enables share → a fresh **anonymous** context opens the link and sees the read-only Overview with **no** owner controls.
- [x] `pnpm run check` green (1513 unit); migration chain clean (0053→0054), `db:generate` reports no pending changes.
- [ ] CI: production build + full E2E.

## Related Issues

Closes #PP-wqit.2 (bead)


timothyfroehlich:
Claude review of head a5fc710 — medium review: fixed share-toggle tab-preserve (startsWith) + removed dup comment; added E2E revocation test; verified empty-collection timeline safe, middleware public-path exclusions correct, no side-effects-in-txn

timothyfroehlich:
Route collapsed to `/c/[id]` (dropped the redundant `collection/` segment) so the single-collection view mirrors `/m/[id]`. Shared link is now `/c/<token>`; owner link `/c/<uuid>`. Pure path rename — no access-model change. —Claude-Collections

PR #1691, opened 2026-07-17
fix(a11y,forms): motion-reduce, keyboard-operable controls, required indicator (PP-y798)

timothyfroehlich:
## What & why

Audit-confirmed accessibility + form-correctness fixes (CORE-A11Y-002/003/004, CORE-FORM-005). `pnpm run check` green (1580 unit tests + lint/types/format).

- **CORE-A11Y-002 — motion-reduce.** Paired every `animate-spin`/`animate-pulse` in `src/` with `motion-reduce:animate-none`. Fixing the shared `Button` spinner and `Skeleton` primitives covers most instances transitively; ~12 standalone sites swept too (`rg`-verified: zero bare occurrences remain in non-test `src/`).
- **CORE-A11Y-004 — real controls.** `multi-select`'s group-select-all header was a `<div role="button" tabIndex onKeyDown>`. Replaced with the native Radix `Checkbox` as the keyboard-operable control + an associated `<label>`. (A wrapping `<button>` would nest inside the Checkbox's own `<button>` — invalid HTML — so this is the correct fix; behavior and appearance preserved, `useId`-based id keeps `htmlFor` collision-safe.)
- **CORE-A11Y-003 — keyboard-sortable headers.** `IssueList`'s sortable `<th>` had `onClick` directly on it (mouse-only). Moved the trigger into a nested `<button type="button">` (matching the already-correct `CollectionOverviewTable`); `scope`/`aria-sort` stay on the `<th>`, layout unchanged.
- **CORE-FORM-005 — required indicator.** Added the existing inline `*` indicator to `CollectionFields`' `Name` label.
- **Progressive enhancement.** `IssueFilters` search input gets `name="q"` so a no-JS submit degrades to `?q=…`; `unified-report-form`'s sequential fields get `enterKeyHint`.

## Testing
- `pnpm run check` green.
- One test's label query loosened (`getByLabelText("Name")` → `/name/i`) to match the new required-indicator markup.

## Notes / not in scope
- `InviteUserDialog` progressive-enhancement refactor → separate PR.
- A custom ESLint rule enforcing the `motion-reduce` pairing, and a shared `RequiredLabel` helper → deferred follow-ups.
- A pre-existing `returnValue` deprecation hint in `discord-config-form.tsx` (deliberate legacy `beforeunload` code) surfaced because the file was touched; left as-is.

From the 2026-07-17 non-negotiables audit (bead PP-y798). 🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/app/(app)/admin/integrations/discord/discord-config-form.tsx
+++ b/src/app/(app)/admin/integrations/discord/discord-config-form.tsx
@@ -222,7 +222,7 @@ export function DiscordConfigForm({
           >
             {validatingToken ? (
               <>
-                <Loader2 className="mr-1.5 size-3 animate-spin" />
+                <Loader2 className="mr-1.5 size-3 animate-spin motion-reduce:animate-none" />
                 Validating...
               </>
             ) : (
@@ -275,7 +275,7 @@ export function DiscordConfigForm({
             >
               {validatingServer ? (
                 <>
-                  <Loader2 className="mr-1.5 size-3 animate-spin" />
+                  <Loader2 className="mr-1.5 size-3 animate-spin motion-reduce:animate-none" />
                   Validating...
                 </>
               ) : (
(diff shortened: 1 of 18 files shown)


timothyfroehlich:
Claude review of head c383f07 — code-reviewed via /code-review medium; applied multi-select group-slug extraction (prevents id/htmlFor drift) + signup autocomplete username fix; check green, verified.

PR #1692, opened 2026-07-17
fix(users): InviteUserDialog progressive enhancement via useActionState (PP-eqhs)

timothyfroehlich:
## What & why

**CORE-ARCH-002.** `InviteUserDialog` was the codebase's only `form.handleSubmit` call site: react-hook-form imperatively built `FormData` and called `inviteUser` inside `startTransition`, so the form did nothing without JS. Converted to the project-standard `<form action={formAction}>` + `useActionState` pattern (mirrors `signup-form.tsx`), dropping react-hook-form in favor of native validation + server-side Zod + server error display.

### `inviteUser` action: throw → `Result`

`useActionState` needs `(prevState, formData) => state`, so `inviteUser` now returns a `Result<{ userId, firstName, lastName }, Code>` discriminated union instead of throwing. **Behavior preserved** — the permission gating (`checkPermission("admin.users.invite", …)`, already matrix-based on `main`), `inviteUserSchema` validation, the dedup checks (auth.users + user_profiles + invitedUsers), and the deliberate "keep the invited row if the email send fails" behavior are all unchanged. Only the error-delivery contract changed. All callers updated (the dialog + the integration test; the other importers only render the dialog).

### Form correctness (CORE-FORM-001..006)

`type="email"`; `given-name`/`family-name`/`email` autocomplete tokens; visible required indicators (existing inline idiom); `enterKeyHint` on sequential fields. `sendInvite`/`role` are now native inputs so the form submits without JS.

## Testing
- `pnpm run check` green (types, lint, format, 1580 unit tests).
- `src/test/integration/admin/user-management.test.ts` updated to the new signature — 11/11.

## Honest scope note
The Radix `Dialog` container still needs JS to open (true of every modal in the app); what this fixes is the **form submission mechanism** — it now goes through the platform `action` path with real named inputs and server validation, instead of RHF's JS-only imperative submit. The form was extracted into a child mounted inside `DialogContent` so `useActionState` resets cleanly on each open.

From the 2026-07-17 non-negotiables audit (bead PP-eqhs). 🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/e2e/full/invite-signup.spec.ts
+++ b/e2e/full/invite-signup.spec.ts
@@ -74,12 +74,10 @@ test.describe("User Invitation & Signup Flow", () => {
     await page.getByRole("textbox", { name: "Email" }).fill(userEmail);
 
     // Ensure "Send invitation email" is checked
-    const inviteSwitch = page.getByRole("switch", {
+    const inviteCheckbox = page.getByRole("checkbox", {
       name: /Send invitation email/i,
     });
-    if ((await inviteSwitch.getAttribute("aria-checked")) === "false") {
-      await inviteSwitch.click();
-    }
+    await inviteCheckbox.check();
 
     await page
       .getByRole("button", { name: /Invite User/i, includeHidden: false })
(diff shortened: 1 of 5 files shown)


timothyfroehlich:
Claude review of head f698c90 — code-reviewed via /code-review medium; applied aria-describedby on send-invite checkbox + email maxLength=254; PE approach preserved; check green + full E2E green, verified.

PR #1704, opened 2026-07-19
fix(pbm): throttle manual PinballMap sync to 20/hour (PP-fwe1)

timothyfroehlich:
## What

`syncPinballMapNowAction` fired a real PinballMap location fetch on **every** call with no server-side interval guard — a latent **CORE-PBM-001** conduct risk. It's not reachable today (no UI wires it), but **PP-o355.7** wires it to a technician-clickable "Sync now" button, so the throttle must exist first. This bead **blocks PP-o355.7**.

## Change

- Refuse a manual sync within `PBM_MANUAL_SYNC_MIN_INTERVAL_MS` (**3 min = at most 20/hour**) of the last **successful** sync, reusing the singleton's `lastSyncedAt` (gated on `lastSyncStatus='ok'`, so a prior failure never blocks a retry).
- A `force=true` form field bypasses the throttle.
- New `THROTTLED` variant on `SyncPinballMapNowResult` so the future button can surface a friendly "try again in ~N min" message.
- Integration tests for the throttled path and the force-bypass path (extends the existing `pinballmap-sync-trigger` suite).

## Why 20/hour (not 1/hour)

The hourly cron already covers automated freshness; PBM's conduct concern is repeated **polling loops**, not a human clicking Sync-now a handful of times. 20/hour is negligible against PBM's real limits while still killing stuck-key / double-click bursts. (Product call: Tim, 2026-07-19 — recorded on PP-fwe1.)

## Testing

- `pnpm run check` green (1588 unit/integration tests).
- New: throttle-refusal + force-bypass integration tests pass.

### DIFF
--- a/src/app/(app)/m/pinballmap-actions.ts
+++ b/src/app/(app)/m/pinballmap-actions.ts
@@ -19,7 +19,11 @@ import { db } from "~/server/db";
 import { userProfiles } from "~/server/db/schema";
 import { checkPermission, getAccessLevel } from "~/lib/permissions/helpers";
 import { type Result, ok, err } from "~/lib/result";
-import { syncLocationSnapshot } from "~/lib/pinballmap/state";
+import {
+  getPinballMapState,
+  syncLocationSnapshot,
+} from "~/lib/pinballmap/state";
+import { PBM_MANUAL_SYNC_MIN_INTERVAL_MS } from "~/lib/pinballmap/config";
 import { reconcileAfterSync } from "~/lib/pinballmap/sync";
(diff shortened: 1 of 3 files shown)


timothyfroehlich:
Superseded by #1712 (PP-hbi0), merged to main. The seam-level throttle in #1712 lives in syncLocationSnapshot() and covers this action's path plus the verify path, and fixes the fail-open (last-attempt vs last-success), TOCTOU, and client-force-bypass gaps this PR's action-scoped version had. Closing in favor of it. —Claude

PR #1705, opened 2026-07-19
feat(pinballmap): wire mandatory X-Api-Token blanket auth gate (PP-uusr)

timothyfroehlich:
## What & why

PinballMap is closing its previously-open API behind a required `api_token` on **July 30 2026** (blog 2026-07-16). Once their `REQUIRE_API_TOKEN` gate flips on, **every v1 endpoint — reads included — requires the token.** Our reads (snapshot sync, link/verify) are anonymous today and would break at go-live. This PR wires the plumbing so the integration keeps working the moment an approved token is seeded. (CORE-PBM-001; unblocks the PP-o355.10 rollout.)

**Two-layer auth** (confirmed from PBM's `Api::V1::BaseController` source):
- `api_token` → blanket access gate for **reads + writes**, sent as the `X-Api-Token` header, 120 req/min global per token.
- Writes **also** still carry the per-operator identity (`user_email` + `user_token`) as query params — a distinct layer.

## Changes

- **Schema + migration `0055`:** `pinballmap_state.api_token_vault_id` + a `SECURITY DEFINER` `get_pinballmap_api_token()` RPC (EXECUTE granted to `service_role` only) that decrypts the Vault secret. Mirrors the `get_discord_config()` pattern from `0028`.
- **`api-token.ts`:** server-only accessor over the RPC (service-role admin client, `assertNotInTransaction` tripwire).
- **Live client:** `createLiveClient(apiToken)` attaches `X-Api-Token` on **every** request (reads + writes), omitted while unprovisioned. `getPinballMapClient()` is now async and loads the token before constructing the live client (2 prod callers updated).
- **Seed:** `supabase/seed-pinballmap-token.mjs` (+ `db:_seed-pinballmap-token`, added to `db:reset`) — one-time bootstrap from `PINBALLMAP_API_TOKEN` → Vault, upserting the singleton; no-op once linked. Mirrors `seed-discord.mjs`.
- **Docs:** CORE-PBM-001, `pinballmap-api-terms.md`, `ENV_VARS.md`, `.env.example`.

## Validation

- `pnpm run check` (types, lint, **1583 unit tests** incl. new header-gate seam tests) ✅
- `build` + PGlite integration + **Supabase integration** (exercises migration 0055 + RPC) ✅
- Manual round-trip vs local Supabase: RPC `NULL` when unprovisioned → seed creates Vault secret + upserts singleton → RPC returns decrypted token → re-seed no-op ✅
- Local `smoke` E2E fails at `auth.setup` (seeded logins rejected) — pre-existing environment issue in auth code this PR does not touch; the change is a dormant PBM path. Relying on CI for E2E.

## Human-gated follow-up (critical path)

The actual token doesn't exist yet: **request one at pinballmap.com/api_token** (human-approved, may take days — July 30 gate). This PR lands the plumbing token-less (column nullable, header omitted when null); seed the approved token via `PINBALLMAP_API_TOKEN` + `pnpm run db:_seed-pinballmap-token` once granted.


timothyfroehlich:
Claude review of head 4a0dcfc — Mandatory X-Api-Token gate. Migration renumbered 0056->0057 (regenerate protocol, no meta hand-edit); db:reset clean, no drift. RLS regression test ADDED and it CAUGHT A REAL VULN: get_pinballmap_api_token relied on REVOKE/GRANT only, but Supabase re-grants EXECUTE to anon/authenticated by default -> any authenticated user could read the decrypted Vault token via PostgREST RPC. Fixed with in-body auth.role()='service_role' guard + RAISE EXCEPTION (mirrors get_discord_config 0028/0029); test asserts anon+authenticated denied, service_role ok (3/3). check green 1602 tests, supabase integration 107/107. Not UI-touching.

PR #1703, opened 2026-07-19
refactor(machines): extract machine mutation services (PP-u4ab.1)

timothyfroehlich:
## What

PR 1 of the MCP remote-admin epic (**PP-u4ab**). Behavior-preserving extraction of machine mutation logic into typed service functions in `src/services/machines.ts`, mirroring the `src/services/issues.ts` convention so PR 2's MCP tools can call the same code the web actions do.

Each service takes a params object + `actorUserId`, owns the whole `db.transaction` (row writes, watcher reconciliation, timeline emits, notification *planning*), resolves `getChannels()` before the transaction, and returns `{ machine, deliveryPlan }`. **Authorization stays in the callers** (CORE-ARCH-008); notification delivery moves to `after()` + `dispatchNotification` post-commit (CORE-ARCH-011).

## Services

- **`createMachine`** — both create paths (normal + guest→member promotion), owner `subscribe` watcher, `machine_added`/`owner_set` lifecycle, active-promote `machine_ownership_changed` notification.
- **`updateMachineOwner`** — owner-only slice for PR 2's `set_machine_owner` tool: watcher reconcile (drop old / subscribe new) + `owner_changed` + old/new-owner notifications. Name and presence untouched.
- **`updateMachinePresence`** — presence update + `presence_changed` emit + unchanged-value no-op guard.

`createMachineAction` and `updateMachinePresenceAction` become thin wrappers (net −143 lines in `actions.ts`).

## Two conscious calls (spec-sanctioned)

1. **`updateMachineAction` left untouched.** Its combined name+presence+owner+pbm+description edit stays one atomic transaction; splitting it out was explicitly out of scope ("don't balloon"). `updateMachineOwner` is therefore a standalone *second* instance of owner-change logic (Rule of Three not yet tripped) — it exists for the MCP tool.
2. **Minor behavior fix:** the `forcePromoteUserId` *create* path now honors `presenceStatus`/`description` (previously silently dropped from that branch's insert). No test pinned the old drop.

## Testing

- New `src/test/integration/machine-services.test.ts` (12 tests) exercises the three services directly at the PGlite integration layer (CORE-TEST-005).
- Existing machine action tests gained the standard `after()` + notification mock scaffolding (the refactor routes notifications through `after()`).
- Green locally: typecheck, lint, format, unit (1588), **build**, **PGlite integration (462)**, **supabase-integration (104)**. Full E2E left to CI.

[...]

PR #1707, opened 2026-07-19
feat(mcp): remote-admin MCP endpoint + Supabase OAuth + six tools (PP-u4ab.2)

timothyfroehlich:
## What

PR 2 of the MCP remote-admin epic (**PP-u4ab.2**). Adds the PinPoint MCP server: a Streamable-HTTP endpoint that lets Claude drive production PinPoint through the app's existing protected code paths, authenticated by Supabase's OAuth 2.1 server and gated to admins.

Spec: `docs/superpowers/specs/2026-07-18-mcp-remote-admin.md` §PR 2. Builds on the PR 1 service layer (#1703).

## ⚠️ Pre-merge gate — do not merge until the handshake is verified

The one genuinely uncertain link (Claude Code ↔ Supabase Dynamic Client Registration) can't be tested in code or CI. Before merge, Tim must:
1. Enable **OAuth Server** on the Supabase project (Dashboard → Authentication → OAuth Server).
2. Enable **Dynamic Client Registration** (or pre-register Claude's client if DCR misbehaves).
3. Deploy, then `claude mcp add --transport http pinpoint <url>`, complete OAuth consent, and confirm `whoami` resolves.

Per Tim's call, tools were built first and the handshake is verified at the end — this PR carries the complete code; the live verification is the remaining step.

## How

- **Route** `src/app/api/mcp/[transport]/route.ts` — `createMcpHandler` + `withMcpAuth`, SSE disabled. RFC 9728 metadata at `src/app/.well-known/oauth-protected-resource/route.ts`.
- **`verifyToken`** (security core) — validate the Supabase JWT via `getClaims` (local JWKS verify) → resolve user → **require `accessLevel === "admin"`** at the door; per-tool `checkPermission` still runs underneath (defense in depth). One structured audit line per call.
- **Six-tool v1 catalog** + a `whoami` diagnostic: `list_machines`, `get_machine`, `set_machine_availability`, `add_machine`, `set_machine_owner`, `create_issue`. Each Zod-validates, `checkPermission`-gates with the same id as the equivalent web action, calls the PR1 service layer, and returns a compact result. Machines resolve by **initials or UUID**; owners by **member name or UUID**; **emails never returned** (CORE-SEC-007).
- **Refactor**: extracted `resolvePbmLinkColumns` → `~/lib/pinballmap/link-columns.ts` (third consumer: the two machine actions + `add_machine`).
- **Zero new env vars** (CORE-SEC-009) — the authorization server derives from `NEXT_PUBLIC_SUPABASE_URL`.
- `tsconfig.app.json`: include `src/app/.well-known/**` so the dot-folder route is typechecked and ESLint can resolve it (TS include globs skip dot-dirs).

## Tests

- `src/lib/mcp/verify-token.test.ts` — admin ✓ / non-admin ✗ / garbage ✗, Supabase verification mocked at the SDK boundary (CORE-TEST-006).
- `src/test/integration/mcp-tools.test.ts` — PGlite; every tool: admin path, member denial, unknown machine, owner name-resolution, presence no-op, duplicate initials.
- Gates green: `pnpm run check` (1598 unit), integration (MCP tools + machine + PinballMap), `next build`.

## Not in scope (v1)

delete/rename machine, comments, assignment, status changes, user admin — deferred by the spec. No Playwright (handshake verified manually + via MCP inspector).


timothyfroehlich:
**Review finding — token audience / OAuth-consent gate** (from a parallel review sweep)

AuthZ is fundamentally sound: `withMcpAuth(required:true)` gates the whole handler; `verifyToken` rejects missing/invalid/non-admin tokens; all six tools + `whoami` resolve auth via `requireMcpAuthContext` and every mutation runs `checkPermission()`. Single-org app so the "acts-as-Tim" cross-org-leak concern is moot; email privacy holds (owner resolution selects only first/last name; `whoami` returns UUID + accessLevel). CORE-ARCH-011 respected (services return a delivery plan; tools dispatch via `after()` post-commit). Types clean (no `any`/`!`/unsafe `as`). Zero new env vars.

**HIGH (design gap — your call):** `src/lib/mcp/verify-token.ts` `verifyClaimsWithSupabase` validates signature + `exp` + `sub` (via `getClaims`) plus an admin-role check. It never validates the token **audience/resource** and does not require the `client_id` claim (defaults to `""` and is accepted). Consequence: **any** project-signed JWT for an admin authorizes the full write-capable MCP surface — including a plain **web-session access token** (from cookies), which never went through the OAuth consent/DCR flow this PR's narrative treats as the security boundary. A stolen/leaked admin session token becomes a remote-admin credential. Blast radius is bounded (admin-only, ~20 members), but the OAuth-consent boundary is not actually enforced at the token layer.

Hardening: require a non-empty `client_id` (OAuth-server-issued tokens carry it; session tokens don't) and/or validate `aud` against the MCP resource, plus a test asserting a session-style token without `client_id` is rejected — currently untested, because it isn't rejected. At minimum this should be a conscious, documented decision.

**LOW / NIT:**
- No rate-limit / abuse throttle on a write-capable production endpoint (`route.ts`, `maxDuration 60`). Admin-gated so low risk; worth a follow-up.
- Per-tool `checkPermission()` is effectively inert as a *gate* — the door already requires `admin` and admins pass every permission, so these never deny in practice. Fine as defense-in-depth, but it's not a second independent authZ layer: if the door gate regressed, `machines.edit` etc. would still pass for any admin. `set-machine-availability` / `set-machine-owner` also pass only `machineOwnerId` (not `invitedOwnerId`) to `checkPermission` — harmless here since only admins reach it.

Test coverage is otherwise adequate (`verify-token.test.ts` covers admin✓ / non-admin✗ / garbage✗ / no-bearer; `mcp-tools` integration covers per-tool admin path + member denial + not-found + name resolution).

—Claude

timothyfroehlich:
**Addressing the review finding** (commit 54c76a15):

**HIGH (token audience / consent gate) — fixed.** `verifyToken` now requires a non-empty `client_id` claim. Supabase's OAuth server stamps `client_id` on tokens it issues; plain web-session (cookie) access tokens carry none — so this enforces the OAuth-consent boundary at the token layer. A leaked/stolen admin *session* token no longer authorizes the MCP surface. Added a test asserting a session-style token (no `client_id`) is rejected. Went with the `client_id` discriminator over `aud` validation: it's the clean, reliable signal (OAuth-issued vs session), whereas pinning `aud` to the resource identifier is more brittle across proxy/host variations.

**Role gate — confirmed admin-only** (Tim's call this session). DCR cannot bypass it: `verifyToken` resolves the token-user's role from our DB (`getUserAccessLevel`) on every request, independent of client registration.

**LOW (rate-limit) — follow-up filed.** Admin-gated + `client_id`-gated, and the spec explicitly defers rate-limiting for v1; tracked as a follow-up bead rather than blocking this PR.

**LOW (per-tool `checkPermission` inert / `invitedOwnerId` not passed) — acknowledged, no change.** It's defense-in-depth by design (spec §"Security posture"); with an admin-only door it never denies in practice, and the `invitedOwnerId` omission is harmless since only admins reach these tools and it mirrors the existing `updateMachinePresenceAction`. If the door ever opens to technician+, both become load-bearing and I'd revisit then.

—Claude

timothyfroehlich:
Claude review of head 816e553 — Reviewed the full MCP endpoint + six tools + verifyToken. Authz sound (admin gate + client_id required + per-tool checkPermission); email privacy held; CORE-ARCH-011 respected; types clean; zero new env vars. HIGH review finding (session-token accepted) fixed with client_id gate + test. Lockfile drift caught and repaired (safe brace-expansion/js-yaml). CI green: 1599 unit, integration, all E2E.

PR #1709, opened 2026-07-21
feat(a11y,forms): app-wide soft-keyboard resilience + reflow CI guard (PP-a0pl)

timothyfroehlich:
## What

App-wide rollout of the soft-keyboard occlusion fix that previously landed scoped to the Machine Settings page (PR #1388), plus a CI regression guard and docs. Closes the remaining work on **PP-a0pl**.

- **Root viewport export** (`src/app/layout.tsx`): `interactive-widget=resizes-content` now applies to *every* page — including the public anonymous issue-report form — so the on-screen keyboard reflows content above itself instead of covering it. Removed the now-redundant per-page export on the settings tab.
- **`scrollIntoView`-on-focus + `scroll-margin`** extended to `InlineEditableField` (the two machine-wide TipTap callouts) and `EditableCell`, matching the already-shipped `RowEditSheet` pattern. This is the **cross-browser floor**: iOS Safari and Firefox ignore `interactive-widget` entirely, so the focus-scroll is what actually keeps a focused field reachable there.
- **New guard** `e2e/full/soft-keyboard-reflow.spec.ts`: at keyboard-open-height geometry (390×430) it asserts focused settings fields stay inside the visible viewport. Runs on all projects incl. **Mobile Safari (WebKit)** — the closest headless iOS proxy.
- **Design-bible §19**: documents `interactive-widget` honestly as a *below-Baseline-floor* progressive enhancement (Chromium-only, safe iOS no-op, `scrollIntoView` floor, verification limit).

## Why this shape

Research pass (MWG + source): `interactive-widget` has **no Baseline/MWG guide** (it's Chromium-only, *not* Widely-available), and the `forms` guide prescribes exactly this approach. The CSS-native `scroll-initial-target` is Limited-availability, so we use the standard `scrollIntoView({block:"nearest"})` — not the non-standard `scrollIntoViewIfNeeded`.

## Verification

- ✅ `pnpm run check` green (types, lint, format, 1580 unit tests).
- ✅ **Guard passes locally on Firefox** (2/2): machine-wide editors reachable + `RowEditSheet` input rides above the fold (mobile tap-to-open at 390px works).
- ⚠️ **Honest limit:** no headless tool has a real virtual keyboard, so the guard is a **reachability/reflow proxy**, not pixel-exact keyboard geometry (that's PP-a0pl Part 3 — a paid real-device cloud, deferred). Documented in the spec header + §19.
- ⚠️ **No iPhone available for manual verification** — the fix is a strict Chromium/Android improvement and a safe iOS no-op backed by the `scrollIntoView` floor; CI's Mobile Safari (WebKit) run is the closest automated iOS check.

> Local Chromium/WebKit E2E are currently broken on the dev host (Chromium sandbox input-injection crash; WebKit missing `libicudata.so.74`) — pre-existing environment issues unrelated to this PR; a separate DevX bead tracks them. CI runs all engines.

### DIFF
--- a/.agents/skills/pinpoint-design-bible/SKILL.md
+++ b/.agents/skills/pinpoint-design-bible/SKILL.md
@@ -626,6 +626,20 @@ These are not in PinPoint today. They require a per-feature opt-in here in §19 
 - **`interestfor` attribute** for tooltips — Chrome-only as of late 2025; defer.
 - **`closedby` attribute** on `<dialog>` — Limited availability (no Safari).
 
+### Adopted below the Baseline floor (safe-no-op progressive enhancements)
+
+A feature below the Widely-available floor may still be used **when the browsers
+that lack it degrade to a harmless no-op** (never a broken experience) and a
+Widely-available primitive already covers the same need as the floor. Each such
+feature is listed here with its status, why it degrades safely, and the
+cross-browser floor that carries the non-supporting browsers.
+
+| Feature                                         | Status                  | Degrades to                            | Cross-browser floor                                                 |
+| :---------------------------------------------- | :---------------------- | :------------------------------------- | :------------------------------------------------------------------ |
+| `interactive-widget=resizes-content` (viewport) | Limited (Chromium-only) | The browser default (`resizes-visual`) | `scrollIntoView`-on-focus + `scroll-margin` in the settings editors |
+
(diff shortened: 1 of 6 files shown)


timothyfroehlich:
Claude review of head 73e30e1 — App-wide soft-keyboard resilience; shipped app code is clean (correct interactive-widget viewport, scrollIntoView, CORE-A11Y/RESP/TS all pass). Known follow-up (filed as P2 bead): the reflow E2E guard is vacuous — Playwright auto-scrolls before the assertion — test-quality only, not a code defect. Synced main for #1711 audit fix.

PR #1683, opened 2026-07-17
feat(machines): PinballMap listing control — read side (PP-o355.12)

timothyfroehlich:
## PinballMap outbound listing — READ side (PP-o355.12, part 1 of 2)

Replaces the machine edit/create form's disabled "List on PinballMap" checkbox with a **state-aware "PinballMap listing" control**. This is the **read-only half** — link/verify/reconnect against PinballMap's *public* lineup (anonymous, no token). The **write half** (operator token + `addMachine`/`removeMachine`, i.e. actually listing/unlisting a brand-new machine) is a **separate follow-up PR** and is deliberately out of scope here.

Builds on the merged foundation (PP-o355.16): `pinballmap_state` singleton + `machines.pinballmap_lmx_id` store-and-heal columns. **No migration** in this PR.

### The model — a machine linked to a PBM catalog title is in one of four listing states
1. **Not listed / on-PBM-but-unlinked** (no lmx) → **Connect to PinballMap**: one read-only lineup lookup that captures the existing entry (→ listed) or reports the title isn't on our lineup yet (ABSENT).
2. **Listed + linked** (lmx set) → **Listed** badge + **Verify link**.
3. **Link broken** (a Verify came back stale) → **⚠ Link may be broken** + **Reconnect**.

lmx = `location_machine_xref` id, the durable listing handle. Resolved from the **stored** snapshot — never a per-render PBM call (CORE-PBM-001, one sync/hour).

### What's here (plan Tasks 1–6)
- **`machines.pinballmap.push`** permission (admin-only) added to the matrix — the gate for the *write* follow-up. Read verbs stay on the existing `machines.pinballmap.link` (owner/tech/admin).
- **`findLmxForMachine`** — resolves a machine's lmx from the stored `pinballmap_state` snapshot.
- **`pinballmap_listing`** timeline event kind (listed / unlisted / linked / reconnected) — icon + formatter.
- **`linkPinballmapEntryAction`** — read-only lineup fetch, captures lmx, flips `listed`, mirrors a `linked` timeline event. Sync-if-absent only; no per-render PBM call. Side effects kept outside the DB transaction (CORE-ARCH-011).
- **`verifyPinballmapLinkAction`** — confirms / heals / flags the stored lmx against a fresh lineup (ok / healed / stale).
- **`PinballmapListingControl`** — the branching UI, rendered as its **own `<form>` region outside** the machine-edit form (nested forms are invalid HTML). Wired into both the edit dialog and the create form.

### Testing
[...]

PR #1711, opened 2026-07-21
chore(deps): patch audit highs (brace-expansion, js-yaml) + bump pnpm pin

timothyfroehlich:
## What

Dependency-hygiene pass. Two independent, low-risk changes kept as **separate commits** so either can be dropped in review.

### Commit 1 — audit overrides (primary, urgent)

Two freshly-published transitive **high** advisories were tripping the chores `pnpm audit --audit-level=high` gate (and blocking #1709's CI). Both are **dev-only transitive** deps reached via `@eslint/eslintrc` — runtime/production is unaffected, and neither is related to any feature work.

`pnpm-workspace.yaml` already carried overrides for both packages, but pinned one patch too low. Bumped them to the patched ranges:

| Package | Old override | New override | Advisory |
| --- | --- | --- | --- |
| `brace-expansion` | `>=5.0.6` | `>=5.0.7` | [GHSA-3jxr-9vmj-r5cp](https://github.com/advisories/GHSA-3jxr-9vmj-r5cp) — ReDoS |
| `js-yaml` | `>=4.2.0` | `>=4.3.0` | [GHSA-52cp-r559-cp3m](https://github.com/advisories/GHSA-52cp-r559-cp3m) — quadratic-CPU merge keys |

> Note: pnpm 11 no longer reads `pnpm.overrides` from `package.json`; overrides live in `pnpm-workspace.yaml` (migrated in PP-w0eq). The bumps were applied there, following the file's existing `>=` convention.

After `pnpm install`, `pnpm audit --audit-level=high` reports **0 high** (was 2).

### Commit 2 — pnpm corepack pin bump (routine)

Bumps the `packageManager` corepack pin `pnpm@11.7.0 -> 11.8.0` (minor, 30-day cooldown clear) via `corepack use pnpm@11.8.0`. `pnpm install --frozen-lockfile` leaves **only `package.json`** changed — no lockfile churn.

## Testing

- `pnpm audit --audit-level=high` → 0 high
- `pnpm run check` → green (types, lint, format, unit, linters, pytest, audits)

## Notes

- Unblocks #1709's CI / clears the chores audit red.
- Chores bead: PP-qehv.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JXzfrMKMgqJBWzzbsD8fEx

PR #1712, opened 2026-07-21
fix(pinballmap): seam-level manual-sync throttle at syncLocationSnapshot (PP-hbi0)

timothyfroehlich:
## What & why

Moves the PinballMap manual-sync throttle **down to the `syncLocationSnapshot()` seam** so every live-fetch caller inherits **one chokepoint** (CORE-PBM-001). Previously the throttle was being bolted onto individual server actions, leaving every other caller of `syncLocationSnapshot → getPinballMapClient().fetchLocation` unthrottled — notably #1683's verify/reconnect path, which forces a full `fetchLocation` on every click.

### The seam refactor

`syncLocationSnapshot({ updatedBy?, trigger })` now takes a caller-supplied `trigger: "manual" | "cron"`:

- **`manual`** (the default, so any future caller inherits the guard): rate-limited to at most one call per `PBM_MANUAL_SYNC_MIN_INTERVAL_MS` = 3 min → ≤20/hr (approved by Tim, 2026-07-19). Returns `{ ok: false, reason: "throttled", retryAfterMs }` when refused.
- **`cron`**: the sanctioned hourly automated refresh — never throttled, but still records its attempt so a manual refresh moments after a cron respects the fresh snapshot.

Enforcement lives at the seam; `syncPinballMapNowAction` keeps a user-facing `THROTTLED` result code.

## Supersedes #1704 (PP-fwe1) — recommend closing it

This fixes **every flaw** the 2026-07-20 review found in #1704's action-scoped throttle:

| #1704 finding | Fix here |
| --- | --- |
| **1 (HIGH)** — guard only engaged after `lastSyncStatus === "ok"`, so on 429/500 it self-disabled and every click re-hit PBM (inverting CORE-PBM-001 backoff) | Throttle is against the **last ATTEMPT**, not last success. New `pinballmap_state.last_sync_attempt_at` column, stamped at the **START** of an attempt, so a failed fetch still refuses repeat clicks. |
| **2 (MED/HIGH)** — `force=true` was a plain client-settable form field with no gate | **Dropped** — not carried forward. No client-settable bypass. |
| **3 (MED)** — TOCTOU: `lastSyncedAt` only advanced on completion, so concurrent double-clicks all passed | The stamp is a **single conditional upsert** (`INSERT … ON CONFLICT DO UPDATE … WHERE`); a row-level lock on the singleton serializes concurrent clicks and only one wins the claim. |
| **Doc drift (MED)** — NON_NEGOTIABLES.md + AGENTS.md still said "one sync call/hour" | Both updated to the approved model: cron = 1/hr automated, manual ≤20/hr human. |

**#1683** (verify/reconnect, PP-o355.12) will inherit this throttle automatically once it rebases on main — its `fetchLocation` goes through the same seam and picks up the default `manual` trigger.

## Migration

Adds one nullable column: `drizzle/0056_uneven_inertia.sql` — `ALTER TABLE "pinballmap_state" ADD COLUMN "last_sync_attempt_at" timestamp with time zone;` (generated via `db:generate`, applied locally via `db:migrate`).

## Tests

Integration (PGlite, PBM pinned to the in-memory mock at the client seam — CORE-TEST-006, never reaches pinballmap.com):

- A 2nd manual sync within the interval is **refused** AND `fetchLocation` call count **stays at 1** (the assertion #1704 lacked).
- The guard **still refuses after a FAILED attempt** (the fail-open regression).
- The **cron/automated path is never blocked**, even right after a manual sync.

Existing foundation tests updated to drive the unthrottled `cron` trigger where they exercise raw persistence with back-to-back syncs.

## Verification

`pnpm run check` green (typecheck app+tests, lint, 1588 unit tests, linters). Full `preflight` green through build + PGlite integration (465) + local-Supabase integration (104). The smoke E2E stage failed locally only at `auth.setup` (login→dashboard) due to a worktree Supabase port/provisioning mismatch — unrelated to this change, which touches no auth/login/middleware path; CI runs the full E2E suite in a clean environment.

Refs PP-hbi0, PP-fwe1.

https://claude.ai/code/session_01JXzfrMKMgqJBWzzbsD8fEx


timothyfroehlich:
Claude review of head 3e769fd — Re-mark after clean main-sync (merged #1711 audit fix; ort auto-resolved lockfile only, no logic change from reviewed b984462). Seam-level PBM throttle PP-hbi0: atomic conditional-upsert guard on last_sync_attempt_at, fixes #1704 fail-open+TOCTOU, no client force bypass, cron unblocked, additive migration 0056, docs reconciled, integration tests cover all 3 cases. Supersedes #1704. NOTE: 0056 collides with unmerged #1705 — second to merge regenerates to 0057.

PR #1720, opened 2026-07-22
fix(pinballmap): enable RLS on pinballmap_state (PP-5gv0)

timothyfroehlich:
## What

Enables Row Level Security on `public.pinballmap_state`.

## Why

The prod Supabase security advisor (2026-07-20 chores sweep) raised an **ERROR-level** lint `0013 rls_disabled_in_public`: `pinballmap_state` is a public table reachable via PostgREST/anon with RLS **disabled**. Its sibling `pinballmap_catalog` had RLS enabled back in migration `0046`; `pinballmap_state` was simply missed in the PBM integration migration (`0052`).

This is distinct from the 14 known-intentional `rls_enabled_no_policy` INFO tables — those have RLS **on** with zero policies. `pinballmap_state` had RLS **off** entirely.

## How

- `.enableRLS()` added to the `pinballmapState` schema definition.
- Generated migration `0056_mighty_harrier.sql`: a single `ALTER TABLE "pinballmap_state" ENABLE ROW LEVEL SECURITY;`.

No policies are added — this is the deliberate **Drizzle-superuser-only** pattern used by the other PBM/app tables. The app connects as the Drizzle superuser (bypasses RLS), while anon/`authenticated` PostgREST access is denied.

## Verification

- `pnpm run db:migrate` applied locally; `pg_class.relrowsecurity` is now `t` for both `pinballmap_state` and `pinballmap_catalog`.
- `pnpm run check` green (types, lint, format, 1588 unit tests).
- `pinballmap-state` integration tests pass (4/4).
- Advisor re-check against prod to be confirmed after deploy.

Fixes PP-5gv0.


timothyfroehlich:
/audit-override transitive-dep advisories unrelated to this PR — fast-uri (GHSA-v2hh-gcrm-f6hx, via @mdx-js/loader>webpack) and sharp/libvips (GHSA-f88m-g3jw-g9cj, via next). Both pre-exist on main; this PR only adds a one-line RLS-enable migration. Dep bumps belong in a chores PR. —Claude

PR #1721, opened 2026-07-22
fix(worktree): branch new worktrees off fresh origin/main, not stale root HEAD (PP-2cpf)

timothyfroehlich:
## Problem

The `WorktreeCreate` hook (`.claude/hooks/worktree-create.sh`) created every worktree with `git worktree add -b <branch>` off `$BASE_PATH` — the **root checkout** — at its current `HEAD`, with no fetch. The root checkout stays on `main` but is never fast-forwarded, so both **bridge sessions** (`CLAUDE_CODE_ENVIRONMENT_KIND=bridge`) and **`Agent(isolation:"worktree")` dispatch** started every new worktree N commits behind `origin/main` (observed **12 behind** this session).

A stale base silently poisons anything that reads local files — most visibly the session **briefing's `pnpm audit`** flagging CVEs that were already patched upstream (a known past false-"regression").

## Fix

Freshen at the chokepoint we already own. Before `git worktree add`:
- Best-effort `git fetch origin main` and branch off the fetched SHA (`FETCH_HEAD`).
- Fall back to `HEAD` when the fetch fails (offline) so worktree creation never hard-depends on the network.
- The fetch runs **outside the flock**, so network latency doesn't extend the cross-session lock hold.
- Bounded by `GIT_HTTP_LOW_SPEED_LIMIT/TIME` (<1 KB/s for 15s aborts) so a stalled network degrades to the fallback instead of hanging every worktree creation.

Fixes both bridge and agent-dispatched worktrees in one place; no bridge-config access required.

## Tests

`scripts/tests/test_worktree_create_hook.py`:
- Upgraded the mock `git` to branch on subcommand — `fetch` honors `$MOCK_GIT_FETCH_EXIT`, `rev-parse` returns a stub `FETCH_HEAD` SHA.
- New: branches off the fetched `origin/main` SHA on success.
- New: falls back to `HEAD` when the fetch fails (offline).
- Existing shape/validation tests updated for the new call sequence.

`pnpm run check` green (typecheck, lint, format, 110 pytest incl. 6 hook tests, shellcheck clean).

Bead: PP-2cpf

### DIFF
--- a/.claude/hooks/worktree-create.sh
+++ b/.claude/hooks/worktree-create.sh
@@ -151,7 +151,7 @@ do_worktree_add() {
   local last_stderr=""
 
   for attempt in $(seq 1 "$max_retries"); do
-    last_stderr=$(git -C "$BASE_PATH" worktree add "$WORKTREE_PATH" -b "$BRANCH" 2>&1) && {
+    last_stderr=$(git -C "$BASE_PATH" worktree add "$WORKTREE_PATH" -b "$BRANCH" "$BASE_REF" 2>&1) && {
       echo "$WORKTREE_PATH"
       return 0
     }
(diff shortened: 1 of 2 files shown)


timothyfroehlich:
/audit-override fast-uri / sharp / @hono/node-server are pre-existing transitive advisories inherited from main (via @mdx-js/loader›webpack›ajv, next›sharp, and the MCP SDK respectively). This PR changes only .claude/hooks/worktree-create.sh + scripts/tests/test_worktree_create_hook.py — zero dependency changes, so the audit red is unrelated to it. The real fix is already in flight in the deps-audit-highs lane. —Claude

PR #1724, opened 2026-07-23
fix(deps): dedupe lockfile + bump next to 16.2.11 — unbreaks main CI (PP-f2yn)

timothyfroehlich:
## Why this is urgent

**Main's CI is currently red.** PR #1718 (tiptap group bump, `c0429cc5`) committed a `pnpm-lock.yaml` with duplicate `@types/node@26.1.1` keys (4 occurrences instead of 2). CI's `Setup Dependencies` step hits `ERR_PNPM_BROKEN_LOCKFILE` and fails, which skips `Build` — so **no new production deploy can ship** until this lands. (No outage; Vercel keeps serving the last good deploy.)

#1721 passed pre-merge only because it branched before #1718; its merge commit combined with main's broken lockfile, turning main red.

## What this PR does

1. **Dedupe the lockfile** — removes the two duplicate `@types/node@26.1.1` blocks (surgical, versions unchanged). This is what actually fixes the install failure / red main.
2. **Bump `next` 16.2.10 → 16.2.11** — clears 4 high-severity advisories (GHSA-6gpp-xcg3-4w24, GHSA-89xv-2m56-2m9x, GHSA-m99w-x7hq-7vfj, GHSA-p9j2-gv94-2wf4), which the now-unblocked `pnpm audit` surfaced. Folded in so main goes fully green in one merge (the dedup alone would leave the audit gate red on these).

## Soak note

The `next` bump is a deliberate expedite past Dependabot soak — high-severity security patch, same-minor bump of first-party Next.js. `pnpm update next` also re-resolved 3 incidental low-risk transitive patches: `prettier` 3.9.5→3.9.6, `electron-to-chromium`, `baseline-browser-mapping` (browserslist data).

## Verification
- `pnpm install --frozen-lockfile` passes → lockfile consistent with `package.json`
- `pnpm audit --audit-level=high` → 0 high (1 pre-existing moderate, `@hono/node-server`, below the gate)
- `pnpm run check` green (1629 tests)

Closes PP-f2yn.

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,7 +130,7 @@
     "drizzle-orm": "^0.45.2",
     "lucide-react": "^1.0.1",
     "mcp-handler": "1.1.0",
-    "next": "^16.2.6",
+    "next": "^16.2.11",
     "nodemailer": "^9.0.1",
     "pino": "^10.3.1",
     "postgres": "^3.4.9",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -101,7 +101,7 @@ importers:
         version: 1.0.12(react-dom@19.2.7(react@19.2.7))(react@19.2.7)
       '@sentry/nextjs':
         specifier: ^10.48.0
-        version: 10.64.0(@opentelemetry/core@2.9.0(@opentelemetry/api@1.9.1))(@opentelemetry/sdk-trace-base@2.9.0(@opentelemetry/api@1.9.1))(next@16.2.10(@babel/core@7.29.7)(@opentelemetry/api@1.9.1)(@playwright/test@1.61.1)(@types/node@26.1.1)(react-dom@19.2.7(react@19.2.7))(react@19.2.7))(react@19.2.7)(webpack@5.108.4(esbuild@0.28.1)(postcss@8.5.16))
+        version: 10.64.0(@opentelemetry/core@2.9.0(@opentelemetry/api@1.9.1))(@opentelemetry/sdk-trace-base@2.9.0(@opentelemetry/api@1.9.1))(next@16.2.11(@babel/core@7.29.7)(@opentelemetry/api@1.9.1)(@playwright/test@1.61.1)(@types/node@26.1.1)(react-dom@19.2.7(react@19.2.7))(react@19.2.7))(react@19.2.7)(webpack@5.108.4(esbuild@0.28.1)(postcss@8.5.16))
       '@supabase/ssr':
         specifier: ^0.12.0
         version: 0.12.0(@supabase/supabase-js@2.110.1)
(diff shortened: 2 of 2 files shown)


timothyfroehlich (commented):
## Review — Claude-McpBearerToken

Reviewed head `e1161d51`. I hit this same broken lockfile independently 30 minutes ago and had opened #1728 as a duplicate — **I've closed mine in favour of this one.** What follows is a real review, not a rubber stamp.

### The dedupe is correct ✅

I verified the whole lockfile, not just the changed hunks — parsed all 2472 top-level keys and checked for repeats **within each section**:

```
duplicates: NONE
total top-level keys: 2472
```

So `'@types/node@26.1.1'` is now singular in both `packages:` and `snapshots:`, and there is no second latent duplicate hiding elsewhere. `CI Gate SUCCESS` independently confirms `pnpm install --frozen-lockfile` parses it. This does unbreak main.

### Finding 1 — the Next bump is partial, and it splits `@next/env` in two 🟡

`package.json` bumps only `next`:

```
"@next/env": "^16.2.3",   →  resolves 16.2.10
"@next/mdx": "^16.2.3",   →  resolves 16.2.10
"next":      "^16.2.11",  →  resolves 16.2.11
```

Result in the merged lockfile — **two copies of `@next/env` coexist**:

```
1078:  '@next/env@16.2.10':     ← direct dep, pulled by @next/mdx@16.2.10
1081:  '@next/env@16.2.11':     ← pulled by next@16.2.11
```

This matters more than "it's just a patch," because `@next/env` is not vendored decoration here — it's the `.env*` loader that three separate entry points import directly:

- `drizzle.config.ts` (`loadEnvConfig`) — decides which `POSTGRES_URL` migrations run against
- `playwright.config.ts` — where the `dev: true` argument is load-bearing, per the comment already in that file
- `scripts/workflow/pr-screenshots.mjs`

So the Next *runtime* now loads env through 16.2.11's copy while the migration/E2E/screenshot tooling loads it through 16.2.10's. Env-resolution semantics drifting between two copies is a quiet failure mode, and this repo is strict about env handling (CORE-SEC-009). Separately, `@next/mdx@16.2.10` against `next@16.2.11` is a skew Next doesn't test — the family ships in lockstep.

**Suggested fix (two lines + a lockfile regen):** bump `@next/env` and `@next/mdx` to `^16.2.11` as well, then `pnpm install --lockfile-only`. That collapses `@next/env` back to one copy and keeps the family aligned.

### Finding 2 — scope coupling 🟡

The PR title is "unbreaks main CI," but the CI unblock and a framework upgrade are riding together. The dedupe alone is sufficient — I verified locally that the 7-line deletion by itself makes `pnpm install --frozen-lockfile` succeed and `pnpm run check` pass, with zero version movement.

Coupling them means a revert of the Next bump also reverts the unblock, and it makes this diff 62/64 lines instead of 0/7. Worth naming; the `^16.2.6 → ^16.2.11` floor raise is also not needed for the fix, since `^16.2.6` already admits 16.2.11.

### Verdict

No blocking defects — the dedupe is verified correct and CI is green. Finding 1 is the one I'd actually fix before merging: it's two specifier lines, and per this repo's "polish before shipping, no fast-follow" rule it's cheaper now than as a follow-up PR. If you'd rather take main green immediately and align the Next family in a separate change, that's a defensible call — just make it a deliberate one rather than an accident of the bump.

— Claude

timothyfroehlich:
Claude review of head e1161d5 — dedupe verified correct (all 2472 lockfile keys scanned, no duplicates remain); Next 16.2.11 bump is load-bearing — it patches 4 high SSRF advisories (GHSA-89xv-2m56-2m9x, GHSA-p9j2-gv94-2wf4) that fail pnpm audit below that floor; 1 open non-blocking finding — @next/env and @next/mdx left at 16.2.10, splitting @next/env into two copies

timothyfroehlich:
**Correction to Finding 2 of my review above — I had it wrong, and it changes the recommendation.**

I called the Next bump "scope coupling" riding along with the CI unblock. It isn't. Reproduced locally on a tree without the bump:

```
high  Next.js: Server-Side Request Forgery in Server Actions on custom servers
      Vulnerable: >=16.0.0 <16.2.11   Patched: >=16.2.11   (GHSA-89xv-2m56-2m9x)
high  Next.js: Server-Side Request Forgery in rewrites via attacker-controlled destination hostname
      Vulnerable: >=16.0.0 <16.2.11   Patched: >=16.2.11   (GHSA-p9j2-gv94-2wf4)
...
10 vulnerabilities found — 6 moderate | 4 high
```

`pnpm audit --audit-level=high` fails on anything below 16.2.11, reached three ways (`.>next`, `.>@sentry/nextjs>next`, `.>mcp-handler>next`). So the bump is **load-bearing** — it clears the audit gate, and 16.2.11 is precisely the patched floor. Bundling it with the dedupe is the right call, not scope creep. Disregard Finding 2.

**Finding 1 still stands and is now the stronger argument.** `@next/env` and `@next/mdx` are still `^16.2.3` → 16.2.10, so the tree carries two `@next/env` copies and the MDX plugin trails the framework. Since you're already touching the Next family to patch a security advisory, taking all three to `^16.2.11` in the same PR is the tidy version of the change you're already making.

— Claude

PR #1727, opened 2026-07-23
feat(mcp): static bearer-token auth, replacing Supabase OAuth (PP-u4ab.7)

timothyfroehlich:
## What

Pivots the MCP remote-admin server's auth from Supabase's OAuth 2.1 server to a **static bearer token**, with Claude Code as the client.

Plan: [`docs/plans/2026-07-22-mcp-bearer-token-pivot-handoff.md`](https://github.com/timothyfroehlich/PinPoint/blob/feat/mcp-bearer-token/docs/plans/2026-07-22-mcp-bearer-token-pivot-handoff.md) (first commit on this branch). Bead: PP-u4ab.7, epic PP-u4ab.

## Why

The OAuth flow worked all the way to the token exchange — discovery, DCR, authorize, our consent page, auth code issued — and then Supabase's own `/oauth/token` endpoint returned a 500. Root cause is [supabase/auth#2339](https://github.com/supabase/auth/issues/2339): Claude requests the `openid` scope, Supabase must sign an OIDC ID token, picks HS256 (which OIDC forbids for ID tokens), and the signing throw is wrapped as a 500. Behind that sit more open bugs (confidential-vs-public client registration), the OAuth server is beta-not-GA, and there is no public end-to-end precedent for "own MCP server + Supabase OAuth + Claude." Tim's call: stop playing whack-a-mole and use a bearer token for this private, single-user admin server.

## What changed

`verify-token.ts` keeps its shape — `createVerifyToken(deps)`, `McpAuthContext`, `requireMcpAuthContext` — so the six tools, `shared.ts`, and the audit trail are untouched. Only the verification logic is replaced:

- **Constant-time compare** of the presented bearer against `MCP_BEARER_TOKEN`. Both are hashed to fixed-width SHA-256 digests first, so `timingSafeEqual` can't throw on — or leak — the secret's length.
- **Acts as `MCP_ADMIN_USER_ID`**, whose access level is re-resolved on *every* request and must still be `admin`. A demotion revokes the token immediately. The DB lookup only runs *after* the token matches, so a wrong guess never touches the database.
- **Fails closed** on any misconfiguration: either var unset, a token under 32 chars, or a non-UUID admin id (which would otherwise reach a `uuid` column and 500 instead of 401). Each rejection logs a distinct `reason` so a 401 is diagnosable.

Two new env vars in the `next.config.ts` build registry (CORE-SEC-009), **Production-only**: previews have no MCP client pointed at them and the auth path fails closed, so a preview build shouldn't be gated on them. Catalogued in `docs/ENV_VARS.md` + `.env.example`.

Dead OAuth plumbing removed: the `.well-known/oauth-protected-resource` route, its `tsconfig.app.json` include glob and middleware public-route allowance (both added solely for it), plus `MCP_RESOURCE_METADATA_PATH` and `getAuthServerUrls`. **The `/oauth/consent` page stays** as dormant code, per Tim.

Note that `withMcpAuth` still advertises mcp-handler's *default* `resource_metadata` URL in its 401 header, and that URL now 404s. That's intentional and harmless — a bearer client is handed its credential out of band and never walks the discovery chain. The code comment says so accurately as of `b6d8accc`.

## Security posture

A static bearer token is a long-lived admin secret — the right tradeoff for a private single-user server, essentially a personal access token. It is server-only (never `NEXT_PUBLIC_`), never reused as another var's fallback, HTTPS-only, hashed before comparison, fails closed when unset, and re-checks the mapped user's role on every call. If leaked, rotate the env var. Per-tool `checkPermission()` still runs underneath each tool as defense in depth.

Note the tradeoff this makes vs OAuth: the token is **static and non-expiring**, rotating only when regenerated by hand. That makes the deferred rate-limit follow-up (PP-u4ab.3) *more* relevant, not less — though still low severity, since `openssl rand -hex 32` gives 256 bits and brute force is infeasible.

## Testing

- `verify-token.test.ts` rewritten for the bearer path (13 tests): right token → context; wrong token; prefix/suffix of the secret; no token; unconfigured; too-short secret; non-UUID admin id; demoted user. Hermetic via deps injection + `vi.stubEnv`.
- `mcp-tools.test.ts` (15 integration tests) unchanged and passing — the tools didn't change.
- CI green: typecheck, ESLint, Prettier, Build, Unit, Integration, Supabase Integration, Verify Migrations, pnpm audit, **E2E smoke + full**.
- Reviewed via the 5-agent `/code-review` fallback (Copilot out of quota). One real finding, fixed in `b6d8accc`; see the review comment below.

### Correction on local E2E

An earlier revision of this description blamed the local smoke failure on the broken `pnpm-lock.yaml` on main. **That was wrong** — I disproved it after that lockfile was fixed and merged.

The actual cause is the local host: headless Chromium on bazzite accepts no keyboard input. `fill()` and `keyboard.type()` both no-op — `keydown` fires and is not `defaultPrevented`, but `beforeinput`/`input` never fire and `el.value` stays `""`. Firefox drives the identical login fine (`POST 200` → `/dashboard`). Since `auth-setup` is pinned to `devices["Desktop Chrome"]` and every project including `firefox` depends on it, the whole local suite dies at setup regardless of branch. Filed as PP-8b6j. CI's E2E is unaffected and green here.

https://claude.ai/code/session_01HFcyx2WpvW76bvQ5AF2w8w


timothyfroehlich:
Cherry-picked the lockfile fix from #1728 (`f4ab42ff` → `ce50d594`) onto this branch so CI can run here without waiting on that merge. CI on this PR was dying at `pnpm install --frozen-lockfile` in *Setup Dependencies* — a broken `pnpm-lock.yaml` on main, not anything in this change.

The two PRs carry the identical 7-line deletion, so once #1728 lands, merging `main` into this branch resolves cleanly (same deletion from a common ancestor). #1728 is the one to merge first if you want the fix on main independently of this feature.

timothyfroehlich:
Claude review of head b6d8acc — 5-agent /code-review on the bearer-token pivot; 1 real finding, fixed in b6d8accc (route.ts wrongly claimed no resourceMetadataPath is advertised — mcp-handler stamps its default into every 401; verified against a running server) plus 2 stale OAuth-era comments; no functional defects, no CLAUDE.md/NON_NEGOTIABLES violations

timothyfroehlich:
### Code review

Five-agent review (CLAUDE.md/NON_NEGOTIABLES compliance, shallow bug scan, git history, prior-PR feedback, code-comment accuracy) against head `7cec08c8`. Copilot is out of quota until ~Aug 1, so this is the sanctioned Claude fallback.

**One real finding, fixed in b6d8accc:**

`route.ts` claimed `withMcpAuth` advertises no `resourceMetadataPath`. That is false — mcp-handler falls back to its own default and stamps `resource_metadata="<origin>/.well-known/oauth-protected-resource"` into every 401:

https://github.com/timothyfroehlich/PinPoint/blob/b6d8accc0e50b7b4e1d2b5e4dd0a91da3e8d7f5c/src/app/api/mcp/%5Btransport%5D/route.ts#L60-L71

Verified against a running server rather than by reading alone:

```
WWW-Authenticate: Bearer error="invalid_token", error_description="No authorization provided",
                  resource_metadata="http://localhost:3200/.well-known/oauth-protected-resource"
```

That path now returns **404** (the RFC 9728 route is deleted in this PR) — which is the honest answer for a server with no authorization server to discover, and harmless for a bearer client that never walks the discovery chain. Behaviour is fine; the comment asserted the opposite of reality, so it was rewritten. Two stale OAuth-era comments (`route.ts` top JSDoc, `audit.ts` `clientId` doc) were corrected in the same commit.

**One agent claim I checked and rejected:** that the advertised URL 302-redirects to `/login` now that `.well-known` was dropped from the middleware allowlist. It does not — unmatched routes 404 before the redirect applies, confirmed with and without the dev-autologin bypass.

**No functional defects found.** Specifically verified: `secretsMatch` hash-then-`timingSafeEqual` is correct and cannot throw on length mismatch; every failure path fails closed; the DB role lookup only runs *after* the token matches (no oracle on bad guesses); `MCP_ADMIN_USER_ID` is UUID-validated before reaching a `uuid` column; `AuthInfo`/`McpAuthContext` shape is byte-for-byte preserved so the six tools need no changes; no dangling references to the deleted route or removed exports; no CORE-* violations.

**History check:** the `client_id`-claim requirement added by #1707 (to stop a stolen admin *session* token driving this surface) is structurally resolved rather than dropped — Supabase JWTs are no longer accepted at all. The `.well-known` middleware allowance and tsconfig glob existed solely for the deleted route, with no other consumer in repo history.

**One correction to my own earlier note, not a code issue:** I had written that the rate-limit follow-up (PP-u4ab.3) was lower priority after this pivot. That's backwards — the OAuth design used short-lived Supabase-issued tokens; this uses a static, non-expiring secret that only rotates manually. Still LOW severity given 256 bits of entropy, but the bead shouldn't be closed as "OAuth-era debt." Bead updated.

timothyfroehlich:
### End-to-end verification against a live server

Rather than leave the acceptance test until after deploy, I ran it locally — `MCP_BEARER_TOKEN` + `MCP_ADMIN_USER_ID` set in the environment, real dev server, real DB, real MCP protocol handshake over `/api/mcp/mcp`.

**Happy path (admin-mapped token):**

```
initialize:    200  { name: 'pinpoint', version: '1.0.0' }
tools/list:    200  7 tools: whoami, list_machines, get_machine,
                    set_machine_availability, add_machine, set_machine_owner, create_issue
whoami:        200  {"userId":"051ab3ae-…","accessLevel":"admin"}
list_machines: 200  {"count":10,"machines":[{"initials":"AFM",…}]}
```

That is the epic's acceptance criterion (`whoami` → admin identity, then `list_machines`) passing before merge rather than after.

**Every rejection path, with the distinct log reason each was designed to emit:**

| Case | Status | Logged |
|---|---|---|
| No token | 401 | — |
| Wrong token (64 chars, correct shape) | 401 | `reason: "bad_token"` |
| Correct token, but `MCP_ADMIN_USER_ID` maps to a **member** | 401 | `reason: "not_admin"`, `accessLevel: "member"` |

The third row is the one that matters — it's the defense-in-depth claim in the PR description, demonstrated rather than asserted: the right secret is *not* sufficient if the mapped user isn't an admin.

**Audit trail** records the synthetic client id as designed:

```json
{"scope":"mcp.tool","tool":"whoami","userId":"051ab3ae-…","clientId":"claude-code-bearer","outcome":"ok"}
```

**Also settles an open PP-u4ab.4 question.** That bead flagged that `after(() => dispatchNotification(…))` might not be valid in the mcp-handler tool context — "mutation committed but tool returns Internal error." I exercised `set_machine_owner`, which is one of the three tools that actually uses `after()` (with `add_machine` and `create_issue`):

```
set_machine_owner(AFM -> admin):   200  isError=false  owner: "Admin User"
set_machine_owner(AFM -> member):  200  isError=false  owner: "Member User"   [restored]
```

Clean result, no thrown dispatch, audit `outcome: ok`. The `after()` concern does not reproduce. Noting precisely: my first attempt used `set_machine_availability`, which does **not** use `after()` — that run proved nothing about this, so I re-ran against a tool that does.

Incidentally the tool correctly refuses an email address as an owner identifier ("No member named ..., pass the user's UUID"), which is consistent with CORE-SEC-007.

All local state was restored; the only mutations were to seeded local rows, changed and changed back.

PR #1729, opened 2026-07-23
fix(mcp): drop MCP env vars from the build registry — they gate MCP, not PinPoint (PP-ogzs)

timothyfroehlich:
## The problem

PR #1727 put `MCP_BEARER_TOKEN` and `MCP_ADMIN_USER_ID` into `REQUIRED_PRODUCTION_ONLY`. That was my judgement error. The registry is a **deploy gate**, so this made an optional admin feature capable of taking production deploys down — and it immediately did: the first production deploy after that merge (`c417e9c3`, deployment `dpl_HDawxihXuc8JC697VPRDtwdoKguj`) hard-failed at `next build` because the vars weren't set in Vercel yet.

Blast radius was contained — Vercel doesn't promote a failed build, so production kept serving `cdd8a635` and stayed healthy, and the PR carried no migrations so `migrate:production` was a no-op. But `main` was un-deployable.

## The fix

These vars are required for **MCP** to work, not for **PinPoint** to work. `verify-token.ts` already fails closed when either is unset: `/api/mcp/mcp` 401s, logs `reason: "not_configured"`, and every other surface is unaffected. That's the correct degraded state for an optional single-user admin tool — and it's literally the §4.2 "degrades gracefully" criterion, which the ENV_VARS.md rows *already stated* while sitting in the §4.1 required table.

- **`next.config.ts`** — `REQUIRED_PRODUCTION_ONLY` back to `NEXT_PUBLIC_SITE_URL` only, with a comment recording why the MCP vars aren't there.
- **`docs/ENV_VARS.md`** — both rows moved §4.1 → §4.2; membership test stated on the §4.1 table; token generation/rotation guidance kept.
- **`docs/NON_NEGOTIABLES.md`** — CORE-SEC-009 said a var that "only logs or degrades" belongs in the registry, which is what invited this. Added the scope boundary (*would users be silently harmed if this were unset?* unsubscribe → yes; MCP → no) plus a note that a registry var must exist in Vercel **before** the PR merges, since the gate fires on the first deploy after merge, not in CI.
- **`.env.example`** — both optional everywhere, including production.

**Unblocks production deploys of `main` with no Vercel changes needed.**

## Verification

- Simulated the production assertion with the MCP vars unset → no missing groups, build proceeds.
- `pnpm run check` green (1632 tests / 201 files).
- Full `next build` completed clean. Not `preflight`: it includes `smoke`, which can't pass on this host (headless Chromium won't accept keyboard input on bazzite — PP-8b6j). CI owns E2E.

Follow-up bead PP-ogzs is repurposed to this change.


timothyfroehlich:
Claude review of head 19c5c1d — no serious findings

timothyfroehlich:
### Code review

Self-review, since this PR fixes my own judgement error and Copilot is out of quota until ~2026-08-01.

**Reviewed for:** whether removing these two vars from the registry is actually correct, and whether anything else in the repo still encodes "these are required".

Two issues found and fixed in `19c5c1d0` rather than left for follow-up:

1. **The add-a-var checklist encoded the failure.** `docs/ENV_VARS.md` §6 listed *"add it to the registry"* (step 4) before *"set it in Vercel"* (step 6). Followed in order, that registers a var whose value doesn't exist yet — which is precisely how `c417e9c3` broke. Steps swapped, with the reason stated inline.

2. **"If production-required" was undefined at the point of use.** That bare phrase is what misled me: it reads as "is a secret" / "is production-only" / "a feature degrades" when it means *PinPoint is broken without it*. Step 5 now carries the membership test and worked examples on both branches (`UNSUBSCRIBE_SIGNING_SECRET` → registry; MCP → §4.2).

`AGENTS.md` §2.1 item 20 picked up the same boundary, per its own instruction to stay in sync with `docs/NON_NEGOTIABLES.md` when a `CORE-*` rule is rewritten.

**Checked and clear:**
- No test, script, or config asserts the registry's contents — `check:config` (`scripts/check-config-drift.py`) doesn't read it, so nothing else needed updating.
- `verify-token.ts` is untouched; the fail-closed path this PR relies on is the one already covered by the 13 tests added in #1727, including `fails closed when MCP_BEARER_TOKEN is unset` and `fails closed when MCP_ADMIN_USER_ID is unset`.
- The stale `docs/plans/2026-07-22-...-handoff.md:77` instruction to register both vars is a dated record, left alone per AGENTS.md §8.

**Verified:** production assertion simulated with both MCP vars unset → no missing groups; `pnpm run check` green (1632 tests); full `next build` clean. Not `preflight` — it includes `smoke`, which cannot pass on this host (PP-8b6j).

**Residual risk:** MCP now has no deploy-time signal that it's unconfigured. That's the intended trade — the signal moves to the point of use, where the sole consumer gets a 401 and a `reason: "not_configured"` log line immediately.

PR #1733, opened 2026-07-24
docs(pinballmap): settle the listing redesign spec + refresh vendored llms.txt

timothyfroehlich:
Lands the settled design for the Pinball Map listing UI as a canonical spec, plus a refresh of PBM's vendored `llms.txt`. **Docs only — no source, no migrations.**

The spec is load-bearing: five open beads (PP-o355.7, .15, .19, .20, .21, .22) reference `docs/pbm-listing-redesign-refresher.md` by path for their state machine, availability matrix, and tie rule. Landing it early means implementation beads can be picked up independently without carrying the design on an unmerged branch.

## What's in the spec

- **Three concepts kept separate** — catalog match / listing / availability. Availability never drives listing.
- **Six derived states** (UNMATCHED, NOT_ON_PBM, NOT_LISTED, LISTED, MISSING_ON_PBM, UNSYNCED), computed at render from (catalog match, `listed`, `lmx`, snapshot) — never stored as a status column.
- **One-way auto-link** — the system links matched+unlisted machines whose title is on the lineup, at sync and at title-save. Nothing automatic ever unlists.
- **Availability × listing validity matrix** — valid / advise / invalid, where invalid is a hard flag that is allowed to exist and surfaced rather than auto-fixed.
- **Tie guard on duplicate titles** — drop availabilities where Listed is invalid, rank the rest by presence, unique winner holds the listing; ties pause sync and hide listing actions on both the machine settings page and the control room.

## Verified against PinballMap's source (new §6a)

The 7-day `lmx` soft-delete/resurrection window was previously an assumption the design leaned on. It's now checked against their Rails source (`api/v1/location_machine_xrefs_controller.rb`) and confirmed, along with four bonus guarantees: resurrect returns 200 vs 201 on mint, soft-deleted rows raise RecordNotFound, no purge task exists so `lmx` ids are never recycled, and their real rate limits are far looser than our self-imposed one-sync-per-hour.

## Vendored docs refresh

`llms.txt` re-fetched 2026-07-23. It now carries a `Last updated:` header (cheap staleness check for future refreshes) and adds an explicit client-side token ban — which **we already comply with**: all PBM access is server-only via `~/lib/pinballmap`, and no client component imports the live client. `robots.txt` unchanged.

## Note on CI

The `pnpm audit` gate is currently red repo-wide on three advisories published 2026-07-24 (js-yaml, postcss, brace-expansion) — unrelated to this PR, tracked in PP-2hax with a fix in flight. This branch will go green once that lands and main is merged in.

https://claude.ai/code/session_01Y39wAcWjeH7BKkB8p96eqo


timothyfroehlich:
## Claude review — head `a6f038ce`

Docs-only, 3 files, +241/−2. CI Gate green. **No blocking findings.** Verified rather than assumed:

- **The compliance claim in `docs/external/README.md` holds.** The refresh note asserts "all PBM access is server-only via `~/lib/pinballmap`, and no client component imports the live client." Checked: no file carrying `"use client"` imports `~/lib/pinballmap`, and every `pinballmap.com` reference in `src/` lives under `src/lib/pinballmap/`. PBM's new client-side-token ban is satisfied.
- **Vendored `llms.txt` provenance is updated correctly** (2026-07-18 → 2026-07-23) and the diff is confined to PBM's own additions — the `Last updated` header, the token ban, and the LMX `name`/`manufacturer`/`year` clarification. No hand-editing of vendored text.
- **§6a is the strongest part of the spec** — the 7-day resurrection window, the 200-vs-201 discriminator, and the rate limits are all cited to controller/line in PBM's source, with an honest dating caveat about `git log -S`. That's the right standard for an external-behavior assumption.

### Non-blocking, worth a follow-up commit

1. **§12's PP-nw4k warning is now false.** It tells the reader `origin/main`'s `pnpm-lock.yaml` has duplicate `@types/node@26.1.1` keys and that it's "worth resolving before starting build work here." That was fixed by **PR #1724** on 2026-07-22. Since five beads point at this file as their spec, that paragraph will send someone to fix an already-fixed problem. Recommend deleting it.
2. **§12's viewer command has a macOS path** (`/Users/froeht/Code/PinPoint/...`). This repo is on Linux now (`/var/home/froeht/Code/PinPoint`), so the command fails as written. The artifacts also point into `.claude/worktrees/pbm-list-unlist/`, which §1 itself calls stale.
3. **Framing vs. role.** The title says "morning refresher", the subtitle says "Read this after compacting," and the body reads as session scratch — but per the PR description this is load-bearing for six beads that cite it *by path*. Per AGENTS.md §8, canonical specs are authoritative and dated artifacts belong in `docs/superpowers/specs/`; this is in neither place. The risk is a future reader treating a canonical spec as disposable scratch. Either retitle it as a spec or move it where records live.

None of these affect correctness of the design, and docs are cheap to amend — flagging rather than gating.

—Claude

PR #1730, opened 2026-07-23
feat(settings): shareable settings sets — owner/community kinds, visibility, tournament tag (PP-tn6t)

timothyfroehlich:
## What

Turns machine settings sets from a flat "one preferred set" list into a full **ownership + visibility model**, reusing the collections sharing pattern. Each set now has:

- **Kind** (drives EDIT): **Owner set** (made by the machine owner — protected, owner+admin only) vs **Community set** (co-edited by technicians+, the owner, and admin).
- **Visibility** (drives VIEW): **Private draft** (creator-only, the default on create) vs **Public**.
- **Tournament** — an orthogonal, non-exclusive tag; tagging needs edit rights.
- **Owner's default** — the former "preferred"; one per machine, always an owner set + public.

Permissions are **role-based** (no per-person grants in v1 — deferred). Creating rides the existing `machines.settings.manage` matrix entry; per-set edit/publish/tag/default go through a new pure module `~/lib/machines/settings-permissions`.

## Model details

- **Auto-default:** the owner's *first* set (no existing default) auto-becomes the public Owner's default; tech-created sets never auto-default.
- **Duplicate** carries the Tournament tag, re-derives ownership (a tech's copy of an owner set becomes an editable community set), and starts as a private draft.
- **Migration `0059`:** adds `is_owner_set` / `is_public` / `is_tournament`; backfills existing sets → public/community and the existing preferred → owner set (preserves today's "everyone sees everything").
- **Timeline:** stops emitting the `settings_set_preferred` event (full type/historical-row removal is a follow-up bead).
- **Filter UX:** single-select category segment (All / Mine / Owner's / Community) + an independent Tournament toggle.
- **Description:** click-to-edit (keeps at-rest cards compact).

## 🔎 Please check these specifically

1. **`canViewSet` / `canEditSet` / `canSetOwnerDefault`** in `src/lib/machines/settings-permissions.ts` — the whole model hinges on these three. Note a private draft is creator-only even when it's a *community* set; community-ness only governs edit rights once the set is visible.
2. **Owner-set protection vs community co-editing** — techs must NOT be able to edit/delete/publish/tag/default an owner set, but CAN co-edit community sets (and the owner can edit community sets). Covered by the new integration tests.
3. **Auto-default heuristic** — only the owner's first set auto-defaults; verify a second owner set and any tech set stay private drafts.
4. **Publish refuses to unpublish the Owner's default** (it must stay public) — `publishSettingsSetAction`.
[...]

PR #1734, opened 2026-07-24
feat(pinballmap): read api_token from PINBALLMAP_API_TOKEN, delete the Vault path (PP-o355.23)

timothyfroehlich:
Closes PP-o355.23.

## Why

PP-uusr (#1705) shipped the PBM blanket `api_token` as env var → seed script → Supabase Vault → `SECURITY DEFINER` RPC → client. That design had a fatal gap: **the seed script is only wired into `db:reset`, which is forbidden on production**, so no production code path has ever been able to write the token. That is the root cause of prod sitting unprovisioned with PBM's mandatory-token gate flipping **2026-07-30**.

The `api_token` is a **platform capability, not tenant data** — PBM issues it to an approved account against a use-plan, i.e. to PinPoint-the-application, and every tenant inherits it without ever touching it. That makes it a deploy-time constant, so it now reads from a plain env var.

Deliberately **not** symmetric with the per-operator write creds (`outbound_email` / `outbound_token_vault_id`), which correctly stay in Vault: those are per-user identity arriving at runtime through a connect flow. Untouched here.

## What changed

- **`src/lib/pinballmap/api-token.ts`** reads `process.env` with no DB call — drops `createAdminClient`, the RPC, and the CORE-ARCH-011 `assertNotInTransaction` tripwire (moot once there is no external round-trip). Null-on-unset contract preserved; now also trims and treats whitespace-only as unset.
- **Removes a per-call DB round-trip.** `client.ts` built an admin Supabase client and issued an RPC on *every* construction just to read a constant.
- **Deletes dead security surface.** A `SECURITY DEFINER` function with a hand-rolled `auth.role()` guard (needed only because PostgREST re-exposes public functions) protecting a non-tenant value.
- **Stops backup leakage.** `db:seed:from-prod` pulls a prod dump into local dev; the vault secret rode along to every machine.
- **Migration 0059** deletes any linked vault secret (NULL pointer is a no-op), drops `get_pinballmap_api_token()`, then drops `api_token_vault_id` — in that order, so the secret is never orphaned.
- **Deletes** `supabase/seed-pinballmap-token.mjs`, its `package.json` script, its step in the `db:reset` chain, and the 70-line RLS integration test that guarded the now-gone RPC. Replaced by a 5-case unit test on the accessor (bug class G, cheapest layer).
- **Docs:** `ENV_VARS.md` row + rationale block, CORE-PBM-001, the AGENTS.md summary line, `.env.example`, and the `pinballmapState` schema comment.

## Registry boundary (deliberate)

`PINBALLMAP_API_TOKEN` stays **out** of the `next.config.ts` build registry. Without it PBM sync degrades and every other surface works, so it fails CORE-SEC-009's "is PinPoint broken without this?" test and belongs in `docs/ENV_VARS.md` §4.2. This is the same boundary #1729 tightened after #1727 hard-failed a production deploy six days ago.

Consequence: **no merge-order constraint.** Tim has already set the value in Vercel production; the build does not gate on it either way.

## Two things this does NOT do

- **Setting the token does not turn PBM on.** `pinballmap_state.enabled` defaults to `false` and gates the sync paths independently — that is the PP-o355.10 rollout.
- **Local dev and CI get nothing.** They have never had this var and must not: PBM is never reached from tests (CORE-TEST-006), and `PINBALLMAP_MODE` defaults to `mock` outside production.

## Accepted tradeoff: the dropped-column window

`migrate:production` runs during the build while the **previous** deployment still serves traffic. `getPinballMapState()` uses a bare `db.select()`, which Drizzle expands to an explicit column list including `api_token_vault_id` — so between the migration and promotion, the old code raises `column ... does not exist` on PBM settings renders and the hourly sync cron.

**Deliberately accepting this rather than splitting into two migrations:** PBM is dormant (`enabled = false`), the affected surface is admin-only, and the window is minutes. Recording it so it's a decision, not an accident.

## Verification

- `pnpm run db:reset` applies migration 0059 clean against a real local Supabase — 60 migrations applied, `api_token_vault_id` gone, `get_pinballmap_api_token()` gone, `outbound_email` / `outbound_token_vault_id` intact.
- `pnpm run check` green · `pnpm run build` green · integration **492 passed** · integration-supabase **104 passed**.
- Preflight's `smoke` step could not run on this host: headless Chromium on Bazzite rejects Playwright keyboard input, so `auth.setup.ts` fails before any spec (PP-8b6j) — pre-existing and unrelated. CI owns the E2E suite.

## Follow-up filed

**PP-w3d9** (P2, unrelated bug found en route): the Discord integration's vault orphan compensation calls `vault.delete_secret()`, which does not exist in the installed `supabase_vault` extension — it ships `create_secret` and `update_secret` only. The call sits in a best-effort try/catch, so it silently never cleans up. This migration therefore deletes from `vault.secrets` directly.

https://claude.ai/code/session_01WDeZYYZ5eMZPwRLyyB8VeF


timothyfroehlich:
## Code review — `/code-review medium` (PM pass on top)

**Verdict: the core change is right and I'd merge it after finding 1.** The env read is correct (trim, null-on-empty), dropping the `assertNotInTransaction` tripwire is safe because it's still enforced at the real side-effect boundary (`client-live.ts` guards all 9 network methods), the migration chain is consistent (`_journal` idx 59, snapshot `prevId` → 0058), and no stale `api_token_vault_id` / `get_pinballmap_api_token` / `db:_seed-pinballmap-token` references survive anywhere in the tree.

I verified findings 2 and 4 against the code myself rather than taking them on trust; both hold.

---

### 1. `drizzle/0059_overrated_inhumans.sql:18` — medium, fix before merge

`DELETE FROM vault.secrets` requires table-level DELETE privilege **even when the subquery matches zero rows** — Postgres checks the ACL at execution, not per row.

In production the pointer is NULL (this PR's own premise: prod was never provisioned), so the statement is a guaranteed no-op there — yet it can still abort the migration and fail the Vercel production build, since `vercel-build` runs `migrate:production` *before* `next build`. Hosted Supabase and preview-branch DBs run different vault extension versions, and preview branch users are already known to be privilege-restricted (per `scripts/migrate-production.ts`'s own comment about `CREATE SCHEMA`).

So it's the one statement in the file that can break a deploy for zero benefit.

**My fault, not the author's** — my design note on PP-o355.23 specified this raw `DELETE`. The intent was to avoid orphaning encrypted material, which is still worth doing, so I'd keep it but make it non-fatal:

```sql
DO $$ BEGIN
  DELETE FROM vault.secrets WHERE id = (
    SELECT api_token_vault_id FROM pinballmap_state WHERE id = 'singleton'
  );
EXCEPTION WHEN insufficient_privilege THEN NULL;
END $$;
```

Worth noting this also sidesteps PP-w3d9 (`vault.delete_secret()` not existing in the installed extension) — a raw table `DELETE` is the correct mechanism here regardless.

### 2. `drizzle/0059_overrated_inhumans.sql:32` — low, your call

**Verified:** `getPinballMapState()` (`src/lib/pinballmap/state.ts:27`) uses bare `db.select()`, which Drizzle expands to an explicit column list including `api_token_vault_id`.

`migrate:production` runs during the build while the *previous* deployment still serves traffic, so between migration and promotion the old code throws `column ... does not exist` on PBM settings renders and the hourly sync cron.

I'd **accept the window** rather than split into two migrations: PBM is dormant (`enabled=false`), the settings page is admin-only, and the window is minutes. But make it a deliberate decision, not an accident — worth one line in the PR description.

### 3. `docs/ENV_VARS.md:122` — low, but narrower than the review states

The review says recommending `--no-sensitive` on a "grants reads only" basis contradicts CORE-PBM-001 calling the token "the blanket access gate for reads _and_ writes."

**Half right.** The token is indeed sent on write requests — but it **cannot authorize a write by itself**: writes additionally require the per-operator `user_email` + `user_token`. So someone holding only this token gets reads. The real blast radius of a leak is **quota abuse** against our shared 120/min limit and traffic attributed to us — not data modification.

So keep `--no-sensitive` (that call stands), but fix the justification text, which is imprecise as written. Something like: "the token alone cannot authorize writes — those additionally require per-operator credentials — so the leak risk is quota abuse, not data modification."

### 4. `.env.example:104` + `src/lib/pinballmap/api-token.ts:24` — low here, but a real latent bug

**Verified:** `getPinballMapMode()` (`src/lib/pinballmap/config.ts:17`) falls back to `NODE_ENV === "production" ? "live" : "mock"`, and `NODE_ENV` **is** `production` in Vercel preview builds and at preview runtime.

So previews resolve to the **live** client while `PINBALLMAP_API_TOKEN` is scoped production-only — real requests to pinballmap.com with the header omitted, which will 401 once the July 30 gate flips. That's also uncomfortably close to CORE-PBM-001's "never reach PBM outside the sanctioned path."

Pre-existing, so **out of scope for this PR** — but this PR newly asserts the mock default as a safety guarantee in two comments. Please either soften those two comments or drop the claim; I'm filing a separate bead for the actual fix (gate on `VERCEL_ENV === "production"` instead of `NODE_ENV`).

---

**Also, merge-order heads-up:** #1730 (PP-tn6t) also claims `0059` (`0059_cultured_bloodaxe.sql`). Whichever of the two lands second must regenerate — never hand-resolve `drizzle/meta` (see the `pinpoint-migration-conflicts` protocol).

—Claude

timothyfroehlich:
Thanks — all four verified against the code before acting, and all four accepted. Fixed in 1c33d34b.

### 1. Vault DELETE privilege — **fixed**, and confirmed empirically

You were right that the ACL is checked at execution rather than per row, so I proved it rather than reasoning about it. Against a deliberately restricted role on the local stack:

```
NOTICE:  UNGUARDED: would ABORT the migration -> permission denied for table secrets
NOTICE:  GUARDED: swallowed, migration continues
```

The unguarded `DELETE` raises `permission denied for table secrets` **despite matching zero rows** — exactly the deploy-breaking case, since `vercel-build` runs `migrate:production` before `next build`. Now wrapped in a `DO` block catching only `insufficient_privilege` (with a `RAISE NOTICE`), so anything else still fails loudly. Full 60-migration chain re-applied clean afterward.

### 2. Transient dropped-column window — **accepting the window**, now documented

Confirmed your reading: `getPinballMapState()` uses bare `db.select()`, which Drizzle expands to an explicit column list. Taking your recommendation to accept rather than split the migration — recorded in the PR description as a deliberate call, with the reasoning (PBM dormant at `enabled=false`, admin-only surface, minutes-long window).

### 3. `--no-sensitive` justification — **fixed**

Agreed the call stands and the justification didn't. Reworded to name the actual blast radius — quota abuse against the shared 120/min limit and traffic attributed to us — rather than the imprecise "grants reads only".

### 4. Mock-default claim — **fixed, but narrower than stated**

Verified the mechanism: `NODE_ENV === "production"` is true in Vercel preview builds, so previews do resolve to the live client. Real latent bug, agreed it's out of scope here.

One correction: only **one** of the two comments makes that claim. `.env.example` said `PINBALLMAP_MODE` "defaults to `mock` outside prod" — fixed, and it now explicitly warns against treating "not production" as a mock guarantee. The `api-token.ts` comment doesn't mention the mock default at all; it cites `pinballmap_state.enabled`, which is unaffected by the `NODE_ENV` issue. Left it alone rather than softening a claim it wasn't making — say the word if you'd rather it carry the warning too.

### Merge-order

Acknowledged, and it's in the huddle daily. #1730 also claims `0059`; whichever lands second regenerates per `pinpoint-migration-conflicts` — never hand-resolving `drizzle/meta`.

### Also worth knowing

Your finding 1 note that a raw table `DELETE` is correct regardless of PP-w3d9 is now confirmed at the source: I queried **PinPoint-Prod** directly — `supabase_vault` **0.3.1**, functions `create_secret, update_secret` only. No `delete_secret` in production either. Same query showed prod's `vault.secrets` holds exactly one row (the Discord token, correctly referenced — no orphans), and `pinballmap_state.api_token_vault_id` **is NULL**, independently confirming this PR's premise and that the cleanup is a no-op in prod.

—Claude

timothyfroehlich:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Five independent passes over head `1c33d34b` (CLAUDE.md/NON_NEGOTIABLES compliance, shallow bug scan, git-history context, prior-PR review comments, code-comment accuracy). Specifically confirmed: migration 0059 fully reverses 0057 in correct dependency order; the `DO $$ ... EXCEPTION WHEN insufficient_privilege` guard only rolls back its own sub-transaction, so unrelated errors still fail loudly; dropping `assertNotInTransaction` does not reintroduce PP-2053, since the CORE-ARCH-011 tripwires on the actual HTTP boundary in `client-live.ts` are untouched; `PINBALLMAP_API_TOKEN` is absent from the `next.config.ts` registry, respecting the boundary #1729 established; and the deleted RLS test only covered the RPC this PR drops.

One finding was raised and filtered as a pre-existing issue (confidence 0): the migration comment notes that supabase_vault 0.3.1 has no `vault.delete_secret()`, which contradicts calls in `src/app/(app)/admin/integrations/discord/actions.ts` and `supabase/seed-discord.mjs`. This PR touches neither file, and the comment already cites the tracking bead. That is real but out of scope here.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1738, opened 2026-07-25
feat(hooks): enforcement layer for the Claude-native context system (PP-22e4)

timothyfroehlich:
> **Resolved.** The `.claude/settings.json` registrations for both deleted hooks are gone as of `8bc1bca4` (Tim authorized the edit directly). Four `PreToolUse` matcher groups remain, every registration backed by a file on disk — verified programmatically, since `verify-guard-stack.cjs` would not have caught the mismatch on its own (one-directional check, filed as **PP-ncla**).

Enforcement layer for the Claude-native context system (**PP-22e4**), plus the design spec it implements.

## What changed since the last review

This PR was originally scoped as **three** new PreToolUse hooks. A separate full review of PinPoint's blocking hook layer reached an organizing conclusion that cut two of them:

- A rule about an ***effect*** should move to the **lowest layer that can express it**, where it becomes airtight and can't misfire on prose.
- A rule about ***intent*** ("who/what initiated this action") can only be expressed at the hook layer, and should be tuned for **low false positives**, not completeness.
- The threat model throughout is explicitly **non-adversarial** — a well-meaning agent doing something dumb, not an attacker evading a filter. No evasion-resistance machinery.

Also, `#1740` and `#1741` merged to `main` in the meantime and this branch has been synced with both (clean merge, no conflicts).

## One hook ships

| Hook | Matcher | Rule | Status |
|---|---|---|---|
| `block-drizzle-push.cjs` | `Bash` | CORE-ARCH-009 | **Deleted.** Superseded by `#1741`: `assertNotDrizzlePush()` is called as the first statement in `drizzle.config.ts`, ahead of `loadEnvConfig()`/`POSTGRES_URL`. Drizzle Kit always reads its own config first, so this is airtight regardless of invocation shape and — unlike the hook — can't fire on prose that merely names the command. (The hook version shipped without effective-command resolution and blocked an unrelated `bd comments add` on its first real firing; moot now that it's gone.) |
| `block-loopback-literal.cjs` | `Write\|Edit` | CORE-SEC-008 | **Deleted.** Tim's call: `localhost`-not-`127.0.0.1` is already a documented non-negotiable, and the hook was structurally blind to the write paths that matter most — generators and heredocs. Deleted outright, no replacement content-check hook. |
| `block-worktree-dispatch-from-linked.cjs` | `Agent` | [#47548](https://github.com/anthropics/claude-code/issues/47548) | **Kept.** Squarely an *intent* rule that can't move to a lower layer. Tuned for low false positives: reads structured fields (`tool_name`, `tool_input.isolation`), three-state on worktree classification so non-repos/git errors fail **open**, takes cwd from the payload rather than `CLAUDE_PROJECT_DIR`. |

Added `block-worktree-dispatch-from-linked.cjs` to `EXPECTED_GUARD_HOOKS` in `verify-guard-stack.cjs` and its test fixture list — it shipped originally without either.

## `check:rule-ids` gate — two defects fixed

1. **Range-shorthand blindspot.** Citations written as `CORE-RESP-001..004` (AGENTS.md §2.1 uses this for `RESP-001..004` and `FORM-001..006`) only ever extracted the range's *start* ID — the end ID and every interior ID were silently never checked against the catalog. A rename/deletion of e.g. `CORE-RESP-003` would go undetected. Fixed by expanding `..`-ranges in `extract_ids()`. Verified by reverting the fix and confirming 3 tests go red, including a new fixture test that plants a real gap (`CORE-RESP-002`/`003` missing from a fake catalog, cited only via range) between a broken gate and a working one.
2. **Vacuous real-repo test.** `TestRealRepo::test_no_unknown_ids_in_this_repo` asserted `main(...) == 0` against the live repo — true even if `collect_citations` silently returned `{}` (e.g. a broken glob), since "no citations" trivially implies "no unknown IDs." Added non-triviality assertions (catalog has 50+ IDs, at least one real citing file found, 20+ total cited IDs). Verified by breaking `CITING_SOURCES` and confirming the test now fails with "expected at least one real citing file to be found."

## Spec re-measured

`docs/superpowers/specs/2026-07-24-claude-native-context-system-design.md` §2's character/line counts were taken against a stale root checkout; re-measured against the current tree (post `#1740`/`#1741` merge) and corrected. §6/§11/§12 rewritten to reflect one hook shipping instead of three, with the two drops explained and pointers to their replacements.

## Testing

`pnpm run check`: types/lint/format/`check:rule-ids`/pytest all green. Full vitest run showed 3 unrelated flaky timeouts (`global-setup.test.ts`, `signup-form.test.tsx` — 5000ms timeouts, nothing to do with this PR's files) under confirmed host memory pressure at the time (the memory-pressure gate itself was blocking new heavy commands, swap at 11.7GB). All hook and `check_rule_ids` tests — the actual subject of this PR — passed cleanly, including `verify-guard-stack.test.ts`.


timothyfroehlich:
## Review 1 of 3 — hook classifiers (independent subagent, read-only)

`block-worktree-dispatch-from-linked.cjs` came back **clean** — `worktreeKind()` verified correct across main root, main subdirs, linked worktree, linked subdirs, non-repos, missing cwd, and bad types. The `path.resolve` on both lines is load-bearing (from a main subdir git returns `--git-dir` absolute but `--git-common-dir` relative) and correct.

`block-loopback-literal.cjs` is broadly sound — **zero** of the eight currently-scoped files contain the literal today, so nothing in the repo is newly blocked, and the lookbehind/lookahead behaves.

**`block-drizzle-push.cjs` needs a rewrite.** Three HIGH findings:

**1. Newlines are not a segment separator.** The split is on `&&`, `||`, `;`, `|` only. A multi-line Bash block collapses into one segment whose first token is the effective command, so a three-line script ending in a real invocation is allowed. Multi-line blocks are the most common way an agent runs a sequence — this is the widest hole.

**2. One-token evasions of the runner/wrapper resolver**, none requiring intent — several are ordinary shapes: `pnpm exec -- drizzle-kit push`, `npm exec -- …` (the *documented* npm form), `npx -y …`, `pnpm --filter=web exec …`, `npx drizzle-kit@0.31.10 push`, `sh -c "…"`, `( … )`, `{ …; }`, `echo hi & …`, `$(…)`, backticks, `node node_modules/drizzle-kit/bin.cjs push` (drizzle-kit's actual bin path), `dotenv -e .env -- …`.

**3. The false positive I claimed to fix is only fixed for one string.** Segmenting ignores quoting, so any `;`, `|`, or `&&` **inside a quoted argument** starts a new segment with `drizzle-kit` as its first token. `bd comments add PP-1 "do not do this; drizzle-kit push corrupts meta"` still blocks. So does a `git commit -m` message containing `&&`. The reviewer hit this live while probing. The hook remains noisy against exactly the workflow — bead comments, PR comments, commit messages — the rewrite was meant to stop bothering.

Shared root cause: regex-splitting a shell string with no quote awareness. The fix needs both quote-span stripping before segmenting *and* the missing separators; they're complementary, not alternatives. The existing test file covers none of the three.

**MEDIUM:** the three new hooks were never added to `EXPECTED_GUARD_HOOKS` in `.claude/hooks/verify-guard-stack.cjs`, so the guard that exists to detect "settings.json silently dropped a hook" won't notice if any of them is dropped.

**LOW:** a literal `null` JSON payload throws in two hooks (exit 1 + stack trace instead of clean exit 0) — still fail-open, since only exit 2 blocks. And `scripts/worktree_setup.py` is unscoped but *generates* both scoped file types, so editing the generator to emit the literal would propagate it to every worktree unseen — the one genuinely load-bearing gap in the loopback scope.

Not merging until at least the three HIGH items are fixed. Two more reviews pending.

timothyfroehlich:
Claude review of head d878974 — Re-review for head d8789748 (two commits since the prior APPROVE). f363acf0 (PP-07g0, authored by a subagent, not me): reviewed the diff and independently verified behavior — descending CORE-RESP-004..001 now expands to all four IDs, ascending unchanged, degenerate 002..002 yields one ID with no false flag; new exit-3 path names file+range; _glob_paths refactor is behavior-preserving; 32/32 pytest, check:rule-ids exit 0 on the real repo. d8789748 (mine, authorized by Tim as a small change): 6-line enabledPlugins deletion; verified permissions/hooks/extraKnownMarketplaces/enableAllProjectMcpServers byte-identical to parent via jq -cS per key, JSON valid. Cosmetic nit only, not filed: a range repeated verbatim in one file prints twice in the exit-3 report.

PR #1745, opened 2026-07-26
fix(huddle): don't let the rotation notice short-circuit registration (PP-2m3l)

timothyfroehlich:
Closes PP-2m3l.

## The bug

`scripts/hooks/huddle-session-start.sh` printed the "rotation needed" notice and then `exit 0`. The identity / registration block lives ~60 lines further down, so it was **unreachable** whenever rotation was pending.

SessionStart fires exactly once per session, so there was no second chance to re-prompt: every session that started on a new day *before* rotation ran stayed unregistered for its entire life.

Two symptoms follow from that:

1. **Self-filter breaks.** An unregistered session sees its own huddle posts injected back to it as if from a peer — confusing, and it burns context re-reading your own updates.
2. **Attribution degrades.** `inject-beads-actor.cjs` resolves the actor from `session-names.json` and falls back to the literal `Claude` on a miss, so posts land as generic "Claude" instead of the session name.

Blast radius on 2026-07-24 alone was at least four sessions (`Claude-SessionBriefing`, `Claude-PbmTokenEnvVar`, `Claude-PbmTieGuard`, `Claude-DoctorAudit`) — all posted to the daily bead, none were in `session-names.json`.

**Observed live again tonight** while this PR was in flight: a `Claude-Briefing2` ledger update at 01:40Z landed attributed to a bare "Claude" despite that session being registered — symptom (2), reproduced in the wild.

## The fix

Rotation and registration are independent concerns, so the notice now sets a `ROTATION_PENDING` flag and **falls through** instead of exiting. The rotation notice still comes first, so it keeps its urgency; the identity/registration block follows it.

The one thing still gated behind the flag is `huddle_reconcile_today`, whose documented contract is explicitly the up-to-date path — with rotation pending, today's daily does not exist yet, so there is nothing to dedup and the rotation subagent is about to create it under its own lock.

## Tests

New `scripts/tests/test_huddle_session_start.py` (7 cases, runs under `pnpm run check:pytest`). Each builds a throwaway git repo so `huddle_state_dir` resolves locally, stubs `bd` on `PATH` to serve canned JSON per bead id, and pipes a SessionStart payload on stdin.

- **Rotation pending** — emits both the rotation notice and the registration block, in that order, with the copy-paste `register` command; same for an already-registered session; recent-activity digest still injects.
- **Up to date** — unchanged behaviour for both registered and unregistered sessions (regression baseline).
- **Early exits that must keep short-circuiting** — `source=compact` suppresses the identity block but keeps the rotation notice; a `/subagents/` transcript path emits nothing at all.

Verified as a real regression guard: with the hook reverted to its pre-fix state, the three rotation-pending cases fail and the four baseline cases still pass.

## Also

- One clarifying sentence in `.agents/skills/pinpoint-huddle/SKILL.md` so the documented rotation notice no longer reads as terminal.
- Filed **PP-txet** for a separate, pre-existing defect the new tests surfaced: an empty `session_id` is misparsed as the `source` value (space-joined fields + default-IFS `read`), defeating the "no session_id → silently exit" guard. Deliberately **not** folded in — this PR is a scoped control-flow bugfix.

The bead's `UNEXPLAINED` observation (a post landing authored "Tim Froehlich", the git `user.name` fallback) is explicitly **out of scope** here and remains un-theorised, per the bead's own "worth reproducing before theorising".

### DIFF
--- a/.agents/skills/pinpoint-huddle/SKILL.md
+++ b/.agents/skills/pinpoint-huddle/SKILL.md
@@ -33,6 +33,8 @@ Run the command, then start or resume your session normally. Bootstrap is idempo
 
 **Stop and dispatch the rotation subagent before continuing your work.** See the "Rotation" section below for the exact dispatch.
 
+This notice does **not** replace your identity block — rotation and registration are independent, so the identity/registration block still follows it. Dispatch the rotation subagent first, then register if you're new. (Before PP-2m3l the notice short-circuited the rest of session-start, leaving the day's first sessions permanently unregistered.)
+
 ### Identity announcement (you're registered)
 
     ## Huddle identity
--- a/scripts/hooks/huddle-session-start.sh
+++ b/scripts/hooks/huddle-session-start.sh
@@ -183,8 +183,13 @@ fi
 # Skipped while rotation is pending — preserving the pre-PP-2m3l behaviour, where
 # this only ever ran on the up-to-date path. The function keys off root notes'
 # `today_bead.date`, which is still YESTERDAY's date until rotation runs, so
-# calling it now would reconcile (and close dupes of) a stale day. Rotation
-# repoints that field under its own lock; the next session start reconciles.
+# calling it now would reconcile (and close dupes of) a stale day.
+#
+# Note this net never collapses stale-day duplicates at all: it only ever targets
+# the date root notes point at, and rotation moves that pointer to today. That
+# gap is pre-existing (this call was unreachable on the rotation-pending path
+# before PP-2m3l too) and out of scope here — do NOT read the skip as "the next
+# session start will catch it".
 if [[ -z "$ROTATION_PENDING" ]]; then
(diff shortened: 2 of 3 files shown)


timothyfroehlich:
Claude review of head 50bb710 — Lead review (Copilot returned only a quota-limited placeholder). Verified: exit 0 replaced by a ROTATION_PENDING flag so rotation and registration both emit; huddle_reconcile_today correctly gated behind the flag since root notes still point at yesterday, with an honest in-code note that the stale-day gap is pre-existing and NOT caught by the next session start; the placeholder warning is a real fix (I hit this exact literal-<today-bead-id> problem myself tonight). 350-line pytest asserts both blocks present AND ordered, and proves the reconcile skip behaviorally by asserting no update/close/comments bd subcommands fire. Python os.chmod for the PATH stub sidesteps the shell chmod denial. No findings.

PR #1746, opened 2026-07-26
fix(scripts): orphan sweep reports Docker UNKNOWN instead of a false zero (PP-5o7b)

timothyfroehlich:
Fixes PP-5o7b.

## The bug

`scripts/worktree_orphan_sweep.py` enumerated Supabase volumes with

```
docker volume ls --format '{{.Name}}|{{.Label "com.supabase.cli.project"}}'
```

That template is not portable. On this host `docker` is Podman, whose
`*types.VolumeListReport` has no `Label` method, so the command exits **125**:

```
Error: template: ls:1:23: executing "ls" at <.Label>: can't evaluate field Label in type *types.VolumeListReport
```

`_docker_label_query` caught that and returned `[]`. The sweep then reported:

```
Note: docker query failed (...); skipping Docker sweep.
Orphan Supabase Docker projects: 1
  - pinpoint-feat-mcp-bearer-token: 5 container(s), 0 volume(s)
  removed 5 container(s) for pinpoint-feat-mcp-bearer-token
```

`0 volume(s)` is a **false zero** — it reads as "there are none". `--apply`
removed the containers, declared success, and left the volume on disk. Worse:
the SessionStart hook runs this sweep in dry-run every 6h and nudges only on
what it counts, so a host with hundreds of MB of leaked *volumes* and no leaked
*containers* looked perfectly clean. That's how ~557 MB across 8 volumes
accumulated unseen.

A false zero is worse than an error, so this PR fixes both halves.

## 1. Enumeration works on both engines

Volumes are now listed with the label **filter** — honoured by Docker and
Podman alike — and resolved to their project with `docker volume inspect`,
whose raw `.Labels` map both engines expose:

```
docker volume ls --filter label=com.supabase.cli.project --format '{{.Name}}'
docker volume inspect --format '{{.Name}}|{{index .Labels "com.supabase.cli.project"}}' <names…>
```

Two calls on purpose. A volume removed by another agent's cleanup *between* the
`ls` and the `inspect` is tolerated (that name is simply dropped) rather than
blinding the whole sweep — this host runs 30+ worktrees concurrently. A real
template/daemon failure produces no parseable output at all and still raises.

Verified against the live daemon read-only: 3 labeled volumes found, correctly
attributed, matching `docker volume ls --filter` by hand.

## 2. A failed query is UNKNOWN, never zero

`get_supabase_resources_by_project` now returns a `DockerSweepResult` carrying
an explicit `unknown_reason` instead of an indistinguishable empty dict. Any
enumeration failure sets it — **including a partial one**, since a project whose
containers listed but whose volumes didn't is exactly the false-count case.

Consequences, all of them load-bearing:

- The UNKNOWN warning prints **even under `--quiet`**, so the SessionStart nudge
  cannot under-report. That nudge is the standing alarm this bug silenced.
- The quiet one-liner says `Supabase Docker project orphans UNKNOWN` rather than
  a count.
- **`--apply` removes no Docker resource at all** in that state, and says
  `Docker sweep SKIPPED (state unknown)`. Slot-manifest reclamation is
  independent and still runs.
- The process exits non-zero (`EXIT_DOCKER_UNKNOWN`). The SessionStart hook
  already swallows exit codes behind `|| true`, so session start is unaffected.
- An absent `docker` binary stays a genuine zero — no binary means no Docker
  resources, and crying UNKNOWN there would be noise.

## What deliberately did not change

Orphan classification. A project is an orphan because it has **no active
worktree**, never because it has no container. `supabase_db_pinpoint-worktree-bridge-cse-01-8ea3ea5e`
has no container but belongs to a live worktree whose Supabase is merely
stopped; Docker counts it as unused. A container-presence heuristic — or
`docker volume prune` — would destroy a live worktree's local database. The
existing check against the active-worktree `project_id` set is correct and is
now covered by a test that pins it.

## Testing

`scripts/tests/test_worktree_orphan_sweep.py`, 17 tests, mocked at the
`subprocess` boundary — no real daemon, no dependence on local Docker state.
The regression that matters:

- a failing `docker volume ls` (the real Podman exit-125 stderr, a daemon-down
  error, and a bare non-zero exit) yields `is_unknown`, not an empty result
- `main()` dry-run prints `UNKNOWN` and **never** `0 volume(s)`, never claims
  "No orphans found.", and exits non-zero
- the `--quiet` nudge says UNKNOWN rather than `0 … orphan(s)`
- `--apply` under UNKNOWN issues **no** `docker rm` and **no** `docker volume rm`,
  while still deallocating slot orphans
- a volume whose project is an active worktree is never removed

Confirmed red against the pre-fix code: replaying the mock through the old
`main()` reproduces `1 container(s), 0 volume(s)` and exit 0.

`pnpm run check` passes (127 pytest, 1655 unit, lint/format/types/shellcheck).

## Safety note

Everything here was verified with **dry-run only** plus read-only `docker volume ls`
/ `docker ps` queries. `--apply` was never executed — several agents are running
live worktrees and Supabase stacks on this host, and removing one of their
volumes would destroy their work.

### DIFF
--- a/.claude/hooks/session-start-orphan-sweep.sh
+++ b/.claude/hooks/session-start-orphan-sweep.sh
@@ -6,7 +6,11 @@
 # orphan slot manifest entries and orphan Supabase Docker resources from
 # the failure modes documented in PP-qlzu (rm -rf without the hook, Claude
 # in Web sandbox sessions). When orphans are found the sweep prints a
-# single-line nudge to stderr telling you to run `--apply` manually.
+# single-line nudge to stderr telling you to run `--apply` manually. If Docker
+# can't be enumerated the nudge says UNKNOWN rather than reporting zero — a
+# false zero here is what let ~557 MB of orphan volumes accumulate unseen
+# (PP-5o7b) — and the sweep exits non-zero, which this hook deliberately
+# swallows so session start is never blocked.
 #
 # Why dry-run (not auto-apply): SessionStart fires on every Claude Code
 # session and can affect Docker resources across the host; we want the
--- a/scripts/tests/test_worktree_orphan_sweep.py
+++ b/scripts/tests/test_worktree_orphan_sweep.py
@@ -1,0 +1,414 @@
+"""Regression tests for worktree_orphan_sweep.py Docker enumeration (PP-5o7b).
+
+The bug these lock down: `docker volume ls --format '{{.Label "…"}}'` blows up on
+Podman (`can't evaluate field Label in type *types.VolumeListReport`, exit 125),
+and the old code swallowed that into an empty list. The sweep then printed
+`0 volume(s)` — a *false zero* that reads as "there are none" — so `--apply`
+removed the containers, claimed success, and left ~557 MB of volumes on disk
+while the 6-hourly SessionStart nudge stayed silent.
+
+So the invariant under test is not "volumes are listed", it is **a failed query
+must report UNKNOWN, never 0**, and `--apply` must not touch Docker when it
+can't see. Docker is mocked at the subprocess boundary; these tests never touch
+the real daemon.
+"""
+
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+
+import worktree_orphan_sweep as sweep  # noqa: E402
+
+LABEL = "com.supabase.cli.project"
+
+# The exact stderr the real Podman CLI emits for the old broken template.
+TEMPLATE_ERROR = (
+    'Error: template: ls:1:23: executing "ls" at <.Label>: '
+    "can't evaluate field Label in type *types.VolumeListReport"
+)
+
+
+def _kind(args: list[str]) -> str:
+    """Classify a docker argv so the stub can answer per-subcommand."""
+    if not args or args[0] != "docker":
+        return "other"
+    head = tuple(args[1:3])
+    if head == ("volume", "ls"):
+        return "volume_ls"
+    if head == ("volume", "inspect"):
+        return "volume_inspect"
+    if head == ("volume", "rm"):
+        return "volume_rm"
+    if args[1] == "ps":
+        return "ps"
+    if args[1] == "rm":
+        return "container_rm"
+    return "other"
+
+
+class DockerStub:
+    """Stands in for subprocess.run, answering docker calls from a script."""
+
+    def __init__(self, **responses: object) -> None:
+        # Each response is either a CompletedProcess-ish tuple
+        # (returncode, stdout, stderr) or an exception instance to raise.
+        self.responses = responses
+        self.calls: list[list[str]] = []
+
+    def __call__(
+        self, args: list[str], **_kwargs: object
+    ) -> subprocess.CompletedProcess[str]:
+        self.calls.append(list(args))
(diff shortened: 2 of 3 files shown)


timothyfroehlich:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Five parallel reviewers (CLAUDE.md/AGENTS.md compliance, shallow bug scan, git-history regression check, prior-PR review comments, comment accuracy). Nothing cleared the confidence bar. Verified specifically:

- `--apply` cannot remove a Docker resource while state is UNKNOWN — `orphan_projects` derives from `by_project`, which is structurally empty on that path, and the removal loop is a no-op. Both early returns in `main()` carry `EXIT_DOCKER_UNKNOWN`, and the UNKNOWN warning prints before either, unsuppressed by `--quiet`.
- Orphan classification is untouched: a project is an orphan because it has no *active worktree*, never because it has no container. The config.toml-based `project_id` resolution added in response to [#1409's Copilot review](https://github.com/timothyfroehlich/PinPoint/pull/1409) is intact.
- The original "never block on Docker being down" intent from PP-qlzu is not regressed — that guarantee lives in the hook's `timeout` + `|| true` + `exit 0`, which this PR doesn't touch.
- Podman claims verified against the real binary: `volume ls --format '{{.Label …}}'` exits 125; the `--filter label=` + `volume inspect '{{index .Labels …}}'` pair works; `docker ps --format '{{.Label …}}'` works, so scoping the two-call workaround to volumes only is correct.

Two sub-threshold observations, recorded rather than actioned:

1. `get_supabase_volumes()` drops a volume whose `inspect` fails for a reason other than a race, rather than escalating to UNKNOWN. The failure direction is safe (under-counts orphans, never causes a wrong removal) and self-heals on the next sweep. Comment tightened in 8d95c506 to describe every path that reports UNKNOWN.

https://github.com/timothyfroehlich/PinPoint/blob/8d95c506c9647f2ba66d2b4d51fe6adc29a66969/scripts/worktree_orphan_sweep.py#L240-L268

2. `scripts/worktree_cleanup.py` still treats a failed `docker volume ls` as an empty list — the same false-zero shape, in a sibling code path. It is *not* hit by this bug (it uses `-q`, not the Podman-incompatible template), and PP-5o7b is scoped to the sweep, so it's left alone. Filed as a follow-up rather than widened into this PR.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

timothyfroehlich:
Claude review of head 8d95c50 — Lead review (Copilot returned only a quota-limited placeholder). Verified the load-bearing safety claim: with Docker unknown, by_project is empty so the --apply reclaim loop body never executes -- no containers or volumes are touched. That rests on the invariant 'unknown => by_project empty', which is documented but not enforced; I checked all three DockerSweepResult construction sites and every one upholds it, so the concern is theoretical. Also verified: DockerNotInstalledError (genuinely zero) is correctly separated from DockerUnavailableError (unknown); the UNKNOWN line is never suppressed by --quiet; EXIT_DOCKER_UNKNOWN returns non-zero even when there are no slot orphans; the quiet nudge prints UNKNOWN instead of a count. Root cause is sharper than the bead's -- Podman, not 'this Docker version', which fits the Bazzite host. 414 lines of pytest mocked at the subprocess boundary. Optional hardening, non-blocking: a __post_init__ assert that unknown_reason implies empty by_project would make the safety structural rather than conventional.

PR #1748, opened 2026-07-26
fix(workflow): treat cancelled CI runs as superseded, not failed (PP-r63o)

timothyfroehlich:
Fixes PP-r63o.

`scripts/workflow/pr-watch.py` is the canonical CI watcher — AGENTS.md tells every agent to use it instead of hand-rolling a polling loop — so a false "✗ CI — failed" from it misleads every agent in the project. It had only two conclusion classes, passing and everything-else, so `cancelled` fell into the fail-safe branch.

Cancellation is routine here: pushing a second commit cancels the in-flight run via concurrency groups, and the Preview Auto-Resync workflow cancels itself the same way. A normal push-twice cycle therefore emitted a spurious failure plus an artifact whose log body read `(no log available)` — a cancelled run has no failed step to log.

It was a hard blocker, not just noise. Once any run for the current head SHA ended cancelled, the completed-runs scan in `main()` re-derived it from the API on every subsequent invocation and exited 1. `--force` skips only the readiness pre-check, not that scan, so there was no way to clear it. On PR #1734 the self-cancelling Preview Auto-Resync poisoned the watcher permanently.

## Part 1 — cancelled is its own class

`_is_passing` / `_is_superseded` / `_is_failing` replace the inline set-membership tests and normalise case (`gh run list` reports conclusions lowercase, `statusCheckRollup` uppercase). Every call site routes through them:

- `watch_run` emits `⊘ <name> — superseded (cancelled)`, records no failure and writes no artifact.
- The startup completed-runs scan excludes cancelled from `early_failures`, so a self-cancelled side workflow no longer hard-exits before watching starts.
- The no-active-runs fallback drops its duplicate hardcoded pass tuple.

`_is_failing` is fail-safe by construction: anything not recognisably passing and not a supersession is a failure, including an empty conclusion. GitHub can briefly report a run as `completed` before its conclusion populates, and every caller gates on `status` first.

## Part 2 — head-SHA-scoped pre-check

- `_ci_gate_state` returned the **first** rollup entry named `CI Gate`. GitHub scopes the rollup to the head commit, but that commit can carry more than one CI Gate check — a re-run or a concurrency-cancelled run leaves its superseded entry behind next to the live one. It now prefers a non-superseded entry, then the most recent, so a cancelled leftover can't shadow the live gate.
- `_pre_check_blocking` no longer treats a cancelled gate as "already failed"; it lets the watch loop wait for the replacement run. **This is the case `--force` was being used to work around.**
- `_finalize_via_ci_gate` gives a cancelled gate a bounded grace period (`SUPERSEDED_GATE_GRACE`, 180s) for a replacement to post, then reports it as superseded rather than failed. Without this the false alarm would just move to the final step.
- `run_audit` (`--check-ready`) still reports a cancelled gate as not-ready — it genuinely isn't green — but says `cancelled (superseded) — needs a re-run or a new push` instead of implying a broken test.

The run-list scans in `main()` were already head-SHA-filtered; that filter is now documented and pinned by a regression test.

## Tests

32 tests in `scripts/tests/test_pr_watch.py`, run by `pnpm run check:pytest`. Mocked at the `gh` CLI seam (`pr_watch.gh`, plus `subprocess.Popen` for the streaming `gh run watch`), so nothing reaches GitHub — CORE-TEST-006. **22 of them fail against the pre-fix script**, verified by checking out `origin/main`'s copy of the script and re-running. Fixtures use the real run shapes from PR #1734 (run `30133783967`, the self-cancelled Preview Auto-Resync at the current head SHA, and run `30132766363`, the CI run cancelled by a main-sync push at an older SHA).

## Out of scope, worth a follow-up

Claude-GhaFlakeTriage root-caused a different false-failure shape tonight: a `gh` API installation-quota 403 (5000/hr, shared across all agent sessions on this repo) redded run `30132971643` — harden-runner died on the identical 403 twenty-eight seconds before setup-cli did. That is adjacent to this bead but **not** fixable in this classification code: a quota 403 makes jobs genuinely conclude `failure`, so pr-watch is reporting it accurately. Distinguishing it needs log inspection, not conclusion classification. Not built here.

## Verification

`pnpm run check` exits 0 (142 pytest, 1655 vitest).

### DIFF
--- a/scripts/tests/test_pr_watch.py
+++ b/scripts/tests/test_pr_watch.py
@@ -142,10 +142,16 @@ def test_failing_conclusions(conclusion):
 
 @pytest.mark.unit
 @pytest.mark.parametrize("conclusion", ["", None])
-def test_undecided_conclusion_is_not_a_failure(conclusion):
-    """An empty conclusion means "not decided yet" — status carries that state."""
-    assert not pr_watch._is_failing(conclusion)
+def test_completed_with_empty_conclusion_fails_safe(conclusion):
+    """A COMPLETED run whose conclusion hasn't populated must not read as green.
+
+    Callers gate on status first, so an empty conclusion reaching _is_failing
+    means GitHub called the run complete without saying how it went. Shrugging
+    at that would let the watcher report green on an unobserved outcome.
+    """
+    assert pr_watch._is_failing(conclusion)
     assert not pr_watch._is_passing(conclusion)
+    assert not pr_watch._is_superseded(conclusion)
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


timothyfroehlich:
Claude review of head 877c386 — Lead review (Copilot returned only a quota-limited placeholder postdating head, so this PR would otherwise have merged unreviewed). Verified: three-class classification is fail-safe by construction; all _is_failing callers gate on status==completed first, including watch_run's queued/in_progress guard which precedes classification; _ci_gate_state rank() correctly prefers non-superseded then newest, fixing real check-shadowing at one SHA; superseded grace clock resets on a fresh non-completed gate; announced_superseded is hoisted outside the retry loop so it dedupes across attempts. Scope boundary honest: gh-error -> ('','') -> failure path knowingly left to PP-qkl8. 469 lines of pytest mocked at the gh seam, no network. No findings.

PR #1749, opened 2026-07-26
docs(chores): name all 9 Supabase CLI pin sites in the version-drift check

timothyfroehlich:
## Why

The chores checklist's Supabase CLI pin item said the pin *"lives in CI setup / config"*. That undersells it — the version appears at **9 call sites across 4 workflow files**:

```
.github/workflows/ci.yml               x6  (lines 504, 566, 622, 704, 786, 869)
.github/workflows/preview-control.yaml x1  (line 118)
.github/workflows/preview-sync.yaml    x1  (line 58)
.github/workflows/preview-reaper.yaml  x1  (line 56)
```

So a bump is a multi-site edit, and a **partial** bump leaves jobs running mismatched CLI versions — which surfaces as a job-specific CI failure that reads like a flake rather than a bad edit. Exactly the kind of thing that costs a triage session to untangle.

## What changed

One checklist bullet in `.agents/skills/pinpoint-chores/SKILL.md`:

- Names the 9 sites and the 4 files, with an explicit **"change all nine or none"**.
- Adds a command that lists every pin alongside the newest release, so the comparison is one paste rather than a hunt.
- Documents the `-A6` window. This isn't incidental: `preview-control.yaml` carries an extra `if:` line before `with:`, so a smaller context window silently finds **8 of 9** — I hit that while writing this, and it's precisely the failure the bullet warns about.

## Context

Found while triaging PP-3w32 (GHA infra-flake ledger). The pin itself is **current** — 2.109.1 is the latest `supabase/cli` release, published 2026-07-07 — so this is drift-prevention, not a stale-pin fix.

Related, deliberately kept separate:
- **PP-nlv6** — the weekly check this bullet belongs to (version freshness).
- **PP-zvqt** — caching the pinned CLI *binary* to stop re-downloading it once per job. Different concern.

## Testing

`pnpm run check` green. Docs-only change to a skill file — no code paths touched.

https://claude.ai/code/session_0178hTxVwCeg3CPdPNKGKaaE

### DIFF
--- a/.agents/skills/pinpoint-chores/SKILL.md
+++ b/.agents/skills/pinpoint-chores/SKILL.md
@@ -39,7 +39,16 @@ Then work the checklist. For each item, note findings as a comment on the bead (
 ### Checklist
 
 1. **Stale version-pin checks** (Supabase CLI — PP-nlv6; pnpm corepack pin — PP-w0eq)
-   - **Supabase CLI pin.** Compare the pinned Supabase CLI version against the latest release. The pin lives in CI setup / config; a version-drift nudge belongs here, not in per-session briefing. If stale, file/refresh a bead to bump it (or bump it if trivial and verified).
+   - **Supabase CLI pin.** Compare the pinned Supabase CLI version against the latest release. A version-drift nudge belongs here, not in per-session briefing. If stale, file/refresh a bead to bump it (or bump it if trivial and verified).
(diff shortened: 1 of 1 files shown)


timothyfroehlich:
Claude review of head e0662d2 — Author self-review, explicitly authorized by Tim (docs-only agent-prompt change). Verified both factual claims against the tree: 9 pin sites across 4 files (ci.yml x6, preview-control/sync/reaper x1 each), and the -A6 window requirement is real — preview-control.yaml carries an extra if: line before with:, so -A3 finds only 8 of 9. Documented command executed and output confirmed. Known limitation, not a defect: the command globs .github/workflows/*.y*ml only, so a future setup-cli use under .github/actions/** would be missed; the stated 'all nine or none' count acts as the checksum. No code paths touched.

PR #1750, opened 2026-07-26
fix(issues): metadata selects silently revert after every change (PP-0fvr)

timothyfroehlich:
Fixes PP-0fvr. **Production bug affecting all members** — reported by Phillip this morning.

## Symptom

Setting an issue to "Fixed" silently flips back to "New" about a second later. Same for priority, severity, and frequency.

## Root cause

`@radix-ui/react-select` **2.3.3** added a form-`reset` listener to the Select root:

```js
const initialValueRef = React.useRef(value);
const reset = () => setValue(initialValueRef.current);  // setValue invokes onValueChange
associatedForm.addEventListener("reset", reset);
```

React 19 **automatically resets a `<form action={…}>` once the action settles**, firing that event. Our four metadata forms wired `onValueChange` → `pendingX` state → `useEffect` → `form.requestSubmit()`, so Radix's reset-driven callback was indistinguishable from a real user selection and fired a **second** submit carrying the value the page originally loaded with.

## Evidence

Vercel production logs, White Water WH20-4 — two distinct POSTs, one second apart:

```
13:57:18  POST /m/WH20/i/4   oldStatus:"new"    newStatus:"fixed"   ✅
13:57:19  POST /m/WH20/i/4   oldStatus:"fixed"  newStatus:"new"     ❌
```

- Same pattern 3/3 across two issues (WH20-4, NE-7).
- Timeline events: 30+ clean single status changes before 2026-07-21; **every** change after is a paired close-then-revert.
- No source changes to the page, forms, service, or action in the regression window — dependency bumps only. Radix 2.3.2→2.3.3 landed in #1714 on 2026-07-21.
- A/B test, everything else identical: **2.3.3 → 2 calls** (`fixed`, then `new`); **2.3.2 → 1 call** (`fixed`).

Each failed attempt also sent the reporter **two emails** (resolved + reopened) and wrote two timeline rows.

## Why not just pin the version

The reset listener is still present in **2.3.7**, the current latest — there is no version to move forward to. It's also defensible upstream behaviour on its own (a Select participating in native form reset is reasonable); the mis-integration is on our side. Pinning back to 2.3.2 would strand us on an old release to preserve a bug of our own making.

## The fix

The four forms now build `FormData` by hand and call their `useActionState` dispatch directly inside `startTransition`, instead of routing through `form.requestSubmit()`. No form is ever submitted, so neither React's auto-reset nor Radix's listener can fire.

This also removes two latent problems in the same files:
- the `requestSubmit`-from-`useEffect` dance and its `pendingX` state, and
- a **duplicate `name="status"` field** — our hidden input plus the one Radix renders for form participation — which worked only because `FormData.get` returns the first match.

`assign-issue-form.tsx` was checked and is genuinely unaffected: it has no `<form>` element and uses Popover + cmdk, not `react-select`. Left unchanged.

| Field | Before | After |
|---|---|---|
| status | 2 calls: `fixed`, `new` | 1 call: `fixed` |
| priority | 2 calls: `high`, `low` | 1 call: `high` |
| severity | 2 calls: `major`, `minor` | 1 call: `major` |
| frequency | 2 calls: `frequent`, `intermittent` | 1 call: `frequent` |

## Tests

New table-driven regression test covering all four fields, driving a **real** Radix dropdown selection with an **unmocked** `useActionState`, so React 19's genuine post-action reset mechanics run.

That last part matters: the existing `update-issue-forms-rollback.test.tsx` mocks `useActionState`, which is precisely why this entire bug class was invisible to the suite. That mock now carries a comment saying so. Its rollback-on-error intent is preserved; assertions moved from the removed hidden inputs to the trigger's `aria-label`.

`pnpm run check` green: 1746 unit tests, 167 pytest, typecheck/lint/format/audit clean.

## Progressive enhancement

Adds a narrow, explicitly interim carve-out to **CORE-ARCH-002** in `docs/NON_NEGOTIABLES.md`, with the `AGENTS.md` §2.1 summary kept in sync.

Worth being straight about the tradeoff: these controls are JS-dependent Radix Selects with no submit button and no non-JS fallback, so the rule was only nominally satisfied here — and the native-submit path it preserved is exactly what caused the outage. This removes a real user-visible bug at the cost of a theoretical no-JS regression that was never actually available.

**PP-nw80** tracks Tim's broader revisit of the rule, which may replace this carve-out outright.

## Verification plan

Two production issues were deliberately **left unrepaired** as the post-deploy test case: WH20-4 (White Water, "Feels like its leaning to the right") and NE-7 (Ninja Eclipse, "Ball getting stuck at boss battle drop target"), both still `new`. Setting them to `fixed` after deploy confirms the fix against the exact reported failure.

https://claude.ai/code/session_01EnpetYKuRAbYMf6cZNwFhi


timothyfroehlich:
**Screenshots intentionally skipped** — no visual change in this PR.

The diff deletes two hidden `<input>`s (invisible by definition) and changes how the submit is dispatched (`formAction(formData)` instead of `form.requestSubmit()`). The rendered markup of the four selects is untouched, so screenshots could only confirm that nothing regressed — they cannot show the fix, which is purely behavioural.

Separately, they aren't producible on this host anyway: `pr-screenshots.mjs` has a Firefox fallback for Bazzite (PP-rsy3), but its first run must invoke `playwright test --project=auth-setup`, which is hardcoded to `Desktop Chrome` (`playwright.config.ts:75`) — that's PP-8b6j, where Bazzite's headless Chromium no-ops `fill`/`type`.

Skip confirmed by Tim. —Claude

timothyfroehlich:
## Independent review — APPROVE

Reviewed at head `10612768` by a fresh-context agent that did not author any of this code. Deliberately not the head-commit author: Tim's `/code-review` pass landed on `5e309326`, and the head has moved twice since (review fixes, then the `main` sync), so self-attesting would have made the gate meaningless.

**The mechanism was verified at source level, not taken on trust:**
- Fetched `@radix-ui/react-select@2.3.3` and confirmed the `reset` listener exists as described, and that `setValue` (via `useControllableState`) does invoke the `onValueChange` prop.
- Read the installed `react-dom@19.2.7` bundle and traced `startHostTransition` → `requestFormReset` → `recursivelyResetForms` → `fiber.stateNode.reset()`. That path is reachable **only** through a native form submission (`<form action={fn}>` submit or `.requestSubmit()`). Calling the `useActionState` dispatch directly never enters it — which is exactly what this PR changes.

**Checked and clear:**
- All four hand-built `FormData` key sets match what `actions.ts` reads (highest-risk part of the diff — a typo here fails at runtime with no compile error). No typos.
- Nothing in `src/` or `e2e/` selected the removed hidden inputs; `data-form="update-*"` selectors still match.
- The regression guard's `not.toBeDisabled()` wait is tied to `isPending` clearing, which is the same commit in which the old code would have fired the reset — a principled settle signal, with the 50ms only as a trailing margin. Not racy.
- Rollback test intent preserved; assertions moved 1:1 from hidden inputs to the trigger's `aria-label`.
- `AGENTS.md` §2.1 and `NON_NEGOTIABLES.md` CORE-ARCH-002 agree and match the code.

**Supporting evidence worth noting:** `assign-issue-form.tsx` already used this exact hand-built-`FormData` + `startTransition(() => formAction(...))` pattern. The fix adopts a pre-existing idiom in this codebase rather than inventing a workaround.

**One non-blocking finding:** the four `<form>` elements are now inert (no `action`, no submit control, `onSubmit` prevented), kept only so `data-form` selectors keep matching. Defensible and well-commented, but a plain element would be more semantically honest. Follow-up, not a blocker.

Stated limitation: the reviewer did not empirically re-trigger the old bug (read-only review), so item 4 rests on source tracing plus the passing Unit Tests job rather than a live revert.

—Claude

timothyfroehlich:
Claude review of head 1061276 — no serious findings

PR #1752, opened 2026-07-26
fix(e2e): fail fast when a browser engine cannot render text (PP-8b6j)

timothyfroehlich:
## What was actually wrong

Local smoke E2E could not pass on the Bazzite host **on any branch**. `auth.setup.ts` kept landing on `/login`, which looked like Playwright's `fill()` silently no-opping on headless Chromium — that's how PP-8b6j was originally written up.

`fill()` was never no-opping. **Chromium was crashing.**

A stale entry in `~/.cache/fontconfig` made fontconfig report **zero fonts**. HarfBuzz shaped 0 glyphs and Skia hard-asserted:

```
ERROR:ui/gfx/platform_font_skia.cc:258] Could not find any font: , sans
ERROR:ui/gfx/render_text_harfbuzz.cc:1016] TextRunHarfBuzz error ... font: '', glyph_count: 0
FATAL:third_party/skia/src/ports/SkFontMgr_FontConfigInterface.cpp:163]
```

SIGABRT the instant a spec typed into a field. The same zero-font state is why headless screenshots wouldn't rasterize text (**PP-rsy3**) — one root cause, two symptoms, as that bead suspected.

**It does not self-heal.** ostree pins `/usr/share/fonts` to mtime 0, and fontconfig keys cache validity on directory mtime, so a bad entry is never invalidated. `fc-list` masks it by rescanning — which is why the host looked perfectly healthy (787 fonts, `fc-match sans` → Noto Sans) the whole time.

The host fix is one line: `rm -f ~/.cache/fontconfig/*.cache-* && fc-cache -f`. **Already applied** to the Bazzite host.

## What this PR changes

The host fix needs no code. This adds the **guard**, so the next occurrence costs minutes instead of days.

`global-setup.ts` already verifies browser binaries *exist*. Both real failures on this host live in the gap between "exists" and "works". So: probe the engines this config needs and confirm they can shape a run of text.

- **Chromium is fatal** — `auth-setup` runs there, so nothing downstream can pass without it.
- **Other engines warn**, so a broken engine doesn't block the ones that work.
- Remedies are specific: a missing binary gets `playwright install`, a missing shared library is named, and only a genuine render failure points at fontconfig.

### Scope: `needed ∩ installed`

**Needed**, so CI's chromium-only runs don't demand firefox/webkit, and so a config with no projects probes nothing — that empty-projects seam is how `src/test/unit/global-setup.test.ts` disables browser checks. **Installed**, so a missing binary stays `checkBrowserBinaries`' story and is never misreported here as a font fault.

Worth knowing for anyone editing this file: `FullConfig.projects` is **not** filtered by `--project` (a `--project=chromium` run still lists all 5), and device-based projects leave `use.browserName` undefined — the engine rides on `defaultBrowserType`. So `needed` resolves to chromium in practice, which is the engine that matters here and the one whose failure was silent. It also means the pre-existing `checkBrowserBinaries` only ever effectively checks chromium; I left that alone, since "fixing" it would make CI's chromium-only jobs demand firefox+webkit and go red.

## Review note: CI caught two real bugs in the first cut

Worth calling out rather than burying, since both are the exact failure mode this guard exists to prevent:

1. The probe originally keyed off **installed binaries only**, ignoring `config.projects`. That broke the unit-test seam, so the unit job tried to launch real browsers on a runner with none — 11 tests failed.
2. It then **misdiagnosed** the result, reporting a missing binary as "rendered no glyphs — the host font stack is broken" and pointing at fontconfig when the real fix was `playwright install`.

Both are fixed, and the second is now locked down by a test asserting the font remedy must *not* appear for a launch failure.

## Verification (on the affected host)

| Scenario | Result |
|---|---|
| Poisoned cache restored | Guard throws immediately, naming the font remedy |
| Printed remedy run **verbatim** | Recovers; 4 passed |
| Healthy host | Silent, no false positive |
| `pnpm run check` | exit 0 (1759 tests, +5 new) |
| `pnpm run smoke` | **159–160 passed** — previously could not start at all |

New unit tests: no-glyphs throws the font remedy; a launch failure names the missing library and must not mention `fc-cache`; a healthy engine proceeds; an empty config launches nothing; an absent binary is not probed.

## Follow-ups (not in scope here)

- **PP-5f22** — WebKit is unusable on this host: `libicudata.so.74` isn't in Fedora 43, so all local `Mobile Safari` specs die at launch. That accounts for the remaining smoke failures; unrelated to PP-8b6j.
- **PP-v3ub** — `e2e/smoke/collections.spec.ts` on Mobile Chrome is flaky locally (1 fail / 2 passes on identical code): the local 5s `actionTimeout` against a cold Next.js dev route. Pre-existing.

https://claude.ai/code/session_019MofLXLs6czkvBRNtQMUky


timothyfroehlich:
Claude review of head 3ce1b42 — no serious findings

PR #1753, opened 2026-07-26
chore(chores): add weekly prod backup validation (PP-qehv)

timothyfroehlich:
## Why

Our disaster-recovery posture — *"Supabase Pro takes daily backups with 7-day retention"* (`AGENTS.md` §7, closed bead `PinPoint-dka`) — has never been verified by anything in the repo. A silently-failing backup job, a retention change, or a plan downgrade would only have surfaced during a restore.

Tim asked for a weekly chore that validates the backups are still valid.

## What

`pnpm run chores:backups` → `scripts/check-prod-backups.mjs`, added as **item 9** on the weekly chores checklist.

It calls `supabase backups list --project-ref <prod> --output json` and asserts:

| Level | Condition |
| :--- | :--- |
| FAIL | newest `COMPLETED` backup older than 48h |
| FAIL | fewer than 7 `COMPLETED` backups retained |
| FAIL | `walg_enabled !== true` |
| FAIL | zero backups / no `COMPLETED` / unparseable timestamps / CLI error |
| WARN | newest `COMPLETED` backup 24–48h old |
| WARN | any non-`COMPLETED` entry |
| WARN | >36h gap inside the retention window (a hole later backups would mask) |
| INFO | `pitr_enabled` reported verbatim, so a posture change is visible rather than silent |

Exit 1 on FAIL only; warnings print loudly under a PASS.

## Scope: attestation, not a restore drill

This proves backups **exist and are retained**. It does **not** prove they restore — that means restoring a physical backup into a throwaway project, which isn't a weekly-cadence activity. The skill and the bead both say so explicitly so a green run isn't misread as "DR is verified".

## Implementation notes

- **`--output json` is load-bearing.** The Supabase CLI emits JSON under agent auto-detection but an ANSI-decorated table on a human TTY. Verified the explicit flag wins even under `--agent no`.
- **Node, not Python.** The script must shell out to `supabase`, and the Claude Code sandbox blocks Python `subprocess`. Auth comes from the CLI's keychain token — there's no `SUPABASE_ACCESS_TOKEN` in the environment.
- **Stays out of `check` / `preflight` / CI** — it hits the network and the prod Management API.
- Appended as item 9 rather than grouped with the Supabase advisor item, because `docs/runbooks/gha-flake-log.md:244` references "chores checklist item #8" and renumbering would break it.
- Assertion logic lives in an exported pure `evaluate(payload, now)` so branches are verifiable without network access.

## Two things found along the way

1. **`pnpm run db:backup` is not a DR artifact.** It's `supabase db dump --data-only --schema public` — no schema, no `auth.users`, no storage. It's a dev-seeding tool. `AGENTS.md` now says so.
2. **Prod has `pitr_enabled: false`**, so the recovery floor is the previous nightly snapshot — up to ~24h of member-entered data loss in an incident. Recorded in `AGENTS.md` §7. That's a posture decision, not a bug, and deliberately not changed here.

## Verification

- `pnpm run chores:backups` against prod → **PASS**, 8 COMPLETED backups, newest 11h old, `walg_enabled: true`.
- All 12 assertion branches exercised via `evaluate()` with synthetic payloads (stale-72h, 30h-warn, only-4-retained, one-FAILED, all-FAILED, walg-off, 40h-gap, zero-backups, bad-timestamp, pitr-on, null payload) — each produced the expected PASS/FAIL and message.
- Bogus `--project-ref` → clean `FAIL` with the CLI's 404, exit 1, no stack trace.
- `--output json` confirmed honored under `--agent no` (the human-TTY path).
- `pnpm run check` green.

Docs kept in sync: skill checklist, `PP-qehv` bead description (the skill requires parity), and `AGENTS.md` commands table + §7.


timothyfroehlich:
Claude review of head ba627ea — Copilot quota-limited (empty review, PP-jw0s false-green). Manual review of the full diff: found and fixed one real defect — the hand-built `file://` direct-invoke guard silently no-ops on URL-encoded paths, making the verification script exit 0 having checked nothing (fixed in ba627ea5, verified against a path with a space). All 12 assertion branches exercised with synthetic payloads. No other findings. —Claude

PR #1754, opened 2026-07-26
fix(config): gate prod-mutating Supabase MCP tools and CLI verbs (CORE-SEC-010)

timothyfroehlich:
## Problem

The claude.ai Supabase MCP connector is wired to **`pinpoint-prod` (`udhesuizjsgxfeotqybn`) — the only project in the org.** Live member data, `pitr_enabled: false`, so the recovery floor is the previous nightly backup (~24h of data loss). Verified by calling `list_projects`, which returned prod **with no permission prompt.**

Nothing guarded it:

- `.claude/settings.json` `permissions.deny` had exactly one DB entry — `Bash(supabase db push:*)`. Several other destructive Supabase CLI verbs had no rule at all.
- The `PreToolUse` hook stack matches `Bash` and `Agent` (plus `mcp__github__merge_pull_request` on the merge guard). **No hook sees Supabase MCP calls.**
- MCP calls bypass Bash entirely, so the script-level guard pattern this repo prefers (`assertNotDrizzlePush()`, `scripts/assert-local-db.mjs`) cannot reach them either.

A permission rule is the only available lever.

### MCP — all `ask`

`execute_sql`, `apply_migration`, `deploy_edge_function`, `pause_project`, `restore_project`, `create_project`, and the five branch mutators (`create` / `delete` / `merge` / `rebase` / `reset`).

Read-only tools (`list_*`, `get_*`, `search_docs`, `generate_typescript_types`) are deliberately **not** gated — prompting on those trains the reflex to click through and devalues the prompt.

Server name prefix `mcp__claude_ai_Supabase__` confirmed against the live tool surface in-session; no other spelling appears anywhere in the repo. Rules are bare tool names — MCP rules take no argument pattern, so no `Bash(...)` wrapper and no `:*` suffix.

### CLI — deny vs ask, per verb

| Verb | Tier | Reasoning |
| :--- | :--- | :--- |
| `supabase db push` | `deny` | pre-existing |
| `supabase projects delete` | **`deny`** | No legitimate use — prod is the only project, and a deleted project is not recoverable from a nightly backup |
| `supabase db reset` | `ask` | See below |
| `supabase db remote commit` | `ask` | Recoverable; legitimate in some workflows |
| `supabase migration repair` | `ask` | Rewrites migration history state, but recoverable |
| `supabase branches delete` | `ask` | Has a real call site — `scripts/workflow/preview/preview-destroy.sh:44` |

`deny` is reserved for verbs with **no** legitimate use here. Everything else is `ask`, per Tim's stated preference.

**Why `db reset` is `ask`, not `deny`:** the local form is routine, and a prefix rule cannot reliably distinguish it from `--linked` (flag order varies, so `Bash(supabase db reset --linked:*)` would miss `supabase db reset --yes --linked`). A single `ask` covers both forms, and the prompt shows the full command — so the human sees `--linked` before approving. That is a better failure mode than a hard block on a routine local operation.

## False-positive verification

Both confirmed directly, not assumed:

1. **`pnpm run db:reset` is unaffected.** It is `db:_restart` (`supabase stop && supabase start`) → drop tables → `db:migrate` → seed chain. It never calls `supabase db reset`.
2. **E2E global-setup is unaffected.** `e2e/global-setup.ts:257` does call `execSync("supabase db reset --yes")` — against the **local** stack, as a fallback when `db:fast-reset` fails. But it runs inside the Playwright global-setup process, not as an agent `Bash` tool call, so no permission rule is consulted. Permission rules gate the agent's own command string; they do not reach child processes a program spawns.

Same reasoning covers `preview-destroy.sh`: an agent running `bash scripts/workflow/preview/preview-destroy.sh` is not matched by `Bash(supabase branches delete:*)`, because the rule matches the typed command, not what the script does internally.

## What these rules are *not*

**A speed bump against accidents, not a security boundary.** Claude Code permission rules are prefix matchers over the command string, so a wrapper shape (`sh -c "..."`, `eval`, env indirection) walks straight past them — exactly as it does past this repo's hook stack (see the PR #1736/#1738 history). And `ask` is a prompt, not a block. Please don't let anyone later mistake these for airtight.

Per that same history, **no `PreToolUse` hook was added** to chase the wrapper gap. This repo tried that approach twice and concluded effect-rules belong at the lowest layer instead.

## Docs

Recorded as **CORE-SEC-010** in `docs/NON_NEGOTIABLES.md` §Security, with the scope caveat above stated explicitly in the rule body. Cited from `AGENTS.md` §7 Deployment → Supabase, where the existing prod-safety rules live (`Prod: NEVER db:reset. Only db:migrate.`) — rather than §2.1, which indexes implementation rules; this is an environment/harness rule.

`pnpm run check:rule-ids` passes (it is already part of `pnpm run check`).

## Checks

- `pnpm run check` — exit 0
- `node -e "JSON.parse(...)"` on `.claude/settings.json` — parses
- `pnpm exec vitest run src/test/unit/hooks/verify-guard-stack.test.ts` — 14/14 pass (the canary asserts `permissions.ask` is a non-empty array)

Not UI-touching, so no screenshots.

## One correction to the framing

The premise that the repo is linked to prod via `supabase/.temp/project-ref` is **not currently true** — `supabase/.temp/` does not exist in either the main checkout or this worktree, so `supabase db reset --linked` would fail on a missing link rather than reset prod today. The rule is still correct as a guard (a `supabase link` can happen at any time, and `--db-url` reaches a remote without any link), but the "resets PRODUCTION right now" framing overstates the present state.

https://claude.ai/code/session_01NFg3c81oW1jvLZvgV6K1Nh


timothyfroehlich:
Claude review of head 26a2091 — Read the full diff — permission rules are correctly formed (MCP rules as bare tool names, no Bash() wrapper or :* suffix; CLI rules as prefix globs). Verified pnpm run db:reset does not call 'supabase db reset' so the new ask rule won't fire on it, and confirmed permission rules don't reach child processes, so e2e/global-setup.ts:257 is unaffected. Docs are honest that these are prefix matchers evadable by wrapper shapes — a speed bump, not a boundary. Copilot's review was quota-limited and inspected nothing.

PR #1758, opened 2026-07-27
fix(scripts): close three prod-safety gaps in the DB scripts

timothyfroehlich:
Every script touched here resolved its target purely from an env var, so "which database" was whatever happened to be in the shell. `PinPoint-Prod` has `pitr_enabled: false` — the recovery floor is the previous nightly backup.

No new mechanism: this follows `scripts/assert-local-db.mjs` (hard loopback allowlist) and `scripts/lib/drizzle-push-guard.ts` (guard the *effect* at the lowest layer that can express it, so no shell wrapper routes around it). No PreToolUse hook — that approach was tried and rejected twice (#1738 → #1741).

## New shared module: `scripts/lib/db-target.mjs`

Two deliberately different questions, because conflating them is what would break things:

| predicate | question | true for a preview branch? | escape hatch |
| :-- | :-- | :-- | :-- |
| `isCloudDatabaseUrl()` | any managed cloud Postgres? | **yes** (same `*.pooler.supabase.com` host shape) | required — it's a human-typed opt-in |
| `isPinPointProductionTarget()` | *the* production project? | **no** (branches have their own refs) | none needed |

`isCloudDatabaseUrl` is the regex lifted verbatim out of `drizzle.config.ts`, which now imports it instead of keeping a second copy. `DRIZZLE_FORCE_PRODUCTION` behaviour is byte-for-byte preserved (verified by running `drizzle-kit check` against a prod-shaped URL — same refusal, same wording).

## Gap A — five seed scripts had no host guard

Split by each script's *actual sanctioned reach*, not by a blanket rule:

**Local-only demo seeds → `assertLocalDatabase` (loopback allowlist):**
- `seed-collections.mjs`
- `seed-machine-settings.mjs`

Neither is wired into the preview pipeline, so the hard allowlist costs nothing. The stale "no host guard, DO NOT point at prod" comments in both headers are replaced. `preview-migrate-seed.sh`'s "PR-specific demo seeds go below this line" comment now warns that these two would exit 2 on a branch DB.

**Remote-capable seeds → `assertNotPinPointProduction` (prod refused, no override):**
- `seed-users.mjs`
- `seed-discord.mjs`
- `seed-pinballmap-catalog.mjs`

⚠️ **This corrects the brief.** `seed-users.mjs` and `seed-pinballmap-catalog.mjs` are remote-capable *on purpose*: `scripts/workflow/preview/preview-migrate-seed.sh:83,97` runs them against an ephemeral Supabase preview branch. `assertLocalDatabase` on `seed-users.mjs` would break `/preview` outright. The 2026-06-18 pooler-cleanup plan made that call explicitly ("Seed scripts are remote-capable — do NOT add a local-only guard to `seed-users.mjs` / `seed-discord.mjs` … This **overrides** PP-z428's wording"). That decision stands; the prod hole it left open is what's closed here.

**`seed-users.mjs` guards two endpoints, not one.** It creates real auth users via `supabase.auth.admin.createUser` against `NEXT_PUBLIC_SUPABASE_URL` — a *different* variable from `POSTGRES_URL`, either of which could point at prod on its own. And the coarse host regex would not have caught it anyway: `https://<ref>.supabase.co` does not contain the substring `supabase.com`. The project-ref predicate does. This case is the main reason the module carries two predicates rather than one.

**`seed-timeline-backfill.mjs` needed no change.** Contrary to the brief, it already calls its own guard (`ALLOW_NONLOCAL_BACKFILL`, lines 46–56) — deliberately overridable, because it has a documented intentional prod one-shot.

**Found while working: `scripts/reset-preview-db.mjs`.** It `DROP SCHEMA public CASCADE` + `DELETE FROM auth.users`, and its existing prod check is conditional on the caller supplying `PROD_PROJECT_REF` — which the preview workflows do and a hand-run `node scripts/reset-preview-db.mjs` does not. Now guarded unconditionally, with the caller-declared check kept as layer 2.

## Gap B — `mark-migration-applied.ts` had zero prod check

Stays prod-capable, because it *is* the documented stuck-migration recovery path. Gate shape follows the `DRIZZLE_FORCE_PRODUCTION` precedent so there's one idiom:

1. `MARK_MIGRATION_FORCE_PRODUCTION=1` required against a remote target. The refusal prints the redacted target host and the migration number, and spells out the exact command to re-run — an operator mid-incident must not hit a wall.
2. A y/N confirm showing target + resolved migration tag, gated on `process.stdin.isTTY`, so a non-TTY caller never reaches the prompt and nothing can hang. Manually verified both ways (prompt renders and `n` aborts without writing; non-TTY sails through on the token alone).

Verified `mark-migration-applied.ts` is **not** on the Vercel build path — `vercel-build` is `migrate:production && next build`.

Token propagated to `scripts/migrate-production.ts:151`'s printed recovery instructions, `AGENTS.md` §7, and `docs/DEVELOPMENT.md`.

## Gap C — stale `supabase link` state: documented, not automated

**Decision: print a precise post-backup warning; do not unlink, do not re-plumb blind.**

`backup-production.sh` depends on the link twice over — its own preflight check reads `supabase/.temp/project-ref`, and the bare `supabase db dump` resolves both the project and its stored DB password from that state. Auto-unlinking would break the next `pnpm run db:backup`, which the brief ruled out.

**Correction to the stated footgun.** "A later bare `supabase db reset` hits production" is **false** on Supabase CLI 2.109.1 — bare `db reset` is local-only; remote requires an explicit `--linked` or `--db-url` (`supabase db reset --help`). The commands that genuinely default to the linked project are `db dump`, `db push`, `db pull`, `migration up --linked`, `db reset --linked`, `secrets` — and of those `db push`/`db pull` aren't live paths here (Supabase migration config is disabled, `supabase/migrations/` is gitignored). Real but narrow, so the warning names exactly that set and explicitly says bare `db reset` is unaffected.

The proper fix — `--db-url` / `--project-ref` + `SUPABASE_DB_PASSWORD` so the script never depends on ambient state, after which unlinking becomes safe — needs a real prod dump to validate and a prod credential sourced outside the CLI credential store. Filed as **PP-oifw** to pick up alongside the next hands-on prod backup, where a failed attempt is cheap to notice.

## Tests

`src/test/unit/scripts/db-target-guards.test.ts` (33 cases) — pure-predicate tests plus real `spawnSync` runs of every guarded script asserting exit codes, matching the `src/test/unit/hooks/*.test.ts` pattern. A guard that's exported but never actually *called* is the exact regression these catch. Coverage includes:

- prod URL refused (exit 2) / localhost passes through to the connection attempt, for each seed
- **preview-branch URL still passes** the remote-capable seeds — the check that keeps `/preview` working
- `seed-users.mjs` refused via `NEXT_PUBLIC_SUPABASE_URL` alone *and* via `POSTGRES_URL` alone
- Gap B: token required, token permits, no hang without a TTY, and the refusal never prints the password
- a wiring assertion that `drizzle.config.ts` imports the helper rather than re-inlining the regex

## Verification

`pnpm run preflight` — unit 1787 ✅, build ✅, integration 508 ✅, integration-supabase 104 ✅, smoke 160 ✅. The 28 remaining smoke failures are all `[Mobile Safari]` and all the same host error: `MiniBrowser: error while loading shared libraries: libicudata.so.74`. That's a missing system library on this Bazzite host, not a code failure — CI's runner has it.

Also verified by hand: `drizzle-kit check` still passes and still refuses `push`; the prod refusal in `drizzle.config.ts` is unchanged; `pnpm run db:fast-reset` runs the whole seed chain clean through the new guards.

Closes nothing on its own; PP-oifw tracks the Gap C follow-up.

https://claude.ai/code/session_01NFg3c81oW1jvLZvgV6K1Nh

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -214,7 +214,7 @@ How Tim wants agents to behave. (§1 has the one-line version; this is the detai
 ### Vercel
 
 - Vercel runs `pnpm run migrate:production` on build (production only).
-- Stuck migration fix: `POSTGRES_URL=<prod_url> tsx scripts/mark-migration-applied.ts <n>`.
+- Stuck migration fix: `MARK_MIGRATION_FORCE_PRODUCTION=1 POSTGRES_URL=<prod_url> tsx scripts/mark-migration-applied.ts <n>`. The token is required — the script refuses a remote target without it, and prompts once more when run in a TTY. It writes to `drizzle.__drizzle_migrations` without running the migration, so a wrong number makes prod's schema diverge from history permanently.
 
 ### Preview deployments (on-demand, TTL'd Supabase branches)
 
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -161,11 +161,19 @@ For preview and production, we use **Automated Migrations** via Vercel build hoo
 
    ```bash
    # 1. Identify the failed migration number from Vercel build logs
-   # 2. Mark it as applied (replace 0001 with actual migration number):
-   POSTGRES_URL=<production-url> tsx scripts/mark-migration-applied.ts 0001
+   # 2. Mark it as applied (replace 0001 with actual migration number).
+   #    MARK_MIGRATION_FORCE_PRODUCTION=1 is required against a remote database:
+   #    the script refuses without it, and prompts once more when run in a TTY.
+   MARK_MIGRATION_FORCE_PRODUCTION=1 POSTGRES_URL=<production-url> \
+     tsx scripts/mark-migration-applied.ts 0001
 
    # 3. Trigger a redeployment (push a trivial change or use Vercel UI)
    ```
+
+   This records the migration as applied **without running its SQL**. Marking the
+   wrong number makes every future `migrate:production` skip a migration that
+   never ran, so prod's schema diverges from the migration history permanently —
+   double-check the number before you set the token.
 
    **Prevention:** Always use the automated migration system. Avoid manually running migrations in production/preview unless absolutely necessary.
 
--- a/drizzle.config.ts
+++ b/drizzle.config.ts
@@ -1,6 +1,7 @@
 import { loadEnvConfig } from "@next/env";
 import { defineConfig } from "drizzle-kit";
 
+import { isCloudDatabaseUrl } from "./scripts/lib/db-target.mjs";
 import { assertNotDrizzlePush } from "./scripts/lib/drizzle-push-guard";
 
 // Safety: `drizzle-kit push` is banned (CORE-ARCH-009). This runs FIRST — before
@@ -50,11 +51,10 @@ if (!directUrl) {
   );
 }
 
-// Safety: prevent drizzle-kit from accidentally running against production
-const isProductionUrl = /supabase\.com|neon\.tech|rds\.amazonaws\.com/.test(
-  directUrl
-);
-if (isProductionUrl && !process.env.DRIZZLE_FORCE_PRODUCTION) {
+// Safety: prevent drizzle-kit from accidentally running against production.
+// The host match lives in scripts/lib/db-target.mjs so drizzle-kit and
+// scripts/mark-migration-applied.ts share one copy of it.
+if (isCloudDatabaseUrl(directUrl) && !process.env.DRIZZLE_FORCE_PRODUCTION) {
   throw new Error(
     `SAFETY: drizzle-kit would run against a production database!\n` +
       `   URL: ${directUrl.replace(/:[^:@]+@/, ":***@")}\n` +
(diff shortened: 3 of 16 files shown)


timothyfroehlich:
Claude review of head 5812f9e — Read the full diff. The two-predicate design in scripts/lib/db-target.mjs is correct and load-bearing: isPinPointProductionTarget() matches the prod ref, which the coarse cloud-host regex provably misses for the Supabase API URL (https://<ref>.supabase.co contains no 'supabase.com') — and seed-users.mjs writes through BOTH that API URL and POSTGRES_URL, so a single-channel guard would have left a hole. Verified the test suite covers the subtle case (prod NEXT_PUBLIC_SUPABASE_URL with a local POSTGRES_URL is refused) and that preview-branch URLs still pass, which is required because preview-migrate-seed.sh runs these seeds against ephemeral branches. Guards are placed before any client construction. MARK_MIGRATION_FORCE_PRODUCTION uses the same truthiness idiom as the existing DRIZZLE_FORCE_PRODUCTION, so '=0' would enable it — consistent with precedent, not a regression. Copilot's review was quota-limited and inspected nothing.

PR #1762, opened 2026-07-27
feat(machines): convert the Edit modal into a machine Manage tab (PP-o355.19)

timothyfroehlich:
Converts the machine **Edit** modal into a full page at `/m/[initials]/edit`, surfaced as a **Manage** tab (PP-o355.19), part of the PinballMap epic PP-o355.

## What changed

- **New Manage tab** at `src/app/(app)/m/[initials]/(tabs)/edit/page.tsx` — a Server Component that resolves the machine, `notFound()`s when it can't, and redirects to `/m/[initials]` when `checkPermission("machines.edit", …)` fails. Three sibling sections: Details, Pinball Map, Danger zone. It lives in the `(tabs)` group, so the machine header, tab strip, and translite come from the shared layout and the file renders only the panel.
- **The tab is the entry point; there is no Edit button.** The layout computes `machines.edit` and only then renders the tab, so visibility is the gate — the page's own check remains as the deep-link guard. The tab is appended **last** rather than slotted beside Settings: it is permission-gated, so appending keeps every other tab at the same index for every role and the strip doesn't reflow depending on who is looking.
- **URL slug stays `edit`; the visible label is "Manage".** The tab holds the machine's *record* — name, model, availability, PBM listing, ownership. Settings holds the machine's *DIP switches, software settings and Jones plugs*. Calling this one "Edit" invited "do I change availability in Edit or Settings?".
- **Details** and **owner transfer** are now separate forms with separate saves. Ownership moved out of the details save into the Danger zone, so a details save no longer touches the owner columns.
- **Owner picker uses an inline disclosure, not a nested dialog** — PP-o355.13 established that a Radix popover portalled to `<body>` reads as an outside-click and dismisses the surrounding dialog.
- **Deleted** `update-machine-form.tsx` (593 lines) and its test, plus `edit-button-tooltip.tsx`. The latter existed only to explain to a non-permitted viewer why the Edit button was disabled — with no button there is nothing to disable, and a hidden tab says nothing, so **that explanation is gone.** Flagged rather than silently dropped; worth revisiting if members ask why they can't edit.

## React 19 post-action form reset (PP-1ajq)

React 19 fires a `reset` on a `<form action={...}>` once the action settles — **on failure as well as success**. Both forms on this page stay put on failure, so the reset landed on a form the user was still looking at:

- **Details** — the uncontrolled Machine Name input snapped back to its `defaultValue`, and `@radix-ui/react-select` ≥2.3.3 replayed Availability's mount-time value through its own form-`reset` listener. Silent data loss under an error banner.
- **Owner transfer** — a native reset blanks the *controlled* hidden `id`/`name` inputs in the DOM without re-rendering React (state is unchanged, so React has no reason to re-sync), so a second attempt after a failed transfer would have submitted an empty machine id. Latent; found while porting the fix.

Both now drop `action={formAction}` and dispatch `useActionState` directly from a FormData snapshot inside `startTransition`. Native constraint validation still runs — a submit event only fires after it passes. The owner form also picks up the descendant-form guard: React propagates events through the *React* tree, so the portalled `<form>` inside `OwnerSelect`'s `InviteUserDialog` reaches this handler.

## Layout: the tab is compact enough to fit above the fold

The tab wasted horizontal space — single-column full-width fields in a narrow panel, with Danger zone below the fold at 1440×900. The panel is now an `@container` capped at `max-w-4xl`; Machine Name pairs with Availability and Model with Edition at `@xl`; vertical rhythm is tighter. The whole tab now fits above the fold.

**The two pairings are not equivalent, and the code comments say so.** Model/Edition is semantic — Edition is meaningless without a Model. Name/Availability is purely a layout convenience, taken knowingly: multi-column forms cost scanning speed when the columns are unrelated, and it moves the keyboard path to Name → Availability → Model. **A new test pins that order** so a later layout tweak can't undo the decision silently (verified red-then-green).

A follow-up bead, **PP-qagr**, proposes lifting a 1280px content cap into `MainLayout` site-wide so pages stop depending on each one remembering `PageContainer`.

## Pinball Map listing control is stubbed, deliberately

**PP-o355.21 replaces the #1683 control wholesale** — states *derived* from the last sync snapshot instead of discovered by pressing "Connect", with "Connect"/"Verify" retired. Shipping the old control on a new tab would teach an idiom we're about to remove, and its "Not listed" badge is wrong for machines that are in fact on the map.

So the box holds its place and says "Listing controls are coming soon" rather than saying something it would have to walk back. **"Sync now" above it is real and still works.** This does mean the Connect/Verify writes are unavailable until .21 lands — a conscious trade, since .20 (auto-link) makes linking automatic anyway.

`PinballmapListingControl` and its tests are left in the tree; .21 rewrites that file, so deleting now means recreating.

## Accessibility fix found by this PR

`text-destructive` (#dc2626) renders at **3.96:1** against the page background — under WCAG AA's 4.5:1 for normal text. That's error banners, required-field markers and destructive buttons.

It was missed because the token's comment reads "red-600 … passes (~5.3:1)" — but that 5.3:1 is *white-on-red for filled controls*. The inverse case was never checked. Compacting this tab pulled the "Change owner…" button into the axe viewport in `responsive-overflow.spec.ts`, where it could finally be evaluated.

Adds `--color-destructive-text: #ef4444` (red-500, 5.09:1), chosen over the lighter red-400 from a rendered playground — 400 washes toward pink on dark and reads less alarming. Fills keep #dc2626. **This PR converts the Manage tab only**; 78 further occurrences across 42 files (including 3 shared `src/components/ui` primitives) are swept under **PP-mjms** so they get their own visual review.

## Verification

- `pnpm run check` green — 215 files / 1821 unit tests, plus 217 pytest.
- The failed-save guard leaves `useActionState` unmocked so React 19's real post-action mechanics run, **verified red-then-green**: restoring `action={formAction}` fails it. A second test asserts the FormData the action actually receives.
- The new keyboard-order test also verified red-then-green.
- `responsive-overflow` 22/22 including the a11y pass on `/m/EBD/edit`; `invite-signup` 5/5; `technician-role` + `machine-info` 11/11.
- Screenshots posted below at head SHA.

## Known follow-ups

- `PP-o355.21` — rebuild the listing control (derive, don't discover).
- `PP-mjms` — sweep the remaining 78 `text-destructive` usages.
- `PP-qagr` — site-wide content-width cap.
- `PP-o355.26`, `PP-o355.27` — deferred minors from the earlier review.
- `PP-nlnx` — Mobile Safari overflow flake; evidence appended showing the new route behaves identically to untouched sibling routes.


timothyfroehlich:
## UI Screenshots — head `b82a2197`

### Issues List

| --- | --- |
| desktop | mobile |

| --- | --- |
| desktop | mobile |

### Report an Issue

| --- | --- |
| desktop | mobile |

### Dashboard

| --- | --- |
| desktop | mobile |

| --- | --- |
| desktop | mobile |

### Collections

| --- | --- |
| desktop | mobile |

| --- | --- |
| desktop | mobile |

### New Machine

| --- | --- |
| desktop | mobile |

| --- | --- |
| desktop | mobile |

PR #1767, opened 2026-07-28
chore(config): make destructive-command permission rules machine-portable (PP-ah8o)

timothyfroehlich:
## What

Copies the destructive-command permission rules out of the Mac's machine-local `~/.claude/settings.json` and into the checked-in project settings, so they apply on Bazzite too.

**`deny` gains** — the three commands AGENTS.md §4 Process safety already forbade in prose but nothing enforced, plus host- and repo-destructive verbs:

`pkill` · `killall` · `supabase stop --all` (both spellings) · `shutdown` · `reboot` · `mkfs` · `gh issue delete` · `gh repo delete`

**`ask` gains** — the reversible-but-costly set:

`rm -rf` · `docker rm` · `docker volume rm` · `kill` · `xargs kill` · and the git history-rewriting verbs: `rebase`, `reset --hard`, `clean -f`/`-fd`, `branch -D`, every `push --force` spelling

Net: `deny` 2 → 11, `ask` 17 → 40.

## Why the project file and not user settings

`~/.claude/settings.json` is deliberately not stowed into the dotfiles repo — it carries absolute `/Users/froeht` paths (confined to the `permissions` and `statusLine` keys), a macOS `sandbox` block, and a Mac-only `skillOverrides` entry. Bazzite's home is `/var/home/froeht`, so those paths would be wrong there.

Claude Code offers no way to split it. There is **no user-scope `settings.local.json`** (project scope only), no `extends`/`include`, and no `$HOME`/`~` interpolation in settings values — so the `.zshrc` → `.zshrc.local` pattern the dotfiles use everywhere else has no equivalent at `~/.claude`.

The portable channel is the permission arrays themselves. `permissions.allow`/`deny`/`ask` are the documented **merge exception**: they concatenate and dedupe across every settings scope rather than highest-scope-wins. So a rule checked in here reaches both machines through `git pull`, and duplicating a user-scope rule costs nothing at runtime.

This route already has a precedent in-repo — the Supabase MCP guard rules (CORE-SEC-010) landed in this same file and took effect on the Mac with no local change.

## Copied, not moved

The Mac's user settings keep their copies. Moving the rules would have left `rm -rf`, force-push, and friends unprompted in every repo that isn't PinPoint — a silent widening of posture well outside this change's scope.

## Scope caveat

Everything CORE-SEC-010 says applies here: these are prefix matchers over the command string, so a wrapper shape (`sh -c "…"`, `eval`, env indirection) walks straight past them. `ask` is a prompt, not a block. **A speed bump against accidents, not a security boundary.** They also don't reach child processes, so `pnpm run db:reset` and the Playwright global-setup are unaffected.

## Verification

- `pnpm run check` green — 1807 unit tests, 217 pytest, types + lint + format clean
- `verify-guard-stack.cjs` canary still passes (it asserts the deny/ask arrays are non-empty; both grew)
- JSON validated; `deny` 11 entries, `ask` 40

Closes PP-ah8o.


.claude/settings.json
```diff
@@ -21,7 +30,30 @@
...
+      "Bash(docker volume rm:*)",
+      "Bash(kill:*)",
+      "Bash(xargs kill:*)",
+      "Bash(git rebase *)",
+      "Bash(git -C * rebase *)",
+      "Bash(git reset --hard)",
+      "Bash(git reset --hard *)",
+      "Bash(git -C * reset --hard)",
+      "Bash(git -C * reset --hard *)",
+      "Bash(git clean -f *)",
```
timothyfroehlich: Trailing-space patterns require at least one argument, so bare/other-flag forms slip past this gate: `git clean -fdx` and `git clean -xdf` match neither `git clean -f *` nor `git clean -fd *` and run unprompted. Consider `Bash(git clean:*)`.

.claude/settings.json
```diff
@@ -21,7 +30,30 @@
...
+      "Bash(git clean -fd *)",
+      "Bash(git branch -D *)",
+      "Bash(git -C * branch -D *)",
+      "Bash(git push --force)",
+      "Bash(git push --force *)",
+      "Bash(git push --force-with-lease *)",
```
timothyfroehlich, line 52: `git push --force-with-lease` with no trailing argument (the common form on the current branch's upstream) does not match this pattern and runs without a prompt; `Bash(git push --force-with-lease:*)` would cover both.

.claude/settings.json
```diff
@@ -21,7 +30,30 @@
...
       "mcp__claude_ai_Supabase__merge_branch",
       "mcp__claude_ai_Supabase__rebase_branch",
-      "mcp__claude_ai_Supabase__reset_branch"
+      "mcp__claude_ai_Supabase__reset_branch",
+      "Bash(rm -rf:*)",
+      "Bash(docker rm:*)",
+      "Bash(docker volume rm:*)",
+      "Bash(kill:*)",
+      "Bash(xargs kill:*)",
+      "Bash(git rebase *)",
```
timothyfroehlich, line 40: `git rebase *` misses the bare `git rebase` (and `git rebase` continuation forms are matched, which may over-prompt); the `:*` prefix form used elsewhere in this same block (e.g. `Bash(rm -rf:*)`) matches both bare and argument forms consistently.

.claude/settings.json
timothyfroehlich: Fixed in d084c14a. Confirmed the mechanism — a trailing ` *` requires the prefix to be followed by a space **or end-of-string**, so `git clean -f` followed by `dx` genuinely misses. Replaced both rules with `git clean *` (flag-order agnostic), plus `git -C * clean *`, and added `rm -fr *` for the same class. —Claude

.claude/settings.json
timothyfroehlich: Declining — the premise does not hold. Per the [permissions docs](https://code.claude.com/docs/en/permissions): a trailing ` *` "enforces a word boundary, requiring the prefix to be followed by a space **or end-of-string**." So `Bash(git push --force-with-lease *)` already matches the bare no-refspec form. —Claude

.claude/settings.json
timothyfroehlich: Declining — the stated root cause is inverted. The docs say the two forms are equivalent: "The `:*` suffix is an equivalent way to write a trailing wildcard, so `Bash(ls:*)` matches the same commands as `Bash(ls *)`." Both match end-of-string, so `git rebase *` already covers bare `git rebase`, exactly as `rm -rf:*` covers bare `rm -rf`. —Claude

timothyfroehlich:
Claude review of head d084c14 — 3 findings: 1 real (git clean flag-order gap) fixed in d084c14a; 2 refuted against the permissions docs and declined with replies

PR #1768, opened 2026-07-28
fix(workflow): teach the merge gates the Claude fallback, add --automerge (PP-jw0s)

timothyfroehlich:
Found while landing #1767, where the gates first said `PASS: reviewed: Copilot review covers head commit` against a Copilot comment that had reviewed nothing.

## Three defects in the review-state gates

**1. A Copilot non-review counted as a review.** When Copilot is quota-limited — or finds no analyzable files — it still posts a review *object* carrying a real login and `submitted_at`. Both gates selected on login alone, so a review that read nothing satisfied them. That's strictly worse than no review, because no-review already has an honest path (the SHA-pinned Claude marker). Both gates now match those bodies and treat them as absent.

**2. `currency` didn't know the Claude marker exists.** `reviewed` checks the marker first and short-circuits. `currency` had no marker check at all — so a head commit that demonstrably *had* been reviewed still sat out the full 600s timer waiting on a reviewer that can't respond until quota resets ~Aug 1. Pure dead time. Both gates now honour it.

**3. The two gates had drifted.** They asked nearly the same question using the same duplicated three-line jq, which is exactly how only one of them learned about the fallback. Extracted `_latest_copilot_review_at` and `_claude_marker_present`.

Filtering non-reviews also fixes an inverted incentive: a PR Copilot never touched hit `SKIP` instantly, while one carrying a *useless* comment served the full 600s wait. Having received a worthless review was penalized relative to having received none. With non-reviews invisible, both take the SKIP path.

### Behaviour change worth knowing

During a quota outage a PR now needs a real review, attested with `mark-claude-review.sh`. Previously the quota comment waved it through. That's the point — but it's **more** friction than yesterday, not less, and it'll bite Dependabot PRs too.

## New: `-a` / `--automerge`

Polls the gates instead of evaluating them once, and merges the moment they all pass — so a PR can be fired off before CI finishes instead of coming back to it. Terminates in exactly one of three states, each named on exit:

| State | Meaning | Exit |
|---|---|---|
| `MERGED` | gates went green, PR squash-merged | 0 |
| `RED` | a gate hard-failed; no merge, `ready-for-review` dropped | 1 |
| `TIMED OUT` | budget expired; PR untouched, label intact | 2 |

A WAIT (CI running, review pending) keeps it polling — only a hard failure stops it. `AUTOMERGE_TIMEOUT` (default 3600s) and `AUTOMERGE_POLL_INTERVAL` (default 30s) tune it; mutually exclusive with `--dry-run`.

This required the gate runner to distinguish `rc=2` (transient) from `rc=1` (hard), which it previously collapsed into one bucket. **The one-shot path still reports both as failures, byte-identical to before** — pinned by a test.

It also re-reads head before merging in the automerge path, so `--match-head-commit` pins the commit the gates just approved rather than one read before a long wait.

Output stays readable over a long poll: the gate block prints on the first poll and again only when the picture changes, instead of repeating five lines every 30s.

## Verification

19 new tests driving the real bash through a stubbed `gh` — the jq filters and the BSD/GNU date branching are exercised, not mocked.

- **12 gate tests** covering both non-review bodies, marker/no-marker, wrong-SHA marker, mixed real+non-review, and the WAIT/WARN boundary.
- **7 automerge tests** covering all three terminal states plus flag validation and the unchanged one-shot path.

Ran the gate suite against `origin/main`'s `_pr-gates.sh`: **the 5 regression tests fail there and the 7 unchanged-behaviour tests pass both ways** — including reproducing the original false-green.

`pnpm run check` green (1807 unit, 217 pytest, shellcheck clean).

Closes PP-jw0s.


timothyfroehlich:
Claude review of head 31e7936 — medium review: 5 findings — 4 fixed (head-SHA TOCTOU, CI-not-registered false RED, oversold docs, stale label snapshot), 1 addressed differently (tightened non-review patterns rather than anchoring, which would eat partial reviews)

PR #1774, opened 2026-07-31
fix(scripts): worktree cleanup stops reporting success on two silent failures (PP-omz3, PP-3w4g)

timothyfroehlich:
Two silent-failure bugs in `scripts/worktree_cleanup.py`. They share a PR because they share a file and a failure mode: the script exits 0 while a worktree leaks, and its exit code is the only signal Claude Code's `WorktreeRemove` hook gets.

## PP-omz3 — a missing target path was a silent no-op reporting success

`worktree_cleanup.py /path/that/does/not/exist` warned and `return`ed, i.e. exit 0, having done nothing at all: no `supabase stop`, no Docker volume removal, no slot deallocation. Everything upstream read that 0 as "worktree cleaned", so the slot manifest entry and the Supabase volumes leaked exactly as AGENTS.md warns a raw `git worktree remove` does — except via the sanctioned path, silently.

Exit 0 for a missing path now requires **evidence** of a clean state, using the discriminator the bead asked for: the path absent from **both** `git worktree list` (including prunable registrations) and the slot manifest. Anything else — residue in either, or a source of truth we couldn't read — exits `3`, says exactly what residue was found, and points at `worktree_orphan_sweep.py --apply`. Nothing is reclaimed in that branch, and the message says so rather than implying otherwise.

## PP-3w4g — a failed `docker volume ls` meant zero volumes

The non-zero exit was warned about and then collapsed into `volumes = []`, indistinguishable from "this project genuinely has no volumes". Teardown continued: worktree removed, slot deallocated, exit 0, volumes still on disk.

Unknown is now explicit and exits `4`, following the PP-5o7b / #1746 shape from the sibling script — `DockerNotInstalledError` (a real zero: no binary, no volumes) vs `DockerUnavailableError` (unknown), surfaced as a frozen `VolumeQuery` dataclass. Per the non-blocking review note left on #1746, the invariant is **structural** rather than conventional here: `__post_init__` raises if an unknown result carries volumes, so no count derived from a query that didn't run can be reported or acted on.

The worktree and slot are still reclaimed in the unknown case, deliberately: the sweep matches on the Docker label, not the manifest, so the leaked project surfaces on the next sweep — a delayed leak, not a permanent one, which is the bead's own reasoning and why it's P3. But the run did not finish the job, so it must not report success.

The query itself is untouched, as PP-3w4g requires: the `--filter label=<k>=<v>` + `-q` form is already Podman-portable, so the template half of #1746 does not apply. Only the error handling is mirrored.

## One change outside the two beads

`.claude/hooks/worktree-remove.sh` had its own `[ ! -d "$WORKTREE_PATH" ] && exit 0` pre-check, which would have made the PP-omz3 fix **unreachable from the hook path** — the script would never be consulted for the exact case it now adjudicates. The hook now delegates that decision and reports the script's exit code instead of flattening every failure to `"cleanup failed"`. The hook still exits 0 itself: a failing `WorktreeRemove` hook would wedge Claude Code's own teardown.

## Deliberately out of scope (both beads list these)

- The `supabase stop` warning path — a stack that was never started legitimately looks like an error; that non-fatal handling is intentional.
- The `docker volume rm` failure path — different bug, not filed.
- `worktree_orphan_sweep.py` — already fixed in #1746, and not refactored to share the new machinery. The duplication is annotated in place; the sweep already imports from this module, so hoisting later is a one-liner, but doing it now means editing the sweep for no behaviour change.
- The branch-derived `project_id` in the volume query. #1773 (PP-4936) notes that a branch renamed after setup makes that label wrong; pinning the id is filed separately, and PP-3w4g explicitly says not to rewrite this query.

## Testing

27 new tests, everything mocked at the `subprocess` boundary in the style of the PP-5o7b suite — nothing touches the real Docker daemon, the real slot manifest, real worktrees, or real beads.

- `scripts/tests/test_worktree_cleanup.py` (21): both false-zero shapes, the `VolumeQuery` invariant, the both-absent idempotent success, each residue source independently, unreadable manifest, failed `git worktree list`, plus regression cover for the two pre-existing contracts the `sys.exit` → `return` conversion could have broken (main-worktree refusal, failed removal keeps the slot).
- `scripts/tests/test_worktree_remove_hook.py` (6): the hook-level *reachability* of the PP-omz3 fix — a missing path reaches the script, exit 3 and exit 4 surface as `INCOMPLETE` instead of being flattened, a clean run stays quiet, and an absent/null payload path still skips cleanup.

`pnpm run check` green (includes `ruff` + `check:pytest`); `shellcheck` clean on the hook.

Closes PP-omz3, PP-3w4g.

https://claude.ai/code/session_01DbU3rdz9X3LJptWiR6dUTL


timothyfroehlich:
Claude review of head c02e4bf — Reviewed the full diff (+893/-33) at head c02e4bf0. **Verdict: approve.**

**The load-bearing fix is the hook, not the script.** `worktree-remove.sh` used to `exit 0` on its own `[ ! -d "$WORKTREE_PATH" ]` pre-check, so `worktree_cleanup.py` was never consulted on exactly the path PP-omz3 is about — fixing the script's exit code alone would have been inert in production. `test_worktree_remove_hook.py` locks that reachability down with stubs on PATH.

**Verified by inspection:**
- `|| echo "... exited $? ..."` does capture python3's status — `$?` expands when `echo` runs, and python3 is the immediately preceding command.
- `worktree_orphan_sweep.py` imports only `MANIFEST_PATH` and `deallocate_slot`; both are unchanged, so `main() -> int` is not a breaking change for the one importer.
- `FileNotFoundError` is caught before `OSError` in `_run_docker` — subclass ordering is correct, so a missing binary reads as a genuine zero and a broken daemon as unknown.
- Residue detection resolves paths on both sides (`slot_manifest_paths`, `registered_worktree_paths`, and the target), so the `/home` → `/var/home` symlink on this host cannot fake a "no residue" verdict.
- `VolumeQuery.__post_init__` makes unknown-implies-empty structural rather than conventional. Good call.

**Two non-blocking follow-ups** (filing beads, not asking for changes here):

1. **The new diagnostic is discarded by the one non-hook caller.** `scripts/workflow/stale-worktrees.sh:151` runs `worktree_cleanup.py ... 2>/dev/null || echo "WARN: Failed to remove $branch"` — stderr is where the whole reason string lives, and exit 3 vs 4 both collapse to one generic message. `--clean` gets more accurate (it now warns where it used to stay silent) but tells the user nothing actionable.

2. **The PP-qlzu path still exits 0 without evidence.** When `.git` is absent the branch can't be derived, so the entire `if git_marker_present:` block — including `list_project_volumes` — is skipped, and the run ends at `return EXIT_OK` having never looked for volumes. That's the same "success without evidence" class this PR closes, just reached differently; `EXIT_DOCKER_UNKNOWN` looks like the honest code there.

`EXIT_DOCKER_UNKNOWN = 4` here vs `= 1` in the sweep is deliberate and documented; tracked separately as PP-v8g1. —Claude

PR #1775, opened 2026-07-31
fix(worktrees): pin the Supabase project id per worktree (PP-4936)

timothyfroehlich:
Fixes PP-4936.

## The bug

The Supabase docker project id was re-derived from the git branch name on every post-checkout run. Because that id names the containers and labels the volumes, `git checkout -b` inside a worktree whose stack was already running renamed the stack out from under itself:

- `supabase stop` targeted an id that matched nothing — a silent no-op.
- The old containers kept the slot's ports bound, so the next `supabase start` died with `rootlessport listen tcp 0.0.0.0:56822: bind: address already in use`. `db:reset` / `db:_restart` stayed broken in that worktree until the containers were stopped by their real name.
- `worktree_cleanup.py` looked the Docker resources up under the *new* id and left the old containers and volumes behind — a systematic orphan generator, not a rare accident.

AGENTS.md §5 tells agents to `git checkout -b` inside a worktree, so this was on the happy path.

## The fix

The project id is now derived from the branch only on **first** setup and pinned after that. `resolve_project_id()` reuses the id already recorded in the worktree's generated `supabase/config.toml` — the file the Supabase CLI itself reads, and therefore the authoritative record of what the running stack was started under. `worktree_orphan_sweep.py` already makes exactly this assumption (it reads `config.toml` first and only falls back to branch derivation), so this makes the two agree instead of drift.

A pinned id is honored only when it has the shape `branch_to_project_id` emits (`pinpoint-` + `[a-z0-9-]`, ≤ 40 chars):

- `worktree_orphan_sweep.py` identifies PinPoint-owned resources by the `pinpoint-` prefix, so honoring an out-of-shape id would create containers the sweep can't see.
- It also rejects the template's bare `project_id = "pinpoint"`, so a `config.toml` copied straight from the template still gets a real per-worktree id instead of one shared by every worktree.
- Validation uses `fullmatch`, not `^...$` — `$` accepts a trailing newline, which would have been written back into `config.toml` inside the id.

`read_pinned_project_id()` never raises (catches `OSError` **and** `UnicodeDecodeError`, which `read_text()` raises on undecodable bytes and which is a `ValueError`, not an `OSError`). This runs from the post-checkout hook, where an exception would skip the rest of the worktree's config generation.

The ports were always slot-derived and are unaffected; only the project id changes behavior.

## Design call

The bead offered two candidate fixes and left the choice open. I picked "pin the existing id" over "derive the id from the worktree path or slot", because deriving from path/slot changes the id for **every existing worktree on the host** on its next checkout — instantly orphaning every currently-running stack, and breaking `worktree_cleanup.py`'s `branch_to_project_id(branch)` call for all worktrees rather than just renamed ones. Pinning is a strict improvement with no churn: existing worktrees keep the id their containers already use, and no consumer's behavior gets worse.

## Known remaining gap (out of scope here)

`worktree_cleanup.py` still derives the id from the branch for its `docker volume ls --filter label=com.supabase.cli.project=…` query, so on a renamed branch that query targets the wrong label and the volumes still leak. Pinning does fix the worse half of the same code path — cleanup's `supabase stop` runs with `cwd` set to the worktree, so the CLI reads the pinned `config.toml` and now stops the right containers. The remaining one-line change belongs in `worktree_cleanup.py`, which is owned by a concurrent PR (PP-omz3 / PP-3w4g); filed separately rather than reached across. `worktree_orphan_sweep.py` still reconciles any volumes that leak in the meantime.

## Verification

- `pnpm run check` green (1813 unit tests, 214 files; 60 in `test_worktree_setup.py`).
- 10 new tests covering: reading a generated id, fresh worktree (no `config.toml` → derive), no `project_id` key, commented-out `project_id`, the bare template value, four malformed-id shapes, trailing-newline rejection, non-raising behavior on a directory / undecodable bytes, pinned-id survival across a rename, and a generate→resolve round trip.
- Verified in this worktree that `supabase/config.toml` is gitignored, so a fresh worktree genuinely has no id to pin, and that `worktree_setup.py` is the only writer of that file.
- Note for reviewers: `.husky/post-checkout` runs `worktree_setup.py` **from the main worktree**, so this fix only takes effect for other worktrees once it's on `main`. Live before/after in a feature worktree isn't possible without editing the root checkout, which is read-only; the round-trip test covers the composition instead.

https://claude.ai/code/session_01DbU3rdz9X3LJptWiR6dUTL


timothyfroehlich:
Claude review of head d1e7c20 — Reviewed the full diff at head d1e7c20f. **Verdict: approve.**

**The design question I had is now answered in the code.** The bead didn't prescribe a fix, and `resolve_project_id`'s docstring plus the AGENTS.md bullet make the reasoning explicit: the project id names the Supabase containers and labels the volumes, so re-deriving it from the branch on every checkout renames the stack out from under a *running* instance — `supabase stop` then matches nothing, the old containers keep the slot's ports bound so `supabase start` fails with "address already in use", and every id-keyed cleanup path misses the old resources. Pinning in `config.toml` (the file the Supabase CLI itself reads, and the same source of truth `worktree_orphan_sweep.py` assumes) is the right place for it.

**Verified the pin-acceptance gate rejects everything it should:** the bare template `project_id = "pinpoint"` (otherwise every worktree would share container names), non-`pinpoint-` prefixes (invisible to the orphan sweep), uppercase and underscores (illegal Docker/Supabase ids), over `MAX_PROJECT_ID_LEN`, and a trailing newline — `fullmatch` rather than a `$`-anchored search is the right tool for that last one, and the test names the trap.

The exactly-40-char case is the one I'd have expected to break, since every `worktree-agent-*` and `feat/…-PP-…` branch hashes to exactly the limit — `test_hashed_max_length_id_is_pinnable` covers it, and `> MAX_PROJECT_ID_LEN` is the correct comparison.

`read_pinned_project_id` swallowing `OSError`/`UnicodeDecodeError` is correct here specifically because this runs from `post-checkout`, where a raised exception would skip the rest of the worktree's config generation.

**One cosmetic note, not worth a change:** `_PINNABLE_PROJECT_ID_RE = pinpoint-[a-z0-9-]*` also accepts a bare `pinpoint-` and `pinpoint--foo`. `branch_to_project_id` can't emit either, so nothing reachable hits it.

Adding `project_id=` to the setup summary line is a genuine debugging improvement — the pinned-vs-derived divergence is otherwise invisible until something breaks. —Claude

PR #1776, opened 2026-07-31
fix(agents): stop instructing agents to invoke /code-review (PP-lsb4)

timothyfroehlich:
## Problem

`pinpoint-orchestrator` told every dispatched subagent to self-review by running `/code-review`, calling it "the model-invocable local review". It isn't — `/code-review` is a Claude Code harness built-in that only the user can trigger. There is no `code-review` skill on disk and it never appears in the model-invocable skill list.

A dispatched agent hit this live on 2026-07-31 ("The `/code-review` skill is not model-invocable here") and burned a turn discovering the instruction was impossible. The worse case was the **lead backstop** (`Phase 4 → Ensure every PR is reviewed`): an impossible instruction there means the backstop silently doesn't happen, and the PR reaches Tim with only a `mark-claude-review.sh` marker attesting a review that was never performed in the prescribed way.

## Fix

Every site now prescribes what an agent can actually do: a deliberate manual pass over its own diff against `REVIEW.md` (the canonical rubric) plus the bead's acceptance criteria and out-of-scope list. The existing accurate note that `ultra` is user-triggered/billed and unlaunchable by an agent is preserved, and `/code-review` is now described the same way rather than as model-invocable.

The three sites named in the bead:

- `.agents/skills/pinpoint-orchestrator/SKILL.md:136` — Phase 3 agent prompt template
- `.agents/skills/pinpoint-orchestrator/SKILL.md:208` — Phase 4 lead backstop
- `.agents/skills/pinpoint-orchestrator/references/agent-prompt-template.md:16` — annotated template

Plus the same defect elsewhere, per the bead's repo-wide acceptance criterion:

- `.agents/skills/pinpoint-pr-workflow/SKILL.md:183` — Phase 3.4 "Claude fallback review" step 1 (an identical invoke instruction)
- `.agents/skills/pinpoint-pr-workflow/SKILL.md:292` — §4.4 escape-hatch pointer back to that fallback
- `.agents/skills/pinpoint-superpowers-bridge/SKILL.md:65,88` — claimed the authoritative review gate was "CI Gate + `/code-review`"
- `scripts/workflow/mark-claude-review.sh:13` — header comment asserting `/code-review` is "the model-invocable local review"

Left alone deliberately (descriptions, not invoke instructions — the bead says these are fine): `REVIEW.md`, `superpowers-bridge:59` (accurately calls it a built-in), and the dated records under `docs/plans/` and `docs/superpowers/`.

[...]

PR #1777, opened 2026-07-31
feat(memory-review): weekly curated pass over recorded context (PP-uoqg)

timothyfroehlich:
Adds a weekly human-in-the-loop pass that reviews everything recorded about how we work, proposes changes, and applies only what Tim approves.

**Bead:** PP-uoqg · **Spec:** `docs/superpowers/specs/2026-07-28-memory-context-review-design.md` · **Plan:** `docs/superpowers/plans/2026-07-28-memory-context-review.md`

## Why

Two failures, one skill.

**Recorded context rots and nobody looks at it.** The tmux `-CC` server-wedge fact currently lives in three places — a Bazzite memory file, `dotfiles/mac/.zshrc.local` comments, and the `bazzite` skill — and the single most useful part, the `kill -WINCH` recovery, exists only in the copy the Mac cannot see.

**Claude auto-memory is per-machine and syncs nowhere.** The store directory name encodes the absolute repo path, so the Mac (`-Users-froeht-Code-PinPoint`) and Bazzite (`-var-home-froeht-Code-PinPoint`) hold **fully disjoint** sets: 22 files vs 11, zero filenames in common. Bazzite has known that tmux root cause since 2026-07-16; the Mac rediscovered it from scratch on 07-27 and recorded a worse version. Upstream issues #25739 and #56793 are open with no response, and no third-party tool solves the path-encoding problem — the most complete one requires identical repo paths on every machine, which Fedora Atomic's `/var/home` rules out.

## The rule

Every destination differs in when you pay for it — hooks are free and always fire; `CLAUDE.md` costs every session; `bd remember` costs every session *unfiltered*; rules fire on path match, skills on task match, auto-memories on description match.

**Route each fact to the cheapest tier that still guarantees it fires.** Read from the other end that is PP-22e4's "highest tier it qualifies for", and the two agree.

The review hunts three defects: too expensive for its value (demote), too cheap to fire when needed (promote — including *could this be a hook?*), and present in more than one tier (dedupe, merging first where copies carry distinct detail).

## What's here

| File | |
| :--- | :--- |
| `scripts/memory_review/collect_stores.py` | Enumerates one machine's stores + context files as JSON |
| `scripts/memory_review/snapshot_stores.py` | One-way timestamped copy with retention pruning — the undo the store otherwise lacks |
| `scripts/memory_review/apply_remote.sh` | Ships an approved manifest and lets the far machine's own Claude apply it |
| `.agents/skills/pinpoint-memory-review/` | The runbook, plus a routing reference loaded only when routing a fact |
| `.agents/skills/pinpoint-chores/SKILL.md` | New checklist item (mirrored on bead PP-qehv) |

Both Python scripts are **standard library only with no `subprocess`** — enforced by an AST test — so they can be piped to a remote `python3` over SSH, where Bazzite's non-interactive shell has no brew on PATH.

## Design decisions worth reviewing

- **No git-backed memory store.** Tim rejected it over coupling to Anthropic's undocumented on-disk format and merging the `MEMORY.md` index. This skill is the sync mechanism instead, which bounds divergence at ~a week.
- **The remote half runs the far machine's own agent**, not file writes over SSH, so each machine's memories are written using whatever the sanctioned mechanism is at the time. Hand-writing another machine's store would hard-code today's format — the same fragility that ruled out git-backing.
- **One round trip, no negotiation.** The remote agent may *dispute* an action; disputed items are left unapplied and escalated to Tim, never argued between agents.
- **The veto list is presented in-session, one line per item.** A long document gets rubber-stamped, which would turn the gate into theatre.

## Verification

`pnpm run check` green. 26 new tests. Beyond unit coverage, run against the real machines:

- The collector on the live Mac store immediately found a genuine defect: `project_beads_server_cutover.md` has **no line in `MEMORY.md`**, making it unreachable by description-based recall.
- A real snapshot captured all 23 files including the index.
- `apply_remote.sh --dry-run` ran end-to-end against Bazzite: remote snapshot taken (12 files), then stopped without shipping.
- The remote permission posture was spiked rather than assumed — Bazzite's `defaultMode: auto` already lets a headless `claude -p` write unattended, verified end to end. `--stage-only` remains as the opt-in manual path.

## Not in scope

Auditing skill *contents* (deliberately excluded as too much for one pass), continuous file sync, and any scheduler — the existing `weekly-chore` bead already nags on either machine and clears everywhere.

## Follow-up spotted, not fixed here

Bazzite's `SessionEnd` hook fails with `node: command not found` on every headless `claude -p` session — non-interactive SSH has no brew PATH. Harmless to this work (exit 0) but it means that cleanup hook never runs there.


timothyfroehlich:
## Code review round — all 7 findings addressed

Ran `/code-review medium` and fixed every finding; none were declined. Tests went 26 → 42.

**HIGH — stale result document.** `apply_remote.sh` read `/tmp/memory-review-result.json` at a fixed path that was never cleared. A remote run that exited without writing one would return the *previous* run's document, and the lead would report last week's ids as applied this week — the exact false-sync this feature exists to prevent. Now cleared before dispatch, and the result is schema-validated before it is trusted.

**`--repo` did not actually scope the corpus.** `~/.claude/projects` holds a store for every repo Tim has ever opened; all of them landed in the review set indistinguishable from PinPoint's. A pass could have proposed deleting another project's memories. They now come back as `other-project` and are explicitly out of scope.

Fixing that introduced a second bug I caught only by re-checking against the live store: run from a worktree, `--repo` is the worktree path, so the main checkout's 22 memories were classified `other-project` and dropped from the corpus entirely. Green tests did not catch it. Added `main_checkout()`, which resolves a worktree path back to the checkout it belongs to — by string, since the module must stay subprocess-free to survive being piped to a remote interpreter.

**Also fixed**
- Line counts now match `wc -l`. They feed PP-22e4's ≤10-line `AGENTS.md` CI gate, where an off-by-one reads a compliant stub as a violation.
- Snapshot pruning ignores directories that are not snapshots. The design expects humans to restore by hand from that root, and name-ordering made it worse than it sounds — digits sort before letters, so a stray `restored/` survived while a real oldest snapshot was destroyed in its place.
- `--manifest` with no value tripped `set -e` on `shift 2` before the usage message could print.
- `--dry-run` and `--stage-only` are now mutually exclusive.
- The remote agent's exit status no longer aborts retrieval — it may have applied part of the manifest before failing, so the record of what changed has to come back regardless.

**One test was wrong rather than the code:** `assert "subprocess" not in source` matched the module docstring explaining why it uses none. Replaced with an AST import allowlist — a stronger check, not a weaker one.

**Merged `main` twice.** The second conflicted on `AGENTS.md`: PP-22e4 deleted the skills table outright, so this skill needs no registration row — skills are discovered by frontmatter description now. `REVIEW.md`'s list is reviewer-facing and correctly excludes it.

Last push fixes a CI failure that only reproduced against the merge result: #1763 retired `CORE-ARCH-002`, which `references/routing.md` cited by ID.

—Claude

timothyfroehlich:
Claude review of head 11f6df7 — /code-review medium at bf1d91a1: 7 findings, all fixed. Delta review of 11f6df77: clean auto-merge of #1763+#1776, no manual resolution; only own change is dropping the retired CORE-ARCH-002 citation.

PR #1778, opened 2026-07-31
refactor(agents): absorb docs/patterns into the owning skills (PP-22e4)

timothyfroehlich:
PR 5 of the Claude-native context system rebuild (PP-22e4).

## What this does

Deletes `docs/patterns/` (17 files), `docs/PATTERNS.md`, `docs/E2E_BEST_PRACTICES.md`, and the `pinpoint-patterns` skill — absorbing only the audited survivors into the skill that owns each subject, and repointing every citation.

Every claim was verified against real source before it moved. Where a doc contradicted the code, the code won and the claim was dropped rather than carried forward.

One exception, recorded because it is the point of having a review gate: an early draft asserted "135/135 exports follow the `Action` suffix." That number came from a mis-flagged `rg -oh` (ripgrep's `-h` is `--help`, so the command printed its own 135-line help text for any pattern). A fact-validation pass caught it — the real figure is 49 of 59, with ten live exceptions including `inviteUser`, which is wired straight into `useActionState`. The rule is now stated as house style for new code, with an explicit "don't rename existing ones."

That finding prompted a sweep of every other numeric claim in the diff. **All census-style counts have been removed** rather than re-pinned to today's numbers — this PR exists because precise claims rot silently, and fresh counts would just restart the clock. Categorical claims are kept deliberately (`revalidateTag` has zero usages; there is no `src/server/data-access/`; there is no db→app converter layer), because they tell you what *not* to go looking for and don't drift by one every sprint.

## Where the content went

| Deleted | Survivor → destination |
| :--- | :--- |
| `patterns/server-action-forms.md` | `pinpoint-ui` → **Server Action Forms** (Radix Select carve-out, both bug shapes, why controlling the Select doesn't help, the don't-mock-`useActionState` rule) |
| `patterns/create-form-reset.md` | `pinpoint-ui` → **CREATE form reset**, plus the PP-lql stale-option-ID section. Two fixes applied: the false `InviteUserDialog` react-hook-form claim (it remounts on reopen), and the missing PP-1ajq reset-before-clear ordering rule |
| `patterns/config-driven-enums.md` | `pinpoint-ui` → **Config-Driven Enums (authoring a new one)** — template only; the "never freestyle status colors" rule already exists in two places |
| `patterns/ui-patterns/discriminated-union-props.md` | `pinpoint-ui` → ~3 sentences, with the nonexistent `IssueConsistency` variant corrected to `frequency` and the missing `size` prop added |
| `patterns/url-construction.md` | `pinpoint-security` → **§4 Site URL Construction & Safe Redirects**, extended with `requireSiteUrl` / `getSafeRedirect` and the one live violation |
| `patterns/severity-naming.md` | `pinpoint-design-bible` → **§24 Severity Vocabulary** |
| `patterns/database-migrations.md` | `pinpoint-deployment` → **Database Migrations (day-to-day)**, plus CORE-ARCH-009 and `scripts/ensure-test-schema.ts` |
| `patterns/machine-status-derivation.md` | `pinpoint-ui` → the principle only, citing `src/lib/machines/status.ts`. The doc's inlined algorithm had drifted to being behaviorally wrong |
| `patterns/logging.md` | `pinpoint-ui` → pointer to canonical `docs/LOGGING.md` + the one unique fact (`~/lib/logger` imports `fs`, so `console.*` is correct in client components) |
| `patterns/type-boundaries.md` | `pinpoint-typescript` → there is no db→app converter layer and none should be built |
| `E2E_BEST_PRACTICES.md` | `pinpoint-e2e/references/e2e-best-practices.md` → the Environment Defaults block (autologin + 3 opt-outs, global-setup chain, cleanup API) |
| `patterns/{authentication,data-fetching,mutations,file-organization,domain-rules,testing-patterns,service-layer-transactions}.md`, `PATTERNS.md`, `pinpoint-patterns` | nothing survived — audited |

## Routing

Every skill whose subject set grew also had its `description` frontmatter extended with the distinctive terms an agent would search for — the description is the only text consulted when routing, so new content is unreachable without it. No pre-existing term was dropped from any description.

## Not touched

Frozen records keep their historical references: `docs/superpowers/**`, `docs/plans/**`, `docs/testing/e2e-audit-2026-05.md`.

`pnpm run check` green.


timothyfroehlich:
Claude review of head 1170aa0 — no serious findings

PR #1796, opened 2026-08-01
fix(lint): make ESLint actually run on e2e/ and scripts/ (PP-ojv5)

timothyfroehlich:
## Summary

`lint` was `eslint src/ --quiet`. `e2e/` and `scripts/` are sibling directories, so **neither had ever been linted** — every rule the flat config matched against them was dormant.

`eslint.config.mjs:251` asserted the opposite:

> the CORE-ARCH-011 backstop instead lives in a separate custom rule … in the main `**/*.ts(x)` block above, **which also matches `e2e/**/*.ts(x)` — so e2e files are covered at "error" with no hole**

True of *match* scope, false of *execution* scope. This PR makes it true rather than deleting it — the surrounding comment records a real decision (flat config REPLACES rather than merges rule options, so that block can't also carry the side-effect selectors).

## What running them actually turned up

**`e2e/` — 0 errors, 8 warnings.** The hole was real but empty. Nothing to fix; wiring it up just stops the next side-effect-in-transaction in an e2e file from slipping through.

**`scripts/` — 137 errors**, which decompose into two very different things:

- **133 `no-undef`** on `console` / `process`, across 11 `.mjs` files. Not bugs — a missing Node globals block, which never existed because these files were never linted. Added, mirroring the existing `supabase/**/*.mjs` block.
- **4 real `preserve-caught-error`**, all in `check-prod-backups.mjs`. That script is the weekly chore that verifies prod backups exist and retention is intact; it threw without `cause`, discarding the underlying Supabase CLI failure — which is precisely the information you need when the chore goes red. Fixed.

## Cost

Lint 15.1s → 17.1s standalone. `check` parallelises, so less in practice.

## Deliberately not included

The CI-only guard on `e2e:full`/`e2e:all` and the no-spec-path guard on `playwright test`. Different feature — they get their own PR rather than making this one broad.

## Test Plan

- `pnpm exec eslint src/ e2e/ scripts/ --quiet` — exit 0.
- `pnpm run check` — green (218 test files, 2004 vitest, 456 pytest).

## Related Issues

PP-ojv5. Found while re-planning PP-22e4.


timothyfroehlich:
Post-merge `/code-review` pass on `66417891`. No correctness bugs; three low-severity findings, tracked as **PP-ru2p**.

**Verification first:** `pnpm run lint` exits 0, and re-running with `--format json` confirms the coverage claim isn't vacuous — 61 `e2e/` `.ts` files (0 errors, 8 warnings) and 16 `scripts/` files (0 errors) are genuinely linted now. `scripts/*.ts` resolve to a real project (`tsconfig.app.json` includes root-relative `**/*.ts`), so `projectService` doesn't blow up. The core change does what the commit message says.

---

**1. `package.json:19` — `.mjs` is now gated in CI but by nothing locally.**

`lint` now covers `scripts/**/*.mjs`, and 11 of the 16 newly-linted files are `.mjs`. But `lint-staged` (`package.json:204`) runs `eslint --fix` on `*.{ts,tsx}` only, pre-commit runs just lint-staged + typecheck + format, `.husky/pre-push` runs no lint at all, and Prettier's `format` glob (`**/*.{js,jsx,ts,tsx,json,css,md}`) excludes `.mjs`.

Scenario: edit `scripts/db-fast-reset.mjs`, introduce a `no-undef` or an unformatted block, commit and push — every local hook passes, then CI Gate goes red on a file no local gate ever touched.

**2. `eslint.config.mjs:309` — glob narrower than its sibling.**

The new Node-globals block keys on `scripts/**/*.mjs` + `scripts/**/*.ts`, while the relaxed-rules block at line 287 uses the broader `scripts/**/*`. A future `scripts/foo.cjs` or `scripts/foo.js` falls outside the globals block and immediately produces the exact 133-error `no-undef` wall this block was added to clear. Loud rather than silent, but `scripts/**/*.{mjs,cjs,js,ts}` avoids the rediscovery.

**3. `scripts/check-prod-backups.mjs:202` — a fifth cause-dropping `throw` was left behind.**

`catch { throw new Error("Could not parse CLI output as JSON …") }` uses an omitted catch binding, so `preserve-caught-error` doesn't flag it (`requireCatchParameter` is off) — but it discards the `SyntaxError`, which is the same information loss the other four sites were fixed for. If the CLI emits a truncated or HTML response, the weekly chore fails with 400 chars of output and no parser position. Partially mitigated by the existing `stdout.slice(0, 400)`.

—Claude

PR #1800, opened 2026-08-01
feat(lint): oxlint local mirror + make `check` a static gate (PP-4zcj)

timothyfroehlich:
Closes **PP-4zcj** — TS7 Phase 3, the last open phase of the TypeScript 7 rollout.

## Why now

Phase 3 has been parked since 2026-07-02 on "revisit when oxlint type-aware is production-stable." **That cleared on 2026-07-22**, when oxc shipped [type-aware linting stable](https://oxc.rs/blog/2026-07-22-type-aware-linting-stable) — oxlint v1.75.0, `oxlint-tsgolint@7` tracking TypeScript 7.0.2, 59 of typescript-eslint's 61 type-aware rules. The 2026-07-12 spike on this branch had already cleared its gate against the *alpha*; stable removes the alpha-coverage objection that made waiting correct. Lands on oxlint 1.76.0 / oxlint-tsgolint 7.0.2001.

## Shape — unchanged from what we agreed on 2026-07-12

**Local mirror, authoritative CI backstop.** `eslint.config.mjs` stays complete and authoritative, and **CI keeps running full ESLint exactly as today** (`ci.yml` is untouched) — lint isn't on CI's critical path, E2E is, so there's no reason to bet coverage on a second engine there. Only the *local* path swaps: `pnpm run check` now runs `lint:local` = `oxlint` (type-checked bulk) ∥ a slim ESLint pass (residual plugins, project service off).

Drift **fails safe**: anything the mirror misses, CI's full ESLint still catches — worst case a CI-only failure, never silent loss.

The slim pass is the same `eslint.config.mjs` behind `PINPOINT_LINT_SLIM=1`, using typescript-eslint's own `disable-type-checked` set rather than a hand-listed one, so a type-aware rule added later is handled automatically instead of crashing the slim pass. With the flag unset — which is always, in CI — the config CI evaluates is unchanged.

## Numbers (3-run medians, this repo)

| | wall | peak RSS |
| :-- | --: | --: |
| `lint` (full ESLint, authoritative) | 14.86s | 3152 MB |
| `lint:_oxlint` | 0.94s | 932 MB |
| `lint:_slim` | 3.47s | 1182 MB |
| **`lint:local`** | **3.76s** | **~1.2 GB** |

**`pnpm run check` wall-clock is essentially unchanged** and that is expected, not a regression — `check` runs its legs in parallel and its long pole is the ~40s unit-test leg, not lint. The wins are the standalone lint path (~4×) and the **3.2 GB → 1.2 GB** memory drop, which is what actually matters on a 16 GB box running several agent sessions against `FORCE_MEM_PRECHECK`.

## Fidelity: zero false negatives

Seeded-probe union test — for every rule the full config caught, oxlint ∪ slim caught it too: `no-explicit-any`, `no-floating-promises`, `no-unsafe-assignment`, `restrict-template-expressions`, `no-base-to-string`, `ban-ts-comment`, `explicit-function-return-type`, `no-unnecessary-condition`, plus residual `pinpoint/no-side-effects-in-transaction` (CORE-ARCH-011), `unused-imports`, `better-tailwindcss`, `eslint-comments`, `react-hooks`, `jsx-a11y`. `no-misused-promises` with `checksVoidReturn.attributes: false` stayed faithful — an async JSX `onClick` is correctly **not** flagged. Probes deleted before commit.

## Three findings worth reviewing

1. **`no-unnecessary-condition` is still nursery in oxlint**, so `@oxlint/migrate` silently omits it — a false negative on a rule we set to `"error"`. Now listed by hand in `.oxlintrc.json`, along with the two per-override `"off"`s (tests, e2e) that `eslint.config.mjs` has. This is the main ongoing maintenance hazard and it's documented in `AGENTS.md` § "Lint engines".
2. **`outDir` is _not_ inert on these tsconfigs** — contrary to the original spike note, which called it "inert for all real gates." tsgolint needs it (composite projects can't set `noEmit` (TS6310), so `.mjs` inputs hit TS5055 "would overwrite input file"), but adding it makes composite demand every program file be listed, which **broke `pnpm run typecheck` with 4 × TS6307** on JSON imports. Fixed by adding `content/*.json` + `src/**/*.json` to `tsconfig.app.json`'s `include`. Verified only `.tsbuildinfo` is ever written to `.oxlint-tsbuild/`, never JS, and `*.tsbuildinfo` was already globally gitignored.
3. **oxlint found a true positive typescript-eslint missed** — `no-unnecessary-type-assertion` on a redundant `as` in `src/lib/notifications/dispatch.ts`. Confirmed genuinely redundant (typechecks clean without it), so the code is fixed and the rule kept.

## Also in here

- Corrects the stale **"~12s"** figure for `pnpm run check` in `AGENTS.md` (measured ~40s; the number was repeated in the key-commands table and the "which tests to run" list).
- Retargets the chores checklist's TS-7 item, which still pointed at **PP-rl1l** (closed). Only Phase 4 — `next build` on the native compiler — is left, still gated on the TS 7.1 stable JS API (no announced date).

## Verification

- `pnpm run check` — green (2004 unit tests)
- `pnpm run typecheck`, `typecheck:tests`, `typecheck:e2e` — all green
- `pnpm run build` — green (this is the real test of the `outDir` change)
- `pnpm run lint` (authoritative) and `pnpm run lint:local` — both clean on a clean tree

**Not run: `pnpm run preflight`.** This worktree's Supabase stack won't start — `pgsodium_root.key: Read-only file system` / `invalid secret key`, reproducible on a freshly-recreated volume, so it's a host-level issue rather than anything in this diff. The tsconfig risk is fully covered by `typecheck` + `build` above; what's left uncovered is the integration suite over the one-line `dispatch.ts` change. Leaving that to CI deliberately rather than claiming a gate I didn't run.

https://claude.ai/code/session_019SBvbjTjNcS4CoarvpVdhT


timothyfroehlich:
## Update — second change folded in per Tim: `check` is now a static gate

With the lint leg down to ~4s, **unit tests were `check`'s long pole by a wide margin** — ~40s of a ~40s run. They're also not where unit tests conventionally live here: pre-commit never ran them (lint-staged + typecheck + format only), and CI runs `test-unit` as its own required job that `ci-gate` and every e2e job already depend on.

So `check` becomes what its name and its "run this before committing" role imply — a fast **static** gate. **~40s → ~17s.**

Unit tests now live in: CI `test-unit` (required), `preflight` (still runs them, for non-trivial changes), and on demand via `pnpm run test` / `test:watch`. Deliberately **not** added to pre-push — that would just re-add the 30s somewhere less useful.

### One correction to the earlier estimate

The 2026-07-12 plan note predicted "check drops to ~2–3s." It doesn't — measured per-leg:

| leg | time |
| :-- | --: |
| `check:pytest` | **14.47s** |
| `format:fix` | 6.86s |
| `lint:local` | 4.35s |
| `check:linters` | 1.63s |
| everything else | < 0.7s each |

`check:pytest` (456 hook tests) is the new pole and accounts for ~14.5s of the ~17s. Whether those should also move out is a separate question I haven't actioned.

### Docs

Updated everywhere that claimed `check` runs unit tests: `AGENTS.md` (§2.2.2, key-commands table, "which tests to run"), `README.md`, `docs/DEVELOPMENT.md`, and the `pinpoint-testing` + `pinpoint-pr-workflow` skills.

Verified: `pnpm run check` green (~17s), `pnpm run test` green on demand (2004 tests).

timothyfroehlich:
Claude review of head 3b8ad37 — /code-review high ran at d761d37f: 7 findings, 6 fixed. The 7th (prefer-optional-chain missing from the mirror) was re-tested and resolved by enabling the rule and fixing both call sites rather than dropping it. Delta since that review, judged minor by Tim: 2 oxlint rules enabled (no-useless-assignment, prefer-optional-chain), check:python split out of check, a null-guard rewrite in PinballMapLinkField.tsx, a scoped oxlint-disable in lib/cookies/client.ts, and docs. Verified on head: check, authoritative lint, check:python, 2024 unit tests, next build, CI Gate — all green.

PR #1801, opened 2026-08-02
chore(agents): retire Copilot review; the marker is the review gate (PP-4ric)

timothyfroehlich:
## Summary

Copilot review is gone. Its free tier was too small to review PinPoint's PRs — quota outages were the normal state rather than the exception — so this purges it from every live context file, skill, and workflow script.

**The merge bar is unchanged**: a PR still cannot merge without a review covering its head commit, threads resolved. What changed is who reviews. Tim runs `/code-review` on the branch (an agent cannot — it's a harness built-in), and `mark-claude-review.sh` records it, promoted from "fallback for a request Copilot didn't answer" to the only thing that satisfies the `reviewed` gate. Genuinely trivial changes may be self-attested with a stated reason.

Net −1,088 lines.

## Decisions worth a second look

Four things I had to decide rather than mechanically delete:

1. **`threads` now counts unresolved threads from ANY author.** Left filtered to Copilot logins it would have matched nothing and become a permanent PASS — a false green worse than no gate. It also fixes a real hole: Tim's own review comments never blocked a merge under the old filter.
2. **The `currency` gate is deleted.** It existed only to soft-report where a PR stood with Copilot's request clock. With one observable fact left (does a marker pin head?), `currency` and `reviewed` asked the same question and returned the same answer. The merge script runs 4 gates now, not 5.
3. **`--check-ready` reports the review state but no longer gates on it.** That mode decides whether a PR is worth Tim's `/code-review`; requiring the review to have already happened would be circular and permanently red. The `reviewed` gate is still the hard block at merge time.
4. **`.github/instructions/*` deleted along with `copilot-instructions.md`.** Copilot's `applyTo` path-scoping was the only consumer of both. They had also drifted unmaintained — `auth.instructions.md` still asserted a no-JS requirement that CORE-ARCH-012 dropped.

## What this costs

`reviewed` stops being a mechanical gate and becomes a policy one. The only thing that can PASS it is a marker the authoring agent posts itself, and the scripts cannot tell an attested review from a fabricated one — the same honesty model as `merge-pr.sh --force`. Under Copilot the gate checked an independently-produced artifact; it no longer does. That is the real price of losing the bot, and it is worth stating rather than implying. A second reviewer (the Gemini work) restores the independent artifact.

## Gate machinery

Six review states collapse to three — `marker` / `stale_marker` / `unreviewed`. The request timeline, the 600s wait threshold, and the quota-limited non-review body matching all modelled a bot answering on its own clock, so they go with it (PP-lzaw, PP-jw0s resolved by deletion, not regression). Nothing WAITs any more: with no reviewer in the loop there is never an answer already on its way, so a WAIT would poll to timeout.

`stale_marker` keeps the old `pushed_after` diagnostic alive — the PR visibly HAS a review and it isn't of head. That's the state most likely to be misread as a flaky gate.

The review-polling thread in `pr-watch.py` is removed; nothing posts unprompted reviews to poll for. **Note for the Gemini-review session:** if Gemini auto-reviews on push, that thread is what you'd want back, and `_compute_review_state` in `_pr-gates.sh` is where a second reviewer slots in.

## Test Plan

- [x] `pnpm run check` green (2004 unit tests, 454 pytest, lint/format/shellcheck/ruff)
- [x] `scripts/tests/test_pr_gates.py` rewritten — 15 tests over the three states, marker pagination, and the author-agnostic threads gate
- [x] `scripts/tests/test_pr_watch.py` updated — bash/Python vocabulary stays pinned together via the marker prefix and state names
- [x] `check_rule_ids.py` still exits 0 after dropping the two `.github` globs from `CITING_SOURCES`
- [x] Tim's `/code-review` — 3 findings, both code defects fixed in `255342de`:
  - marker reader read the newest comment while the writer rewrote the oldest, so with duplicate markers a reviewed head reported `stale_marker` forever. Fixed at the semantics — the gate now passes on **any** marker pinning head, which no comment ordering can break.
  - `pr-watch.py`'s default mode blocked on unresolved threads, refusing to watch the very push that addresses them. Now a reminder; `merge-pr.sh` still refuses the merge.
  - the honor-based `reviewed` gate is accurate and deliberate — see "What this costs" above.

## Related Issues

PP-4ric

https://claude.ai/code/session_01QRsmNVS2yRseop7yXngxbC


timothyfroehlich:
Claude review of head 255342d — 3 findings: marker reader/writer duplicate disagreement and pr-watch thread block both fixed in 255342de; honor-based reviewed gate accepted as deliberate

PR #1802, opened 2026-08-02
fix(mcp): dedupe retried create_issue, report list_machines truncation (PP-u4ab.4)

timothyfroehlich:
Closes the two real findings from the PR #1707 review of the MCP tool surface, and records the other two as non-defects rather than hardening against bugs that don't exist.

## What was actually wrong

**`create_issue` could never dedupe a retry.** It built a fresh `randomUUID()` on every call, so the service's idempotency check had nothing to match on and a transport-level retry filed a second issue. The key is now derived from the call itself — admin, machine, every content field, and a 10-minute window.

The window is what keeps this honest in both directions. A retry moments later is the same report; the same fault re-reported weeks later (a repair that regressed) must open a **new** issue rather than silently resolving to the closed one. A retry straddling a window boundary degrades to the old behaviour — a duplicate — which is the safe way to fail.

**`list_machines` answered counting questions wrongly and confidently.** `count` was the length of the page, so a 50-row page of a 120-machine collection read as "there are 50" with no signal it was truncated (CORE-ARCH-012). It now reports `total` and `hasMore` beside `count`, and the tool description tells the model to narrow the filter or raise the limit before answering a counting question.

## The thing only a real database could tell us

`issues.idempotency_key` is a Postgres `uuid` column — which is *why* the original code reached for `randomUUID()`. The first attempt passed a hex digest and failed at query time with `invalid input syntax for type uuid`. The digest is now rendered as a **UUIDv8**, the RFC 9562 version reserved for custom layouts: v4 would misrepresent these bits as random, and v5 specifically means SHA-1 over a namespace.

This surfaced only because the work was probed against a live local MCP endpoint. Unit tests alone would have sailed past it.

## Findings closed as non-defects

- **`after()` is valid inside an mcp-handler tool context.** Probing `create_issue` on local dev returned ok *and* delivered real notification emails to Mailpit (2 → 4), so the mutation commits and post-commit dispatch runs. No fix needed.
- **Audience binding** is moot after the bearer-token pivot — there is no third-party issuer to confuse — and the **cached claims client** it referenced was deleted with the OAuth code in #1727.

## Testing

- 6 new unit tests on the key builder: v8 UUID shape, stability, per-field and per-actor divergence, window behaviour, and NUL-join boundary safety.
- 4 new integration tests: truncated vs complete pages, retry-returns-original, different-content-files-new. `mcp-tools.test.ts` goes 15 → 19.
- `pnpm run check` green.
- Live-verified on a local dev MCP endpoint: identical call twice → same issue number, one row; different title → new issue. `list_machines` → `{count: 3, total: 10, hasMore: true}`.

**`pnpm run preflight` did not run locally** — the memory precheck hard-blocked it (8.5 GB swap in use from concurrent sessions, over the 6.5 GB threshold), and forcing past that gate on a 16 GB machine risks a lockup. CI runs the full suite.

Bead: PP-u4ab.4. PP-u4ab.3 (rate limiting) stays open — separate concern, deliberately not bundled.

### DIFF
--- a/src/lib/mcp/tools/create-issue.test.ts
+++ b/src/lib/mcp/tools/create-issue.test.ts
@@ -13,7 +13,14 @@ import { createIssueIdempotencyKey } from "./create-issue";
 
 const USER = "11111111-1111-4111-8111-111111111111";
 const MACHINE = "22222222-2222-4222-8222-222222222222";
-const NOW = 1_770_000_000_000;
+/**
+ * Deliberately *mid*-bucket, not on a boundary. `1_770_000_000_000` is an exact
+ * multiple of the 10-minute window, so a test anchored there always sits at the
+ * start of a bucket and the "a retry seconds later collides" assertion holds
+ * trivially — it would keep passing even if the near-boundary protection
+ * regressed to nothing.
+ */
+const NOW = 1_770_000_000_000 + 100_000;
 
 const base = {
   machine: "AFM",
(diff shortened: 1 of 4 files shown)


timothyfroehlich:
Claude review of head bafd41e — code-review medium + low; 4 findings from the medium pass all fixed in bafd41e8; low pass on head clean

PR #1803, opened 2026-08-02
fix(pinballmap): stop a form post claiming a PinballMap listing (PP-o355.29)

timothyfroehlich:
## Summary

- **`pinballmapListed` was settable from a request body.** It records that a listing exists on PinballMap's public map, but `readPbmLinkFormFields` read it straight off the FormData and both the create and update schemas accepted it. A POST could therefore make PinPoint record a machine as listed on the public map with a null lmx and **no PBM write ever made** — asserting something about a public map we never did (CORE-ARCH-012). It also consumed the one-lister slot in `machines_pinballmap_listed_unique`, so the cabinet that genuinely holds the lmx could no longer be listed. Not reachable through the rendered UI — no form draws the control — but a server action is POST-addressable regardless of what the UI draws.
- **Fix:** drop the field from `createMachineSchema` / `updateMachineSchema` and from `readPbmLinkFormFields`. It is now written only by paths that actually talk to PinballMap (`linkPinballmapEntryAction`, the verify/heal transaction), plus the edit carry-over, which reads it from the **stored row**.
- **Three comments asserting the field "is not a form field" were wrong.** They are now true, and are written as an enforced invariant rather than an observation.

## Follow-on cleanup (worth a look — it touches PP-o355.15, merged yesterday)

With create unable to list, `machines_pinballmap_listed_unique` became unreachable from `createMachineAction`: a row with `pinballmap_listed = false` is not in the partial index at all, so create can only ever raise the *initials* 23505. The constraint-name disambiguation added in #1798 had no remaining caller, so:

- `isPbmListingConflict` is **removed**; the two create catches go back to a bare `23505` = initials.
- `updateMachineAction` **keeps** its catch. The carry-over rewrites a value its own row already holds, so it cannot collide with itself today — but PP-o355.20 runs auto-link at title-save time, which makes it genuinely reachable again. Commented as a backstop rather than a live path.
- `getPostgresErrorConstraint` stays as general infrastructure beside `isPgErrorCode` (the postgres-js / PGlite spelling split is expensive to rediscover), explicitly labelled as having no current production caller.

## Test Plan

- [x] The two tests that provoked the collision by POSTing `pinballmapListed=on` are **replaced, not repaired** — re-adding the field to make them pass would reopen the bug. The bead called this out in advance.
- [x] Collision coverage moved to `linkPinballmapEntryAction`, the one path that still reaches the index; also asserts the refused cabinet stays unlisted with a null lmx (no half-apply).
- [x] Two new regressions: a submitted `pinballmapListed=on` is ignored on create and on update. Deliberately seeded with **no incumbent**, so they fail on the flag itself rather than incidentally colliding.
- [x] Both regressions **red-verified** against a temporarily reopened hole — they fail on `expect(row.pinballmapListed).toBe(false)` receiving `true`, which is the bug exactly.
- [x] `pnpm run check` green (2024 unit tests, types, lint, format, ruff, shellcheck, actionlint).
- [x] PBM integration suites green (21 tests across listing-conflict, link-capture, reconcile).
- [ ] `pnpm run preflight` **not run** — the memory-pressure gate hard-blocked it twice (7.6 GB swap, parallel sessions on the 16 GB Mac) and I did not force past it. CI runs build + the full suite.

No UI changes — no screenshots.

## Related Issues

PP-o355.29. Unblocks PP-o355.21 (listing controls), which must not be built on a write path that can be spoofed.


timothyfroehlich:
Claude review of head 819e0e5 — One finding: updateMachineAction's 23505 catch shipped uncovered after its test was deleted. Probed the suggested fix — the seed it needs is the state the unique index forbids, so the branch was unreachable, not just untested. Removed it; PP-o355.20 re-adds it with the test auto-link makes possible.

PR #1804, opened 2026-08-02
docs(agents): cut the TypeScript corpus to its one decision (PP-22e4)

timothyfroehlich:
Sweep PR 2 of PP-22e4. **−653 / +48.**

Deletes `docs/TYPESCRIPT_STRICTEST_PATTERNS.md` outright and cuts `.agents/skills/pinpoint-typescript` from 334 lines to 42.

## What went, and why it was safe

The patterns doc was 318 lines of generic TypeScript/Drizzle/Supabase examples with no project-specific choice in it — and stale besides: it declared severity as `"minor" | "playable" | "unplayable"` (the real enum is `cosmetic | minor | major | unplayable`), typed Next 15 `params` as non-Promise, and queried a `users` table that is called `userProfiles`.

From the skill, three candidates were checked individually and confirmed **not** losses:

- **Date-helper null narrowing** — the helpers take `Date | string | number` (`src/lib/dates.ts:75,117,125`), so a nullable argument is already a compile error.
- **FormData boolean coercion** — the rationale lives in a doc comment at `src/app/(app)/report/validation.ts:20-30`; the deleted line was a pointer to it.
- **Zod schema colocation** — covered by `CORE-TS-005`.

## What survived, and got stronger

The one decision: **there is no db→app converter layer and none should be built.** It now carries the history that makes it stick, which previously existed nowhere.

PinPoint ran exactly that layer until **#480**: a type-level `DrizzleToCamelCase<T>` over essentially every response type, plus a runtime `transformKeysToCamelCase` invoked **76 times across 23 files** — the DAL, every tRPC router, and a family of dedicated `*-response-transformers.ts` modules — which forced an `as` assertion at nearly every one — the one currency ts-strictest exists to refuse. It existed only because the **v1 schema omitted column names** (`email_verified: timestamp()`); the v2 schema declares both (`firstName: text("first_name")`), and that single authoring choice is what made the layer unnecessary.

## CORE-TS-003 corrected

Its `Do` said *"Keep snake_case in schema, convert to camelCase at boundaries."* That wording was introduced in `36c13acc` — **the very commit that deleted those converters** — and was never revised, so it described the v1 codebase for nine months.

Conversion is correct only where Drizzle's mapping does not run: `db.execute()` and raw SQL, Supabase RPC (`src/lib/discord/config.ts`), auth metadata (`src/lib/auth/profile.ts`), and external APIs (`src/lib/pinballmap/parse.ts`). Those targeted parsers are legitimate and stay.

## Deviation from the plan

The plan said delete the skill entirely, justified by `.github/instructions/typescript.instructions.md` remaining as the routing fallback. **#1801 deleted that file.** Deleting both would leave TypeScript guidance only in `NON_NEGOTIABLES.md`, which nothing auto-loads — and the failure mode here is sessions re-deriving the wrong model, which a rule nobody loads cannot prevent. So the skill is shrunk, not deleted. Approved explicitly before implementing.

`REVIEW.md`, `README.md`, and `docs/DEVELOPMENT.md` all still route to a skill that exists.

## Verification

- **242** explicitly-named columns audited; every TS key is exactly `camelCase(sql_name)`, zero divergences. No column omits its name string.
- `drizzle-orm` 0.45.2, no `casing` option set. Mechanism confirmed at source (`drizzle-orm/utils.cjs:87-101` keys results by TS property, never the SQL name) and empirically against live PGlite.
- `check_rule_ids.py` passes; `pnpm run check` green (219 files, 2024 vitest, 488 pytest).
- Full history and evidence recorded in **PP-7ulb**.

## Review findings already applied

A cold-read review caught **four overclaims in the added prose** — all narrowed before this PR opened:

1. Claimed strictest compliance is purely a compiler concern. It isn't: `any` is an ESLint error, and `!` is caught by nothing.
2. Claimed strictest applies generally. `tsconfig.tests.json` extends only `tsconfig.base.json`, so `exactOptionalPropertyTypes`/`noUncheckedIndexedAccess` are off under `src/**/*.test.*`.
3. "Ran on every read / at each call site" overstated the runtime transform's 9 sites (one of which carried no `as`).
4. "ts-strictest makes a missed case a compiler error" is true of the cited exemplar, not of TypeScript generally.

That review also surfaced **PP-8k07**: `CORE-TS-007` is Critical severity, and its `!` ban is enforced by neither the compiler nor ESLint (`@typescript-eslint/no-non-null-assertion` is never enabled — verified via `eslint --print-config`). Out of scope here.

## Note for the reviewer

Unlike #1799, this PR **invents prose** rather than only deleting and moving verbatim — the shrunk skill and the corrected rule are new text. That's the higher-risk part of the diff and where a second pair of eyes is most useful.

Dead link left deliberately: `docs/testing/e2e-audit-2026-05.md:160` references the deleted doc, but it is a frozen dated record (AGENTS.md §8).

---

## Review rounds applied

Two `/code-review` passes on this branch, ten findings total, all verified against the repo before applying.

**Round 1 (6):** restored the `exactOptionalPropertyTypes` idiom that #1801 orphaned; corrected the call-site count (I had written "nine across the DAL" from a reviewer that only searched `src/lib/dal`); switched `CORE-TS-001…008` to `CORE-TS-001..008` so `check_rule_ids.py` actually validates the range rather than just its endpoints; pointed at the `~/lib/types` barrel CORE-TS-001 mandates; named all four globs `tsconfig.tests.json` owns; added supabase-js PostgREST paths to the conversion list.

**Round 2 (4):** the prescribed conditional spread guarded on truthiness, dropping `""` and `0` — corrected to `!== undefined`; withdrew a false claim that widening a property to `| undefined` violates CORE-TS-007; added `.from()` queries to CORE-TS-003's `Do`, which lagged the skill on the always-loaded surface; restored CORE-TS-004's content to Quick Start item 2.


timothyfroehlich:
Claude review of head 768417b — `/code-review medium` — 3 low findings: supabase-js writes need conversion too; barrel-gap list omitted NotificationPreference; barrel bypass conflicted with CORE-TS-001 — all fixed

PR #1805, opened 2026-08-02
feat(workflow): compute the PR merge handoff instead of narrating it (PP-9onv)

timothyfroehlich:
## What

`scripts/workflow/merge-handoff.sh <PR>` — the block an agent pastes when handing a PR to Tim to merge. Every field is computed from `git` + `gh` rather than written from memory.

```
PR #1803 — fix(pinballmap): stop a form post claiming a PinballMap listing (PP-o355.29)
  fix/pbm-listed-not-form-field-PP-o355.29 -> main · bead PP-o355.29 · <url>

  review        NONE — no review marker on this PR
  ci            CI Gate conclusion=SUCCESS
  threads       0 unresolved review threads
  mergeable     MERGEABLE · 0 commit(s) behind main
  main merged   2026-08-02T15:48:24-05:00 (23 minutes ago) — 1 commit(s) ago

  diff vs main  +207 -138   8 file(s)   src +86 -108 · tests +121 -30   (3 commit(s))
  since review  n/a — nothing reviewed to compare against
  migrations    none
  new env vars  none
  ui            yes · NO screenshots posted

  ! bash scripts/workflow/merge-handoff.sh 1803   (re-run — this report is a snapshot)

  NOT MERGEABLE YET — 1 gate(s) blocking:
    - reviewed: unreviewed — Tim runs /code-review, then the agent attests
```

## Why

The handoff was prose — "CI is green, you reviewed it a couple of commits back, mostly docs." Each of those is derivable, and each is a claim an agent can get wrong without noticing. The two `!`-prefixed lines are the point of the format: the block is a snapshot, so Tim re-runs it himself instead of asking whether it still holds.

**The merge command prints only when all four gates pass.** An un-ready PR gets the blocking reasons instead — a merge command handed over while CI is yellow invites a merge on a guess, and the gates would refuse it anyway.

## Notable

- **`mark-claude-review.sh` now takes a required `<depth>`** (`low|medium|high|xhigh|max|ultra|trivial`). Which review ran was never recorded, so the report could not state it without an agent recalling it. The depth lives in its own HTML comment — the `reviewed` gate compares the SHA marker's contents to head by string equality, so anything added *inside* it would fail every gate. Pre-existing markers read as `unrecorded`, not as "no review".
- **`_marker_verdict` is now the first two fields of `_marker_record`** — one implementation, so the gate and the report cannot disagree about whether head was reviewed.
- **Reads both sides off `FETCH_HEAD`**, updating no ref, so it is safe to run from any worktree while other agents hold theirs.
- Env-var detection subtracts names on removed lines, so editing a registry line doesn't report existing vars as new.

## Testing

- `scripts/tests/test_merge_handoff.py` — 25 tests driving the real bash against a real temp git repo + stubbed `gh`: gate-blocking, bucket classification (incl. `src/**/*.test.ts` → tests, and the `other` bucket that stops a workflow edit hiding inside "mostly docs"), migrations, env vars, screenshots, review distance (incl. the force-push case where the distance is unknowable), merge-from-main distance, bead extraction.
- `scripts/tests/test_pr_gates.py` — 4 new tests for `_marker_record`, including backward compat and the record/verdict agreement.
- 490 pytest pass; `pnpm run check` clean. Smoke-run against live PRs #1802 (merged, old-format marker), #1803, #1804.

Closes PP-9onv.


timothyfroehlich:
Claude review of head 96b7565 — `/code-review low` — no findings; note a low pass skips test hunks, and most of this diff is tests

PR #1806, opened 2026-08-02
feat(mcp): rename a machine over MCP (PP-u4ab.10)

timothyfroehlich:
Adds `set_machine_name` to the MCP tool catalog (bead PP-u4ab.10). A machine's display name is the label a fleet pass most often finds wrong — `Elvira's House of Horrors` that should read `Elvira's House of Horrors (Premium)` — and until now the only way to fix one was the web edit form, so a floor pass that corrected 100+ catalog links still left 100+ labels to hand-edit at a desk.

## What it does

`set_machine_name(machine, name)` — resolve the machine by initials (case-insensitive) or UUID like every other tool, gate on `machines.edit` with the same ownership context the edit form's name field uses, then rename inside a transaction that also emits the existing `name_changed` lifecycle event.

- **Exactly one timeline event per rename**, carrying `from` and `to`. No new event variant — `emitMachineUpdated` already emits `name_changed` when the name differs and no-ops when it doesn't.
- **Renaming to the current name writes nothing** and reports `changed: false` rather than claiming a change that did not happen (CORE-ARCH-012). Same shape as `set_machine_availability`.
- **Initials are untouchable.** They are the FK target for `issues.machineInitials` and the `/m/<initials>` URL segment, so changing them is a cascade plus a link break, not a rename — the input schema has no initials field and the service writes only the `name` column. Wrong initials get filed separately.

## Shape

`updateMachineName` sits next to `updateMachinePresence` in `~/services/machines` and mirrors it exactly: same params object, same pre-loaded `current` snapshot, same `{ changed }` return, same no-op guard. The bead said "no new service seam"; I read that as "the `name_changed` emit already exists, don't invent one" — putting the transaction in the tool instead would have made it the first MCP tool to write to the DB directly, against the layering the services header comment spells out (authorization in the caller, transaction in the service). Happy to move it if that reading is wrong.

Note: `src/lib/mcp/tools/index.ts` still says "four mutations" in its doc comment — three tool PRs are in flight against that file at once, so I kept my change there to the single registration line rather than churn a comment every one of them touches. Worth one cleanup pass once they've all landed.

## Testing

`pnpm run check` clean. Six new cases in `src/test/integration/mcp-tools.test.ts` (PGlite, worker-scoped, direct `run*` invocation like the rest of the file): rename writes the row + exactly one `name_changed` event attributed to the caller and leaves initials alone; no-op reports `changed: false` and writes no event; a non-owning member is denied and the name is unchanged; the machine's own owner can rename; unknown machine throws `not_found`. 25/25 pass in that file.

### DIFF
--- a/src/lib/mcp/tools/index.ts
+++ b/src/lib/mcp/tools/index.ts
@@ -7,6 +7,7 @@ import { registerCreateIssue } from "./create-issue";
 import { registerGetMachine } from "./get-machine";
 import { registerListMachines } from "./list-machines";
 import { registerSetMachineAvailability } from "./set-machine-availability";
+import { registerSetMachineName } from "./set-machine-name";
 import { registerSetMachineOwner } from "./set-machine-owner";
 
 /**
@@ -18,6 +19,7 @@ export function registerPinpointTools(server: McpServer): void {
   registerListMachines(server);
   registerGetMachine(server);
   registerSetMachineAvailability(server);
+  registerSetMachineName(server);
   registerAddMachine(server);
   registerSetMachineOwner(server);
   registerCreateIssue(server);
--- a/src/lib/mcp/tools/set-machine-name.ts
+++ b/src/lib/mcp/tools/set-machine-name.ts
@@ -1,0 +1,92 @@
+import "server-only";
+
+import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import { z } from "zod";
+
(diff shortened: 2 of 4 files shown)


timothyfroehlich:
Claude review of head 88995ee — Clean review: one name_changed event, permission gate mirrors updateMachineAction, transaction-safe, no denormalized name to desync. Two minor non-blocking notes (stale tool-count comment → PP-x8jb; pre-existing concurrent-rename pattern).

PR #1807, opened 2026-08-02
feat(mcp): filter list_machines by PinballMap link state (PP-u4ab.9)

timothyfroehlich:
## What

Adds a `pinballmap` filter to the `list_machines` MCP tool: `unlinked` | `linked` | `excluded`.

## Why

The fleet linking pass (PP-h059) is a bookkeeping sweep over 100+ machines, executed by Claude over MCP. It runs on one question — *which machines still have no PinballMap catalog match?* — and `list_machines` could filter by name and presence but not by PinballMap state. Without this, the pass has to page the whole fleet and call `get_machine` on every row to find out: ~100+ extra round trips per sweep, and resuming a half-finished pass is guesswork.

## The load-bearing detail

`unlinked` means **no `pinballmapMachineId` AND not `pinballmapExcluded`**. A machine someone deliberately marked as not on PinballMap is *finished work*, not a worklist item — folding it in would make the pass re-examine the same rows on every sweep and never reach empty. A DB CHECK (`machines_pinballmap_link_exclusive`) already forbids a row that is both linked and excluded, so the three states partition the fleet exactly.

## Implementation notes

- The conditions live in a `Record` keyed by the filter union rather than an if/else chain, so adding a state to the enum without a condition is a **type error**. A missed branch would silently return the whole fleet under a narrowing filter name, with a `total` that reads as authoritative (CORE-ARCH-012).
- The filter joins the existing `conditions` array, which the page query and the `count()` query already share — a filter applied to only one of them would report a `total` that paging can never exhaust.
- The tool description tells the model when to reach for `pinballmap: 'unlinked'`.

## Tests

Four integration tests (PGlite, direct `runListMachines`) in the existing `list_machines` block:

- `unlinked` skips excluded machines (the whole point).
- `linked` / `excluded` return their own buckets.
- Composes with `search` + `presence` (three decoys, each failing exactly one filter).
- **110-machine fleet**: pages the filtered set at `limit: 25` and asserts the union of the pages is exactly the seeded unlinked set — no repeats, no gaps, and `total` matches what paging actually reaches.

## Verification

- `pnpm run check` — pass
- `pnpm run test` — 2031 pass
- `vitest --project integration src/test/integration/mcp-tools.test.ts` — 24 pass

Scope is confined to `src/lib/mcp/tools/list-machines.ts`; `list_machines` is already registered, so no registration change.


src/lib/mcp/tools/list-machines.ts
```diff
@@ -131,7 +204,7 @@ export function registerListMachines(server: McpServer): void {
...
...ballMap catalog match. Returns 'count' (this page), 'total' (every match), 'offset', and 'hasMore'. Answer counting questions from 'total', never from 'count' or the array length. There are two ways to get through a collection larger than one page, and picking the wrong one silently skips machines. (A) READING only, nothing changing: keep requesting with offset += limit until hasMore is false (raising limit alone caps at 100 and will not reach the rest). (B) ACTING on what you find, in a way that changes whether it still matches — linking machines while paging pinballmap:'unlinked' is the case this tool is built for: do NOT advance the offset per page. Each machine you link drops out of the filter and the rest shift up, so offset += limit skips as many machines as you just fixed. Re-request offset 0 and let the worklist drain. When you leave a machine as it is — you cannot link it, or you are deferring the decision — raise offset by exactly that many so it does not come back on the next request. You are done when a request comes back EMPTY (count 0), NOT when total reaches 0: every machine you deliberately left keeps total above 0 forever, so waiting for 0 is a loop with no exit.",
```
timothyfroehlich: **Low** — Procedure (B) describes an action the MCP surface cannot perform.

The description tells the model that "linking machines while paging `pinballmap: 'unlinked'` is the case this tool is built for" and to re-request offset 0 while the worklist drains. But there is no MCP tool that links or excludes an *existing* machine — `add_machine` accepts `pinballmapMachineId` / `pinballmapExcluded` only at creation time, and the other tools (`get_machine`, `set_machine_availability`, `set_machine_name`, `set_machine_owner`, `create_issue`) do not touch link state.

Concrete misbehavior: a model told to run the fleet linking pass over MCP follows (B), gets a page, finds it has no way to link anything, and either re-requests offset 0 with identical results (a no-progress loop) or has to fall back to the "leave the machine as it is" clause for every row — i.e. procedure (A) with extra steps. Until the link/exclude tool lands, the description promises a capability that is not there.

Not blocking if the write tool is a following bead; worth a sentence saying so, or trimming (B) until it is real.

timothyfroehlich: Confirmed and fixed in `bd51ccf1`. I verified the capability claim before acting on it: no registered MCP tool writes `pinballmapMachineId`/`pinballmapExcluded` on an existing machine — `add_machine` accepts them only at creation. I also checked `auto-link` (PP-o355.20), which landed in `main` while I was working on this, in case it closed the gap: it doesn't. It writes `pinballmapListed` and skips any row with a null `pinballmapMachineId`, so it can't move a machine out of the `unlinked` bucket either. Only a human in the app can.

**I chose to trim (B) rather than mark it forward-looking**, which was the option you leaned away from, so here's the reasoning.

A tool description is an instruction to a model that will act on it, not documentation a human reads and filters. "Do this, but not yet" is still stated in the imperative — a weaker guard than not stating it at all, and it dilutes the part of the description that *is* live. That's the lesson of the previous three rounds on this file specifically: the model treats this text as authoritative, which is exactly why it kept being worth tightening.

Your objection to trimming was losing the reasoning, and I think that only holds if the reasoning has nowhere else to live. It does. The full procedure — the drain, the leave-alone offset arithmetic, and the empty-page stop condition from rounds 2–4 — is now a block comment above `registerListMachines`, tagged with what to restore and when. That's where whoever wires up PP-u4ab.12 will actually be looking; nobody implementing a write verb reads a tool description for instructions about a capability they're in the middle of adding. Same words, better audience.

Bonus from trimming: what's left is unconditionally true rather than true-with-a-caveat. There's one enumeration procedure now, and the `offset` param can state the exactly-once guarantee flatly instead of hedging it. I kept one honest sentence that a person editing machines in the app can shift the set under a sweep, since that's the one mutation path that does exist today.

—Claude

timothyfroehlich: Correcting my own reply above, since the record shouldn't stand as written. Fixed in `4079a4ed`.

Trimming (B) was right for the reason I gave. What I got wrong was the sentence I put in its place — "nothing you can call from here changes whether a machine matches a filter" — which is a *stronger* claim over *more* filters than the one I removed, and false for two of them. `set_machine_availability` writes `presenceStatus`; `set_machine_name` writes `name`, which is both the `search` target and the primary sort key. Both are registered today.

So "put every off-the-floor machine back on the floor" — page `presence: "off_the_floor"`, act on each row, advance the offset — skips roughly half the fleet and ends on `hasMore: false` looking complete. I deleted a warning about a workflow that couldn't happen and replaced it with a promise that's false for two workflows that can. The correct move was to generalize (B), not delete it.

The drain procedure is back in the description, now stated for any filter rather than for linking, with `add_machine` named as well (its link columns are all optional, so it can insert straight into `unlinked` mid-sweep — that re-reads a machine rather than skipping one). `pinballmap` is documented as the accidental exception it currently is, and the comment notes the procedure covers it unchanged once .12 makes link state mutable.

Two claims in my reply above are now void: "what's left is unconditionally true" and the `offset` param stating the exactly-once guarantee flatly. That param was the same overclaim one field over, and it now says only what the total order actually buys — separate requests agree on where a page boundary falls, for as long as the rows don't change.

—Claude

timothyfroehlich:
/audit-override Four freshly-published high advisories in transitive deps, none attributable to this PR: undici (GHSA-4cwx-7wf7-3272), fast-uri (GHSA-7p8r-x3mc-p8w7), ip-address (GHSA-mwp4-54f8-5fhr), brace-expansion (GHSA-rgw5-rvv9-x895). This branch's dependency graph is byte-identical to main — `git diff origin/main...HEAD -- package.json pnpm-lock.yaml` is empty and the whole diff is source files under src/lib/mcp/. Main is not red only because its last audit job was path-filtered to skipped. Authorized by Tim 2026-08-03; a dependency-bump PR is the real fix and is being tracked separately.

timothyfroehlich:
Claude review of head 4079a4e — `/code-review medium` — Paging guidance still overclaimed after round 5; fixed in 4079a4ed by stating the mutate-while-paging hazard as live rather than hypothetical. Threads resolved.

PR #1808, opened 2026-08-02
fix(lint): enforce CORE-TS-007's non-null assertion ban (PP-8k07)

timothyfroehlich:
CORE-TS-007 is **Critical** and bans `any`, non-null `!`, and unsafe `as`. Only the `any` third had a gate.

`@typescript-eslint/no-non-null-assertion` ships in typescript-eslint's `strict` config, and `eslint.config.mjs` spreads only `recommended-type-checked` + `stylistic-type-checked` — so the rule was never set. `tsc` was never going to catch it either: `!` is valid TypeScript at every strictness setting, which is the entire point of the operator. An agent could write `getUser()!.email` to silence a strict-null error and have both `pnpm run check` and CI Gate go green, then 500 the page at runtime — the exact class the rule exists to prevent.

## What changed

Rule enabled at `"error"`. It found **32 assertions: 3 in production, 29 in tests.**

The three production sites, rewritten rather than suppressed:

| File | Fix |
| :--- | :--- |
| `src/components/issues/IssueFilters.tsx` | Capture `filters.machine` in a local so the `clear` closures reference the array directly — TS can't retain the `?.forEach` narrowing inside a callback, which is why the `!` was there. |
| `src/lib/export/csv.ts` | Read `field[0]` once and test it for `undefined`; that subsumes the `field.length > 0` guard. (`charAt` is not an equivalent rewrite — `"=+-@".includes("")` is `true`.) |
| `src/lib/logger.ts` | Destructure the first stream and branch on it, so the single-stream path narrows instead of asserting. |

**Off in the test and e2e blocks.** In production a wrong assertion 500s a page in front of a user; in a test the same wrong assertion throws inside the test and fails it loudly — which is the honest failure the rule exists to force. All 29 are assertions on fixtures the test itself just created (`targetUser!.id`, `clearButton!`). They stay as they are.

## Mirror

The rule is **syntactic**, so typescript-eslint's `disable-type-checked` leaves it on and the slim local pass enforces it — no `.oxlintrc.json` entry needed, and no double-reporting. Verified rather than assumed: a seeded `x!.length` is reported by both `pnpm exec eslint` and `PINPOINT_LINT_SLIM=1 pnpm exec eslint`.

## Docs

The CORE-TS-007 catalog entry gains an **Enforced by** line, including the honest part: the **unsafe `as`** third still has no gate and is review-enforced. `no-unsafe-*` targets untyped values, not casts between known types.

## Verification

- `pnpm run check` — green
- `pnpm run lint` (authoritative, full ESLint over `src/ e2e/ scripts/`) — exit 0
- `pnpm run test` — 2031 passed / 220 files

Closes PP-8k07.

https://claude.ai/code/session_019SBvbjTjNcS4CoarvpVdhT


timothyfroehlich:
Claude review of head d448eb5 — `/code-review medium` — one finding: no-non-null-assertion was missing from the no-restricted-disable list, so the new rule was bypassable by a one-line disable comment — fixed in 26ba023f

PR #1809, opened 2026-08-02
feat(mcp): read a machine's PinballMap title/edition (PP-u4ab.8)

timothyfroehlich:
## What

The **read half** of the MCP walk-the-floor flow (PP-u4ab.8). Claude can now answer "is our Elvira's House of Horrors recorded as the Premium?" over MCP. The write half is PP-u4ab.12, sequenced behind this so it can echo the shape defined here.

Two gaps closed:

**1. `search_pinballmap_catalog` (new tool)** — a thin read-only wrapper over `~/lib/pinballmap/catalog`, mirroring the web linking picker's two steps:

- `query` → matching *families* (a PBM edition group, or a standalone title), each with an `editionCount`
- `machineGroupId` → that family's individual *editions* (Pro/Premium/LE) with each one's `pinballmapMachineId`

Exactly one of the two. Passing both — or neither — is an `invalid` error rather than a silently ignored argument, so the caller always knows which of the two things ran. `hasMore` is exact (the search over-fetches by one) rather than inferred from a full page, the miscount PP-u4ab.4 fixed on `list_machines`. No offset: this is disambiguation, not enumeration, and the tool description says to narrow the query instead.

**2. `get_machine` gains a `pinballmap` block** — the machine's stored PBM columns plus the linked catalog entry.

## The shape

`McpMachinePinballmap` (exported from `src/lib/mcp/tools/pinballmap-block.ts`) is the deliverable PP-u4ab.12 reuses, so the three states are explicit rather than implied by an absent field:

- `status: "linked"` — `pinballmapMachineId`, catalog `title` (the edition name), `machineGroupId` + `group` (the family), `manufacturer` / `year` / `opdbId` / `ipdbId` as PinPoint stored them, plus `listed` and `lmxId`
- `status: "excluded"` — deliberately not on Pinball Map, with the operator's `reason`
- the containing field is `null` when **neither** — nothing recorded yet, which is a different fact from "excluded"

The field is always present. `title`/`group` are `null` when the mirror no longer holds the id — a stale link still reports as linked, because the stored id is real; the mirror just can't name it.

## MachineRef widening

`resolveMachine` now selects the PBM columns, and `MachineRef` extends the existing `MachinePbmColumns` (the same set the create/edit paths write) rather than redeclaring them. Every tool shares that resolver; all of them still typecheck, and the full unit + integration suites pass. It also gives PP-u4ab.12 what it needs to carry `pinballmapListed` / `pinballmapLmxId` over from the **stored** row rather than from its arguments (PP-o355.29).

## Testing

Nine new cases in `src/test/integration/mcp-tools.test.ts` (PGlite, worker-scoped): the family/edition two-step against a seeded Elvira family, `hasMore` truncation, an unknown group id, both/neither argument rejection, and the four `pinballmap` block states (linked, stale link, excluded, neither).

`pnpm run check` clean; unit (2031) and the PBM + machines integration suites green.

## CORE-PBM-001

Every read is served from the local `pinballmap_catalog` mirror. No code path and no test reaches pinballmap.com — the tests seed the mirror table directly.

## Out of scope

No writes: no linking, unlinking, excluding, listing or renaming (PP-u4ab.12 / PP-u4ab.10).


src/lib/mcp/tools/pinballmap-block.ts
```diff
@@ -0,0 +1,107 @@
...
+ * Returns `null` when the machine is neither linked nor excluded.
+ *
+ * Reads the `pinballmap_catalog` mirror only — never pinballmap.com
+ * (CORE-PBM-001).
+ */
+export async function buildMachinePinballmap(
+  machine: MachinePbmColumns
+): Promise<McpMachinePinballmap | null> {
+  if (machine.pinballmapMachineId !== null) {
+    const entry = await getCatalogEntry(machine.pinballmapMachineId);
```
timothyfroehlich, line 108: **Empty catalog mirror is indistinguishable from a stale link.**

`buildMachinePinballmap` has no equivalent of the `assertCatalogPopulated()` probe the sibling tool in this same PR added. When `pinballmap_catalog` has zero rows — a fresh preview branch, a local DB seeded from a prod dump, or a run of failed weekly `refreshCatalog` runs, exactly the states `assertCatalogPopulated` exists for — `getCatalogEntry` returns `null` for *every* linked machine, so `get_machine` reports `{status: "linked", pinballmapMachineId: 70012, title: null, machineGroupId: null, group: null}`.

That is precisely the signature `McpMachinePinballmapLinked.title` documents as "the local catalog mirror no longer holds the id, which means the stored link is stale". So the walk-the-floor question this bead exists for — "is our Elvira recorded as the Premium?" — gets answered "linked, but the mirror can't name it / the link looks stale" as fact, and a model may go on to re-link a link that was never broken. `search_pinballmap_catalog` explicitly refuses to make that class of claim; this path makes it silently.

Suggest reusing `isCatalogEmpty()` when `entry` is null and distinguishing the two, e.g. an always-present `titleUnavailable: "catalog_empty" | "not_in_catalog" | null` alongside `title`, so both tools tell the same story about an unpopulated mirror.

—Claude

timothyfroehlich: Fixed in a40a4f08. You're right that this is the same empty-mirror class as the search tool, and that this path got missed — reused `isCatalogEmpty()` rather than a second mechanism, probed only on a miss so a resolved link stays one query.

**Reports a state rather than throwing.** The search tool throws because an empty mirror leaves it with nothing to return; `get_machine` still has a real answer (name, presence, owner, open issues), so failing the whole call over the PBM block would take out an otherwise fine request. The linked variant now carries `catalogLookup`:

- `found` — `title`/`group`/`machineGroupId` are real
- `missing` — mirror has rows but not this id; the link is genuinely stale
- `mirror_unpopulated` — mirror is empty; the link is fine, its title just unknown

Worth flagging what I found while testing it: the existing "stale link" test seeded no catalog at all, so it was itself exercising the empty-mirror path while asserting stale-shaped output — the bug was pinned as correct behaviour. It now seeds the mirror (making "stale" a fact), and a new test covers a linked machine against an empty catalog.

**On the `limit` note — leaving it, deliberately.** It's a real asymmetry with this PR's own principle, but the two cases differ in what silence costs. Ignoring `query` vs `machineGroupId` hides *which operation ran*; ignoring `limit` can't misreport anything — the editions branch is unpaged, `returned` is exact, and the param description says it's ignored there. Erroring would reject an otherwise-correct call over an inert argument, and actually *applying* it would be worse: truncating a family's editions hides the very options the caller is choosing between. So the choice is between silent-and-harmless and loud-and-hostile, and I'd rather not fail a correct lookup to satisfy a symmetry.

—Claude

src/lib/mcp/tools/search-pinballmap-catalog.ts
```diff
@@ -0,0 +1,250 @@
...
...s House of Horrors', or a standalone title) with an `editionCount`; then, ONLY for a family with a non-null `machineGroupId` and `editionCount` above 1, pass that `machineGroupId` to list its individual editions (Pro/Premium/LE) with the `pinballmapMachineId` of each. A standalone title — `machineGroupId` null, or `editionCount` 1 — already carries its `pinballmapMachineId`: that is the answer, don't call again. Most pre-1990s machines are standalone. Pass one argument or the other, never both. CHECK `familyName` ON THE EDITIONS RESPONSE before using any id from it: machine ids and machine-group ids are separate id spaces, so the wrong integer can return a real but unrelated family's editions. If `familyName` is not the title you searched for, or `idAlsoMatchesEdition` is non-null, you passed an edition's `pinballmapMachineId` instead of a family's `machineGroupId` — `idAlsoMatchesEdition` names the title you actually meant. `returned` is this page's size, not a total — this tool cannot answer 'how many titles are there'; when `hasMore` is true, narrow the query, as families come back best-match first and there is no paging. Read-only, served from PinPoint's local catalog mirror.",
```
timothyfroehlich: **`idAlsoMatchesEdition` non-null is the normal case, but the description treats it as an error signal.**

PBM machine-group ids and catalog machine ids are separate id spaces, but they *overlap numerically*: group ids are small integers (this repo's own `machine-groups.json` fixture spans 8–137) while catalog machine ids run from 1 into the ~10k range. So for essentially every real `machineGroupId`, `getCatalogEntry(machineGroupId)` finds *some* unrelated catalog row, and `idAlsoMatchesEdition` comes back non-null.

Concrete: a correct call `machineGroupId: 137` returns the right family (`familyName` matches what was searched) **and** `idAlsoMatchesEdition: { pinballmapMachineId: 137, name: "<unrelated title>" }`. This description then tells the model it "passed an edition's `pinballmapMachineId` instead of a family's `machineGroupId`" and that `idAlsoMatchesEdition` "names the title you actually meant" — so the model discards a correct edition list and can link the cabinet to the unrelated title the probe named. That's the exact mis-link the field exists to prevent, now fired on the happy path.

It also contradicts the field's own JSDoc on line ~103 ("Informational, not an error"), and fires even when the colliding edition is a member of the family being returned.

Suggestion: make `familyName` the sole verdict, and either suppress `idAlsoMatchesEdition` when the collision falls inside the returned family / `familyName` already matches, or reword this sentence so the collision is a hint to double-check rather than a conclusion that the caller was wrong.

—Claude

timothyfroehlich: Deleted in 767f8a9f. You're right, and I checked the id spaces before cutting it rather than just taking the finding.

`machine-groups.json` has 30 groups with ids 8–137. Production mirrors the **full** `/machines.json` (~10k titles from id 1 up), so those group ids sit squarely inside the machine-id range — `getCatalogEntry(machineGroupId)` finds an unrelated row for nearly every *correct* group id. The field fired on the common path, not the rare one, and the description told the model that meant it should discard the edition list it had correctly asked for. That manufactures the wrong-title link it was meant to prevent.

There's no salvageable version, so I didn't try to soften it: distinguishing "meant a group id" from "meant a machine id" is undecidable from the integer alone, and since the ranges nest, the probe has no discriminative power at all — it's noise with a confident label on it.

`familyName` is the verdict now, and it was sufficient all along: it describes what came back rather than guessing at what went in, so it holds regardless of how the id spaces overlap.

One thing worth recording — **the local fixtures hid this.** `catalog-apc.json` is location-filtered (114 rows, ids starting at 642), so no group id collides in tests, and my collision test had to hand-construct one. That's why this survived two rounds with green tests. I've written the overlap and an explicit "don't add one" onto `familyName`'s doc so the flag doesn't get reintroduced by someone reasoning from the fixture.

The collision test asserted the misfiring behaviour, so it was pinning the bug — it now asserts only that the response names the family it actually returned, which is what catches a genuine wrong id.

—Claude

timothyfroehlich:
/audit-override Four freshly-published high advisories in transitive deps, none attributable to this PR: undici (GHSA-4cwx-7wf7-3272), fast-uri (GHSA-7p8r-x3mc-p8w7), ip-address (GHSA-mwp4-54f8-5fhr), brace-expansion (GHSA-rgw5-rvv9-x895). This branch's dependency graph is byte-identical to main — `git diff origin/main...HEAD -- package.json pnpm-lock.yaml` is empty and the whole diff is source files under src/lib/mcp/. Main is not red only because its last audit job was path-filtered to skipped. Authorized by Tim 2026-08-03; a dependency-bump PR is the real fix and is being tracked separately.

timothyfroehlich:
Claude review of head 767f8a9 — `/code-review medium` — idAlsoMatchesEdition fired on correct calls (PBM group ids nest inside machine ids); deleted in 767f8a9f, familyName is the verdict. Threads resolved.

PR #1811, opened 2026-08-03
feat(review): agy becomes a gate-satisfying PR reviewer (PP-c6xz)

timothyfroehlich:
Restores an automated PR reviewer, on Tim's Google AI Plus plan via the Antigravity CLI (`agy`).

## Why

Copilot review was retired in #1801 — its free tier was too small (the 2026-08-01 allowance reset at ~02:00 UTC and was exhausted by 20:57, 42 reviews across 19 PRs, over half of them re-reviews). That left Tim's own `/code-review` as the only thing satisfying the `reviewed` gate, making every review a handoff that costs his attention. #1801 deliberately left `_compute_review_state` shaped for a second reviewer; this is that reviewer.

## What lands

```bash
./scripts/workflow/agy_review.py <PR>          # gemini-3.6-flash-high
./scripts/workflow/agy_review.py <PR> --pro    # gemini-3.1-pro-high
./scripts/workflow/agy_review.py <PR> --dry-run
```

Checks the PR head into a throwaway worktree (`HUSKY=0`, so no Supabase slot is allocated), lays the change out **in the tree**, runs agy over it, posts findings as inline review comments, and writes the SHA-pinned marker `<!-- pinpoint-agy-review: <sha> -->`.

What agy is handed, all as files rather than prompt text:

- `AGY_REVIEW_BRIEF.md` — the PR's title, author and **current description**, base/head SHAs, and the right-side line ranges a finding may anchor to. The description is the thing a diff cannot supply: most of a review is checking whether the change does what its author says.
- `<file>.agy-diff.patch` beside each changed file — that file's diff against the merge-base, alone. Sliced from the same diff the hunk map and proof come from, so a sidecar can never disagree with the validator.
- `.agents/skills/pinpoint-agy-review/SKILL.md` — the review procedure, copied in by the wrapper.

The prompt is four lines and names the skill, because **skills do not auto-activate in headless `-p` mode** (measured — description matching never fires). Everything reviewable is a version-controlled file.

- **Gate 3 (`reviewed`)** accepts either marker — agy's or Claude's. One new prefix, one generalised jq helper.
- **Gate 2 (`threads`)** needed no change: #1801 made it count unresolved threads from any author, which is exactly what makes agy's inline findings block a merge.

## The review procedure

`.agents/skills/pinpoint-agy-review/SKILL.md` is adapted from the `adversarial-review` prompt in [openai/codex-plugin-cc](https://github.com/openai/codex-plugin-cc) (Apache-2.0, credited in the skill). The job is to break confidence in the change, not validate it.

The failure classes it works through are **general engineering ones** — auth boundaries, data loss, idempotency, effects escaping a transaction, races, degraded dependencies, schema drift, dishonest failure, observability — with `REVIEW.md` and the `CORE-*` catalogue as an **additional** pass rather than the whole exercise. A change can satisfy every `CORE-*` rule and still lose data.

Two things deliberately not taken from the source: its line-**range** anchoring (ours stays single-line, because one out-of-range line 422s the entire batch and loses everything), and its fourth severity level (`critical` and `high` would both mean "must fix before merge").

## Three agy defects this works around

All measured against real PRs, not assumed:

1. **It confabulates on read failure.** Handed a diff it could not read, agy did not error — it emitted a confident, detailed review of a *completely different PR* with `findings: []`. A false clean pass is the one failure mode a review gate cannot have. Every response therefore carries a `proof` object checked against ground truth, and **the marker is written only downstream of that check**. See *The proof challenge* below for why the shape of that proof matters more than its existence.
2. **A denied command ends the run.** Not "the model adapts" — the CLI shuts the stream down and returns `status: SUCCESS` with an empty response. One denial cost a 52-step review.
3. **It exits 0 when it produced nothing usable.** Validity is judged from the response envelope, never the exit code.

## The proof challenge

`/code-review medium` on this branch found that the first version of the proof guard did not guard anything, which is worth recording because the failure is subtle.

It asked for the file count and the first `diff --git` line, and `AGY_REVIEW_BRIEF.md` printed both under "copy these back verbatim". So a run that read only the brief — which also carries the title, author and full PR description — satisfied the check having never opened a single patch. That is exactly the confabulating run the guard exists to catch, and it would have ended in a posted review plus the SHA-pinned marker the merge gate accepts.

The regression came from moving the diff out of the prompt and onto disk. "Echo a fact about the diff" meant something when the diff *was* the prompt; once the brief summarised it, the same question was answerable from the summary.

Simply not printing the values would not have fixed it. The brief lists every changed path, so the first `diff --git` line is reconstructible from a filename in a fixed format, and the file count is the row count. **Any proof built from diff metadata is forgeable from a file that describes the diff.** It has to be content.

So the guard is now a challenge: the brief names a file and a line number of one sidecar patch, and the response must quote that line back, character for character. The answer appears nowhere in the brief. The line is chosen per commit from `head_sha` so `--at` replays ask the same question, and lines common to every patch (`@@`, `+++`, `index `) are skipped because quoting one back would prove nothing. `files_changed` stays as a cheap sanity check, now documented as explicitly not load-bearing.

## Permissions

agy runs with **no shell at all** — `read_file(*)` allowed, `command(*)` and `write_file(*)` denied. A curated allow-list was tried first and lost: agy improvises orientation commands, bracketed pathspecs defeat any multi-token rule (PinPoint's route segments are full of them), and a broader allow silently overrides a narrower deny. Denying everything is also the stronger guarantee — no shell means no `python3 -c` write-around of the `write_file` denial. The wrapper does all posting, so agy needs no `gh` and no write capability.

Config lives outside the repo (`~/.gemini/antigravity-cli/settings.json`), documented in the script docstring and Phase 3.4.

## Measured on real open PRs

Five verified findings across #1807/#1809/#1818/#1819, **zero false positives**. Two were substantive:

- **#1819** — a false-green guard in `test_pr_watch.py` that cannot fail: the helper hardcodes `commit_id` to a third value, so an implementation reading the wrong field still returns `stale_comments` and the assertion passes. Its docstring claims the opposite.
- **#1818** — `high`: a `SECURITY DEFINER` RPC returning a **decrypted Vault credential** with no in-body `auth.role()` guard, in a repo that had already hardened this exact shape twice (`0029`, `0057`). A `/code-review high` on that PR found six other defects and did not find this one.

Three wrapper bugs surfaced by that testing, each of which silently degraded a review rather than failing it — the skill never reaching the checkout, the response being a turn transcript rather than a JSON document, and sidecars written for generated files. The last one is why #1818 went from zero findings to that `high`: 2650 of its 5364 diff lines were a `drizzle/meta` snapshot, and the context spent on a blob was the context the finding needed.

## What is not yet established

**The adversarial procedure is unproven.** Because of the delivery bug above, every measurement in this PR was taken with **no procedure in effect** — that is the baseline. Its first run with the skill actually in place returned no findings on a PR where baseline had found the test bug. That is n=1 against n=1 on a diff that also changed between runs, so it is not evidence the skill hurts; it is an absence of evidence that it helps.

The honest summary: the harness is solid and the baseline detection is genuinely useful. Whether the adversarial layer improves on it needs more runs.

## Tests

`pnpm run check` and `pnpm run check:python` green (579 pytest). Two tests caught real bugs before they shipped: the diff parser counting the `+++ b/path` header as an added line, and a two-prefix jq that rebound `.` inside `select` so `unreviewed` became unreachable.

## Review note

This PR lowers the routine review bar — a smaller model reading a diff is not equivalent to `/code-review`. Phase 3.4 spells out when to hand off to Tim instead (auth, permissions, migrations, architectural changes). Worth a human read on that judgement in particular, and on the merge-gate change in `_pr-gates.sh`.

https://claude.ai/code/session_01CKTx2ek6cKVXva7QuU7HLX


timothyfroehlich (commented):
The PR cleanly implements the automated Antigravity (`agy`) code review workflow and gate integration (PP-c6xz).

Key changes verified:
- `scripts/workflow/agy_review.py`: Implements the review wrapper with diff proof verification (guarding against model confabulation), line validation against right-side diff hunks, bounded retries, and review payload assembly.
- `scripts/workflow/_pr-gates.sh`: Integrates `AGY_MARKER_PREFIX` (`<!-- pinpoint-agy-review: -->`) alongside the existing Claude marker, correctly handled in `_marker_verdict`.
- Unit tests (`scripts/tests/test_agy_review.py` and `scripts/tests/test_pr_gates_markers.py`): Thorough test coverage for diff parsing, proof verification, line anchoring, payload building, and jq marker verdict filtering.
- Documentation & Skills (`AGENTS.md`, `CLAUDE.md`, `REVIEW.md`, `.agents/rules/antigravity.md`, and `pinpoint-pr-workflow` skill): Updated clearly and consistently across all harness contexts.

No correctness defects or rule violations found. Looks good to merge!

—Antigravity

timothyfroehlich:
agy review of head 521b628 — agy-flash — 1 finding(s) via gemini-3.6-flash-high

AGENTS.md
```diff
@@ -189,7 +198,9 @@ Don't write the handoff summary — **run it and paste it**:
 bash scripts/workflow/merge-handoff.sh <PR>
 ```
```
timothyfroehlich, line 200: An extra code block fence ```` ``` ```` was added at line 200 (and matching ```` ``` ```` at line 270), causing all subsequent sections of `AGENTS.md` (including `### Review comments` and `### UI screenshots`) to be rendered inside a raw code block.

—Antigravity

timothyfroehlich (commented):
The PR introduces `agy_review.py` and updates the workflow documentation and merge gates (`_pr-gates.sh` and `merge-handoff.sh`) to support Antigravity (`agy`) automated reviews alongside Tim's `/code-review`.

The review wrapper implementation, proof verification guards, diff parsing logic, gate updates, and comprehensive test suite (`test_agy_review.py` and `test_pr_gates_markers.py`) look solid.

One formatting defect was found:
- `AGENTS.md`: Stray code block fences (` ``` `) introduced at line 200 and line 270 cause the rest of `AGENTS.md` to be wrapped inside an unclosed code block. Removing those extra fences will restore the proper Markdown section rendering.

—Antigravity

timothyfroehlich:
Closing for now — the agy reviewer work has stalled out and these two PRs both target PP-c6xz with conflicting branches. Reopen or start fresh from main when the work resumes. The bead (PP-c6xz) stays open. —Claude

PR #1815, opened 2026-08-03
docs(agents): the `!` ban is linted now, unsafe `as` is the gap (PP-8k07)

timothyfroehlich:
Two lines I wrote in #1804 went stale the same day #1808 merged. Docs-only, two files, ~10 lines.

**What was false:**

- `.agents/skills/pinpoint-typescript/SKILL.md` — "CORE-TS-007's ban on `!` is enforced **nowhere**: `@typescript-eslint/no-non-null-assertion` is never enabled … That third of a Critical rule rests entirely on review."
- `REVIEW.md:46` — "CORE-TS-007's `!` ban is not linted, so non-null assertions stay a review question."

Both were true when #1804 was reviewed and false when it merged.

**Verified against `eslint.config.mjs` on `main` rather than taken from the merge announcement:**

| | |
| :-- | :-- |
| `no-non-null-assertion` | `"error"` (line 165) |
| disable-comment escape | closed — `eslint-comments/no-restricted-disable` names it (208–211) |
| off where | the test-files block (267) and the e2e block (300) only |
| **not** off in | `*.config.*` / `scripts/**` — which is a real difference from the `no-explicit-any` sentence directly above it in the skill, so the two paragraphs can't share a scope claim |

The test exemption is temporary (PP-9q5k), so the skill says not to read it as blessed.

**What replaces it:** the third of CORE-TS-007 that genuinely has no gate is **unsafe `as`**. `no-unsafe-*` targets untyped values, not casts between two known types, and `tsc` accepts such a cast by construction. `REVIEW.md` now routes reviewers at that instead of at `!`, which lint already catches.

The frontmatter claim "which parts of CORE-TS-006/007 no tool enforces" is unchanged — still true, just a different part. `pnpm run check` green.

Credit: Claude-NonNullAssertion flagged the staleness in the huddle.


timothyfroehlich:
Claude review of head 46b2f81 — `/code-review medium` — 2 findings, both fixed: frontmatter routed on the two now-linted thirds instead of unsafe `as`; scope claim omitted that scripts/config keep the rule at error but disable no-restricted-disable

PR #1816, opened 2026-08-03
test(machines): pin that a clean save disarms the unsaved-changes guard (PP-o355.26)

timothyfroehlich:
## What

Adds one RTL test to `machine-details-form.test.tsx`: after a Details save settles successfully, the unsaved-changes guard disarms.

## Why

PP-o355.26 was filed by the final whole-branch review of PP-o355.19 — "unsaved Details changes are discarded silently on navigation." Working it turned out to be an audit, not a build: the fix landed **inside that same PR** (#1762), in response to the same review. `main` already has both halves of the guard:

- `beforeunload` for exits that unload the document (reload, tab close, off-site links)
- a document-level click-capture guard for in-app navigation, which is what catches the tab-strip links — App Router navigations never unload the document, so `beforeunload` cannot see them

The bead's framing is also stale: it cites "the back link at `edit/page.tsx:134-140`", which no longer exists — the page moved inside the `(tabs)` group, so the navigation vector is the tab strip, and the click guard covers it.

Eight tests already pin the armed side (dirty interception, keep-editing, discard-and-navigate, same-page link, modified clicks) and the Cancel disarm. The gap was the **success** disarm — the one users hit every time they save. Deleting `setIsDirty(false)` from the success effect left 19 of 20 tests green, so nothing was holding the line between "you saved" and "a prompt about the changes you just saved."

## The test

Mirror of the existing "keeps guarding edits made while a save is in flight": that one pins the case where the in-flight snapshot is stale (still dirty, correctly), this one pins the case where it is current. Asserts all three consequences of a clean settle — the note reads "Saved", `beforeunload` stops blocking, and a tab-strip link navigates natively rather than through the discard dialog's `router.push`.

## Decision recorded

The remaining known gap is deliberate and stays: the browser **Back** button within the app fires non-cancellable `popstate`. Guarding it means pushing a sentinel history entry and re-pushing on every pop, which breaks Back for as long as the form is dirty — a user who cannot leave is worse than the bug it fixes. (`beforeunload` does cover Back when it leaves the origin.) That rationale is already written into the source comment; this PR does not change it.

## Testing

- `pnpm run test` — 220 files, 2032 tests, all pass
- `pnpm run check` — clean
- Mutation-checked: breaking the disarm in the source fails this test and only this test

https://claude.ai/code/session_01QEYGHS3gwk3HP1C66s2pyY

### DIFF
--- a/src/app/(app)/m/[initials]/(tabs)/edit/machine-details-form.test.tsx
+++ b/src/app/(app)/m/[initials]/(tabs)/edit/machine-details-form.test.tsx
@@ -447,12 +447,13 @@ describe("MachineDetailsForm", () => {
 
       await user.click(screen.getByRole("link", { name: "Settings" }));
 
+      // Queried synchronously on purpose. The armed case mounts this dialog
+      // synchronously too (the positive tests above `await findByText` only
+      // because userEvent yields), so a guard that failed to disarm would be
+      // visible right here rather than settling in after the assertion.
       expect(
         screen.queryByText("Discard unsaved changes?")
       ).not.toBeInTheDocument();
-      // The click was never intercepted, so the anchor navigates natively —
-      // `router.push` is what the DISCARD path uses, and it must stay untouched.
-      expect(pushMock).not.toHaveBeenCalled();
     });
   });
 });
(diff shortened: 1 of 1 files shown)


timothyfroehlich:
Claude review of head 8c621ed — `/code-review medium` — 1 low finding — a non-load-bearing pushMock assertion whose comment claimed it verified native anchor navigation; dropped, mutation check re-verified

PR #1817, opened 2026-08-03
docs(agents): cut the testing corpus to its layer decisions (PP-22e4)

timothyfroehlich:
Sweep task 4 of PP-22e4. `pinpoint-testing` **420 → 107 lines**. Docs-only; DELETE / MOVE-VERBATIM / POINTER only.

## Kept verbatim

Three things, all decisions that exist nowhere else:

- **The bug-class table (A–J).** AGENTS.md rule 9 routes here for it by name (_"Bug-class table: `pinpoint-testing` skill"_).
- **"Where Existing Coverage Lives."** The canonical file per bug class. The 2026-05 audit's actual finding was that agents write duplicate coverage because they can't see what already exists — this table is the remedy, not a restatement.
- **"Test What We Own" in full** — the diagnostic question, the decision rule, the OK/NOT-OK lists and their casework (PP-e20, PP-uc8, PP-q9r). CORE-TEST-006 is the binding form; the casework is what makes it actionable.

## Deleted, and what owns each fact

| Deleted | Owner |
| :-- | :-- |
| "When to Use This Skill", "Testing Checklist" | the frontmatter description; CORE-TEST-001/002/004/005/006 |
| "Critical Rules" (6 items) | five are CORE-TEST restatements. Item 6 — integration test locations — was the only fact, kept as one line |
| Commands block, "Which Tests to Run", "Key rules for agents", "Reproducing CI failures locally" | AGENTS.md §5 |
| All code examples (unit, PGlite integration, Playwright — ~190 fenced lines) | real tests in `src/test/` and `e2e/` |
| "Test Organization" directory tree | the filesystem |
| "PGlite Best Practices" | CORE-TEST-001 plus its two illustrative snippets |
| "E2E Patterns" (selector strategy, common patterns) | `pinpoint-e2e` — the section itself opened by saying so |

The decision tree was **triplicated** across AGENTS.md, this skill, and `pinpoint-e2e`. Both skill copies are now pointers.

## The one anti-pattern that survived

Of the four anti-patterns, three are textbook (implementation details, arbitrary waits) or CORE-TEST-002 (server components). The **over-mocking** entry is neither: the house pattern is to `vi.mock("~/server/db")` so it *forwards* to worker-scoped PGlite, meaning real SQL runs. Its interaction with CORE-TEST-001 is non-obvious enough to be worth stating — it looks like a violation and isn't, because `getTestDb()` returns the worker-scoped instance. Kept as prose citing `transaction-tripwire.test.ts` rather than as a fenced pair.

## Pointer check

`pinpoint-e2e/SKILL.md:30` pointed at the deleted decision tree — repointed at AGENTS.md §5. Its other two references (bug-class table, coverage table) target surviving sections and are unchanged.

Verified all 16 linked test paths still exist on disk before keeping the coverage table. Dated audit records under `docs/testing/` left alone per the sweep's frozen-records rule.

`pnpm run check` green.


timothyfroehlich:
Claude review of head ec5286a — `/code-review medium` — 4 findings, all fixed: selector-strategy pointer aimed at a skill that never had it (moved verbatim to pinpoint-e2e); test:integration mislabelled as the Supabase suite in DEVELOPMENT.md + src/test/README.md when it excludes that dir; src/test/README.md added to Elsewhere for the setupTestDb/getTestDb contract; four .agent/skills (singular) pointers corrected

PR #1818, opened 2026-08-03
feat(pinballmap): outbound list/unlist plumbing (PP-o355.30)

timothyfroehlich:
## Summary

Wires PinballMap's write verbs to the app. Both ends already existed and nothing connected them: `client-live.ts` has implemented `addMachine` / `removeMachine` since PP-o355.5, `pinballmap_state` has carried `outbound_email` + `outbound_token_vault_id`, the permission matrix has `machines.pinballmap.push`, and the `pinballmap_listing` timeline event already accepted `action: "listed" | "unlisted"`. **`addMachine` and `removeMachine` had never been invoked by anything in the app.** This builds the middle.

How the gap happened: PP-o355.5's notes say outbound list/unlist was "pulled FORWARD into PP-o355.12", but .12 closed as "PBM listing control read-side" — it shipped the read half and the write half fell between the two beads with nothing tracking it.

- **Vault-backed credentials accessor** — `get_pinballmap_credentials()` mirrors `get_discord_config()` (migration 0028) exactly: `SECURITY DEFINER`, pinned `search_path`, REVOKE from PUBLIC/anon/authenticated, EXECUTE to `service_role` only. A half-provisioned row (email without token, or the reverse) reads as *not provisioned* rather than yielding a credential PBM would reject — which we'd then report to the user as a bad token.
- **`listMachineOnPinballMapAction` / `unlistMachineFromPinballMapAction`** — gated on `machines.pinballmap.push`. CORE-ARCH-011 sets their shape: the Vault decrypt and the PBM HTTP call are both non-transactional effects, so they run to completion first and only their *results* enter the transaction.
- **Stored-snapshot edits** — the reason this bead exists (see below).
- **`supabase/seed-pinballmap-creds.mjs`** — one-shot provisioning, modeled on `seed-discord.mjs`.

**No UI.** PP-o355.21 wires these into the six-state listing control. `.30` is deliberately a complete, testable backend slice.

## The unlist ↔ auto-link interaction

Auto-link (PP-o355.20, merged this morning in #1810) re-lists any matched, unlisted cabinet whose title appears on the **stored** lineup. So an unlist that clears `pinballmapListed` but leaves the lmx in the stored snapshot gets silently undone by the next reconcile pass — or by any save on that machine inside the hour, before the cron refreshes. A Remove button that doesn't do this is a button that doesn't work.

Both actions therefore edit the stored snapshot inside the same transaction. We know exactly which lmx we added or deleted, so it costs no extra PBM call.

`survives the next reconcile pass — auto-link does not undo it` is red-checked: neutering the snapshot edit fails that test and **only** that test.

## Notable decisions

- **`authorizeListingRead` → `authorizeListingAction(formData, permission)`** — committed separately as a pure rename. Copying a 45-line preamble to vary one string is the duplication worth collapsing at two rather than three; a permission gate is load-bearing.
- **The seed script diverges from `seed-discord.mjs` twice, deliberately.** It UPSERTs, because no migration seeds the `pinballmap_state` singleton (the app creates it lazily) — a missing row is normal here, not evidence migrations were skipped. And it stays prod-capable behind `SEED_PINBALLMAP_CREDS_FORCE_PRODUCTION=1` rather than refusing prod outright, following `mark-migration-applied.ts`. Discord's blanket refusal exists because its seed runs inside `db:reset`, where an automatic step could push a dev token into prod's Vault; this one only runs when a human types it, and provisioning prod is its entire purpose.
- **Not build-gated** (CORE-SEC-009). Unset credentials → both actions return `NOT_PROVISIONED` and say so; every other surface, PBM sync included, is untouched.

## Test Plan

- [x] `pnpm run check` green
- [x] Unit: 2051 passed (incl. 5 credentials-accessor + 6 snapshot-edit)
- [x] Integration: 551 passed (incl. 8 new in `pinballmap-outbound-write.test.ts`)
- [x] `next build` green
- [x] RPC privileges verified against the local DB: `authenticated=f`, `anon=f`, `service_role=t`
- [x] Seed script end-to-end with a throwaway local dummy — skip path writes nothing; a seeded value reads back through `get_pinballmap_credentials()`, proving migration + Vault write + accessor SQL agree; a re-run with a different token leaves the original in place
- [x] Red-check on the auto-link survival test

PBM is never reached from tests — the client is pinned at the seam (CORE-TEST-006), credentials stubbed at `~/lib/pinballmap/credentials` (PGlite has no `vault` schema).

## Before PP-o355.21 ships

A real PinballMap operator email + token must be seeded into prod's Vault. That's yours to run — no agent has handled or should handle those values.

Closes PP-o355.30.


src/app/(app)/m/pinballmap-actions.ts
```diff
@@ -285,6 +298,225 @@ export async function linkPinballmapEntryAction(
...
+  const lmxId = written.lmxId;
+  // --- transaction: local state only ---
+
+  try {
+    await db.transaction(async (tx) => {
+      await tx
+        .update(machines)
+        .set({ pinballmapLmxId: lmxId, pinballmapListed: true })
+        .where(eq(machines.id, machine.id));
+      if (state.snapshotJson) {
```
timothyfroehlich: **Stale read-modify-write of `snapshot_json` — concurrent list/unlist silently lose each other's edit.**

`state` is read at line 361, *before* the credential decrypt and the PBM HTTP call, and the transaction then writes the whole `snapshot_json` blob back from that pre-call copy. Anything that refreshed the snapshot in the window is clobbered. Same shape in the unlist action (line 498).

Concrete: an admin lists three cabinets in quick succession. `serializeWrite` in the live client queues the PBM POSTs, so all three requests have already read the *same* base snapshot. Each transaction writes `base + its own lmx`; the last one wins. Two of the three machines end up `pinballmapListed = true` with their lmx missing from the stored lineup → `derivePbmMachineStatus` reports `listed_locally_absent_on_pbm`, the desync counter inflates, and the badges stay wrong until the next hourly cron. The hourly cron landing mid-window has the inverse effect: it gets reverted to an hour-old lineup.

Fix: re-read the singleton *inside* the transaction with `.for("update")` and apply `withLmxAdded`/`withLmxRemoved` to that row, so the read-modify-write is serialized by the row lock.

—Claude

src/lib/pinballmap/snapshot-edit.ts
```diff
@@ -0,0 +1,52 @@
...
+ * A no-op when that lmx is already listed — PinballMap's create is
+ * find-or-create, so re-listing a title already on the lineup hands back the
+ * EXISTING lmx rather than minting a second one.
+ */
+export function withLmxAdded(
+  snapshot: LocationSnapshot,
+  lmxId: number,
+  pinballmapMachineId: number
+): LocationSnapshot {
+  if (snapshot.lmxes.some((l) => l.id === lmxId)) return snapshot;
```
timothyfroehlich, line 40: **Dedupe is by lmx `id`, but every consumer looks the row up by `machineId`.**

PBM gives a title exactly one lmx at our location, so the snapshot should never carry two rows with the same `machineId` — but this guard only suppresses an exact `id` match. If the stored snapshot still holds a *stale* lmx for the same title, the append creates a duplicate `machineId`.

Concrete: stored snapshot has `{id: 100, machineId: 42}` (PBM deleted + re-added the row outside its resurrection window, and the cron hasn't refreshed yet). Admin lists a matched cabinet for title 42; `addMachine` returns the new `lmxId: 200`; `withLmxAdded` appends it, leaving both 100 and 200 under `machineId: 42`. `derivePbmMachineStatus` then does `snapshot.lmxes.find(l => l.machineId === 42)` → gets **100** → `row.id (100) !== pinballmapLmxId (200)` → the machine we just successfully listed reports `lmx_drifted`, and `reconcileAfterSync` "heals" `pinballmapLmxId` back to 100, an lmx that no longer exists on PBM. `findLmxForMachine` has the same array-order dependency the test on line 58 says it wants to avoid.

Fix: keep the identical-`id` early return (it preserves the existing row's conditions), then filter out any other row with the same `pinballmapMachineId` before appending.

—Claude

src/app/(app)/m/pinballmap-actions.ts
```diff
@@ -285,6 +298,225 @@ export async function linkPinballmapEntryAction(
...
+          eventData: { kind: "pinballmap_listing", action: "listed", lmxId },
+          actorId: userId,
+        },
+        tx
+      );
+    });
+  } catch (error) {
+    // One lister per title at our location. PBM's find-or-create already handed
+    // us the shared lmx, so the honest report is that another cabinet holds it.
+    if (isPgErrorCode(error, "23505")) {
```
timothyfroehlich, line 478: **The one-lister collision is only detected *after* the PBM write, so a rejected list can leave an orphan public listing.**

`linkPinballmapEntryAction` can afford a post-hoc 23505 catch because it performs no PBM write. Here the POST has already happened by the time the transaction rolls back.

Concrete: cabinet A of title 7 is `pinballmapListed = true` locally but the entry was deleted directly on pinballmap.com — that is exactly the `listed_locally_absent_on_pbm` state `derivePbmMachineStatus` models. An admin tries to list cabinet B (same title). `addMachine` is find-or-create, finds nothing, and **creates a real lmx on PinballMap**. The local transaction then trips `machines_pinballmap_listed_unique`, everything rolls back, and the user is told "another cabinet already holds the listing." Net result: PinballMap's public lineup gained an entry that PinPoint has no `pinballmapLmxId` for and no timeline event about — nothing can unlist it through the app.

Fix: query for a listed sibling (`pinballmapMachineId = titleId AND pinballmapListed AND id <> machine.id`) *before* the credential decrypt and the PBM call, and return the conflict message there. Keep the 23505 catch as the race backstop. It also stops spending a PBM write on a request we already know we will refuse (CORE-PBM-001).

—Claude

src/app/(app)/m/pinballmap-actions.ts
```diff
@@ -285,6 +298,225 @@ export async function linkPinballmapEntryAction(
...
+      return err(
+        "VALIDATION",
+        await pbmListingConflictMessage(titleId, machine.id)
+      );
+    }
+    throw error;
+  }
+
+  revalidatePath(`/m/${machine.initials}`);
```
timothyfroehlich, line 487: **Revalidation is scoped to one machine, but the write mutates the shared snapshot every machine's badge derives from.**

`syncPinballMapNowAction` calls `revalidatePath("/m", "layout")` with the comment "Desync badges on machine Info cards derive from the stored snapshot, so refresh the whole machine subtree after a successful sync." These two actions now edit `snapshot_json` too, so the same reasoning applies.

Concrete: title 7 has two cabinets, A (listed) and B (matched, not listed). B's Info card shows the `on_pbm_not_listed_locally` desync badge off the stored lineup. An admin unlists A — the lmx leaves the snapshot, so B's badge is now wrong — but only `/m/<A-initials>` is revalidated, so B's cached page keeps showing the stale badge until something else revalidates it.

Same at line 516 for the unlist action.

—Claude

src/app/(app)/m/pinballmap-actions.ts
```diff
@@ -285,6 +298,225 @@ export async function linkPinballmapEntryAction(
...
+  const credentials = await getPinballMapWriteCredentials();
+  if (!credentials)
+    return err(
+      "NOT_PROVISIONED",
+      "No Pinball Map operator account is set up yet, so PinPoint can't write to Pinball Map."
+    );
+
+  const client = await getPinballMapClient();
+  const written = await client.removeMachine({ credentials, lmxId });
+  if (!written.ok) {
```
timothyfroehlich: **`not_found` on remove is the one PBM rejection where the desired end state is already true — refusing it strands the machine.**

If PBM reports the lmx is gone, the local row keeps `pinballmapListed = true` and a dead `pinballmapLmxId`, and every subsequent click of Remove hits the same `not_found` and fails again.

Concrete: someone deletes the entry directly on pinballmap.com (the `listed_locally_absent_on_pbm` desync the status module exists to surface). The admin clicks Remove to resolve it. `removeMachine` → `not_found` → `PBM_REJECTED`, "Pinball Map couldn't find that entry. It may already be gone." — and the flag is still set. `verifyPinballmapLinkAction` returns `stale` and deliberately leaves the stored link alone, so the only remaining way out is to unlink the machine from its catalog title entirely and re-link it.

Worth deciding deliberately: treating `not_found` as "already removed" and proceeding to the local clear reaches the state the user asked for and is arguably more honest than a failure toast for a removal that is, in fact, done. Left unfixed pending your call, since it is a behavior choice rather than an outright bug.

—Claude

supabase/seed-pinballmap-creds.mjs
```diff
@@ -0,0 +1,169 @@
...
+    // app upserts it lazily the first time the integration is configured. So a
+    // missing row is normal here, not a sign migrations were skipped, and this
+    // script creates it.
+    const existing = await sql`
+      SELECT outbound_token_vault_id
+      FROM pinballmap_state
+      WHERE id = 'singleton'
+    `;
+
+    if (existing[0]?.outbound_token_vault_id) {
```
timothyfroehlich: **"Already provisioned" tests only the vault id, but `getPinballMapWriteCredentials` requires *both* halves.**

The accessor returns `null` — and the actions return `NOT_PROVISIONED` — when `outbound_email` is null even if `outbound_token_vault_id` is set. This guard treats a vault id alone as fully provisioned.

Concrete: a row ends up with a vault id and a null/blank `outbound_email` (a hand-edited row, or a restore that dropped the plain column). The operator runs the seed script with both env vars set; it prints "✅ PinballMap operator token already provisioned — leaving DB values untouched" and exits 0, while the app permanently reports "No Pinball Map operator account is set up yet." The documented provisioning path reports success and the feature stays dead, with no supported repair.

Fix: select `outbound_email` alongside `outbound_token_vault_id` and require both before short-circuiting.

—Claude

src/app/(app)/m/pinballmap-actions.ts
timothyfroehlich: Fixed in f1e1fae8. New \`editStoredSnapshot(tx, edit)\` re-reads the singleton inside the transaction with \`.for("update")\` and applies the edit to that row, so the row lock serializes the read-modify-write. Both actions use it. —Claude-PbmListUnlist

src/lib/pinballmap/snapshot-edit.ts
timothyfroehlich: Fixed in f1e1fae8. \`withLmxAdded\` now filters any other row with the same \`machineId\`; the identical-\`id\` early return stays, since the stored row carries conditions and \`lastUpdatedByUsername\` a synthesized one does not. Unit test added for the stale-lmx case. —Claude-PbmListUnlist

src/app/(app)/m/pinballmap-actions.ts
timothyfroehlich: Fixed in f1e1fae8. \`findListingIncumbent()\` refuses before the credential decrypt and the PBM call; the 23505 catch stays as the race backstop. Agreed on the asymmetry with \`linkPinballmapEntryAction\` — that one performs no PBM write, so a post-hoc catch costs nothing there. —Claude-PbmListUnlist

src/app/(app)/m/pinballmap-actions.ts
timothyfroehlich: Fixed in f1e1fae8. \`revalidatePath("/m", "layout")\` added to both actions, matching \`syncPinballMapNowAction\`. —Claude-PbmListUnlist

src/app/(app)/m/pinballmap-actions.ts
timothyfroehlich: Fixed in 6ea68106 — Tim called it: treat \`not_found\` as already unlisted. We do reach the state we report (not on PBM, not listed locally), so CORE-ARCH-012 holds; we just did not have to do the deleting. Every other reason still refuses. Integration test added. —Claude-PbmListUnlist

supabase/seed-pinballmap-creds.mjs
timothyfroehlich: Fixed in f1e1fae8. The script now selects both halves and requires both; a row with a vault id and no email errors with repair instructions instead of printing success over a permanently unprovisioned state. —Claude-PbmListUnlist

timothyfroehlich:
Claude review of head 6ea6810 — `/code-review high` — 6 findings, all real: stale snapshot read-modify-write (now FOR UPDATE inside the tx), lmx dedupe by id vs machineId, one-lister collision detected only after the PBM write (now pre-checked), machine-scoped revalidation of a shared snapshot, not_found on remove stranding the machine (Tim: treat as already unlisted), seed script reporting success on a half-provisioned row. All fixed and threads resolved.

PR #1819, opened 2026-08-03
feat(workflow): post /code-review findings to the PR, and count them (PP-97tt)

timothyfroehlich:
## Summary

- The command handed to Tim for review is now `/code-review <depth> --comment <PR#>`. `--comment` posts each finding to the PR as an inline review comment instead of leaving it in his terminal, which makes the `threads` gate enforce that every finding is fixed or declined-and-resolved.
- The `reviewed` gate accepts those comments. They pin the commit the reviewer read, so a review that found something attests itself — no waiting on an agent to notice and post a marker.
- `mark-claude-review.sh` is unchanged in purpose and still required for the cases that leave no comments behind: a review that found nothing, an `ultra` run (the cloud path can't post comments), a run without the flag, and the trivial carve-out. It is also still worth posting alongside comments, which record no depth.

Two flag limits are documented everywhere the command appears, because neither is discoverable: `--comment` only posts when the target is a **GitHub PR** (against a local branch it is silently ignored), and `ultra` has no comment-posting counterpart — the cloud path honours only `--fix`. Both verified against the CLI's own instruction text.

## The two field choices this turns on

- **Comments pin `original_commit_id`, never `commit_id`.** GitHub re-anchors the latter as a PR advances so a still-applicable comment stays on a live line; reading it would report every review comment as pinning head forever — a permanent false green.
- **Replies are excluded** (`in_reply_to_id != null`). A reply is created at whatever head is current, so an agent answering the thread it just fixed would otherwise re-attest a commit no reviewer has seen.

Both are pinned by tests that fail loudly if the wrong field is read.

## Gate states

Three become five. `commented` / `stale_comments` join the marker trio, kept apart because the remedies differ — only the marker clears `stale_comments`, since nothing re-posts a reviewer's comments at a new SHA. Markers outrank comments by fixed precedence rather than timestamp (the two clocks aren't comparable — a marker's `updated_at` moves when the sticky comment is rewritten). A `gh` failure on either lookup reads as "no evidence there", which is fail-closed by construction.

## Known consequence

Because agents act under Tim's GitHub identity, the gate cannot distinguish his review comments from an agent's. This is the same honour system the marker already runs on, now with a second surface — raised before implementing and accepted. `.agents/rules/antigravity.md` is updated to say so plainly rather than repeat the now-false "your review does not satisfy the merge gate".

## Test Plan

- [x] `pnpm run check` clean
- [x] `pnpm run check:python` — 514 passed (17 new tests across the gate, the pr-watch mirror, and the handoff report)
- [x] Hand-verified the rendered output of all five gate states
- [ ] Review

## Related Issues

PP-97tt


scripts/workflow/_pr-gates.sh
```diff
@@ -137,14 +207,22 @@ check_ci() {
...
+#   stale_comments  review comments exist but pin a DIFFERENT SHA — likewise
+#   unreviewed      no evidence on this PR at all — nobody has reviewed it
 #
-#   marker        a review marker pins head's SHA — head has been reviewed
-#   stale_marker  a marker exists but pins a DIFFERENT SHA — head was never reviewed
-#   unreviewed    no marker on this PR at all — nobody has reviewed it
+# The two stale states are kept apart so the report can name WHICH evidence went stale,
+# and because one of them has a trap the other doesn't: a stale marker can be re-attested
+# by rewriting the same sticky comment, whereas nothing re-posts a reviewer's inline
+# comments at a new SHA — so `stale_comments` is cleared by `mark-claude-review.sh` too,
```
timothyfroehlich: **"`stale_comments` is cleared by `mark-claude-review.sh` too, and never by touching the threads" is asserted here, but nothing enforces it.**

`_review_record` never asks *which* comment cleared the state — only whether any top-level comment has `original_commit_id == head`. So a third exit exists that this comment (and `_review_remedy`, `check_review_happened`, REVIEW.md, AGENTS.md and the pr-workflow skill, which all repeat the claim) says does not: **posting a new top-level inline comment on the new head.**

Scenario: Tim reviews `abc123` with `--comment`; the agent pushes genuinely new work, so the gate reads `stale_comments`. The agent then posts one new top-level inline comment on the new head — a note-to-reviewer, a follow-up on its own diff, or a deliberate self-clear — and `_review_record` returns `commented` → PASS. No marker, no review of the new head, no `--force`.

That is the same honour-system hole the file already documents for `commented` in the abstract, but here the prose makes a stronger claim than the code delivers: it tells the reader the marker is the *only* way out of `stale_comments`, which reads as an enforced property rather than a convention. Worth softening to "the marker is the honest way out", so no later change relies on the stated invariant.

—Claude

.claude/settings.json
```diff
@@ -1,5 +1,9 @@
 {
   "permissions": {
+    "allow": [
+      "mcp__github_inline_comment__create_inline_comment",
+      "Bash(gh api repos/timothyfroehlich/PinPoint/pulls/*/comments*)"
```
timothyfroehlich, line 5: **This allow-rule is narrower in practice than `scripts/workflow/AGENTS.md` claims, and wider than it needs to be in one respect.**

*Narrower:* I hit the exact failure this PR says the rule fixes — on this PR, with this rule checked in. `gh api …/pulls/1819/comments --method POST … -f body='<multi-line text>'` was still **refused by the auto-mode classifier**; the identical call with the body read from a file (`-F body=@/tmp/f.md`, whole command on one line) went through. So the prefix rule only matches when the command stays single-line — and a review finding with paragraphs, a fenced suggestion block, or a code snippet is exactly the shape that fails. Worth noting at the rule, because `scripts/workflow/AGENTS.md` now tells the next reader "`.claude/settings.json` now allows both" and "if a review's findings come back as terminal text rather than threads, that rule is what to check first" — which sends them to a rule that is present and still not matching.

*Wider:* the trailing `*` after `comments` matches any suffix, so this also auto-approves `--method PATCH` / `--method PUT` and arbitrary `-f`/`-F` payloads against that path, not only the intended POST of a finding. Given AGENTS.md §2.2 rule 6 ("never wildcard tool permissions … without explicit user approval each time"), pinning `--method POST` — or dropping the Bash rule and relying on the MCP tool alone — would match the stated intent more closely.

Both matter more since this PR: a top-level inline comment on head is now sufficient to clear the `reviewed` merge gate, so this rule is what decides whether posting one needs a human prompt.

—Claude

scripts/workflow/_pr-gates.sh
```diff
@@ -36,61 +38,129 @@ _repo_slug() {
...
-           elif ($markers | length) > 0 then ($markers | last) + { state: "stale_marker" }
-           else { state: "unreviewed", sha: "", depth: "", at: "", summary: "" }
-           end
-         | [ .state, .sha, .depth, .at, .summary ] | @tsv'
+               summary: ($b | split("\n") | last | gsub("^\\s+|\\s+$"; "")) } ]')
+
+  # Grouped by the SHA reviewed, so a review that raised six findings is ONE piece of
+  # evidence rather than six — the summary then reports the count, which is the fact a
+  # reader of the handoff report wants.
+  comments=$(gh api --paginate "repos/${owner_repo}/pulls/${pr}/comments" \
```
timothyfroehlich: **Both lookups page at GitHub's default 30/request, and this function now issues two of them per evaluation.**

`pr-watch.py`'s mirror asks for `?per_page=100` (`_gh_api_list`); the bash side asks for nothing, so `--paginate` walks 30 items at a time. On a PR with 90 inline comments that is 3 requests where 1 would do, and the same for the issue-comments call.

That is normally a shrug, but three things stack here: `_review_record` went from one paginated call to two; `merge-handoff.sh` and `merge-pr.sh` each call it; and `pr-dashboard.sh` calls `_compute_review_state` **once per open PR** in a loop. With ~10 open PRs and a couple of chatty ones, a single dashboard run can now spend several dozen requests on the shared 5000/hr host quota — and a quota 403 mid-run surfaces as a `gh` failure, which this path reads as "no evidence" (fail-closed, but indistinguishable from a genuinely unreviewed head).

Appending `?per_page=100` to both URLs matches the Python mirror and costs nothing.

—Claude

scripts/workflow/_pr-gates.sh
timothyfroehlich: Fixed in ffb27ec4. You're right that the prose read as an enforced invariant — it said "only by the marker" in five places, and nothing enforces that. Reworded to "honest exit" throughout, and I stated the enforced part separately, because it is narrower but still real: what a reviewer already wrote cannot be made to cover a commit they never saw. Posting a *new* comment on the new head does flip the state, and both `_pr-gates.sh` and the pr-workflow skill now say so explicitly rather than leaving it implied by the `commented` discussion further up. —Claude

.claude/settings.json
timothyfroehlich: Both halves fixed in ffb27ec4.

Wider than needed: correct, and worse than you flagged — the trailing `*` also covered `--method DELETE` on that path, so removing a review comment was auto-approved. `deny` now carries `--method`/`-X` entries for DELETE, PATCH and PUT; deny beats allow, so posting is what remains. Posting is what was authorised.

Weaker than advertised: also correct, and the more useful finding — you hit the exact failure the doc credited the rule with fixing. Both docs now say the rule makes posting *usually* work rather than reliably work, and name the shapes you saw refused (multi-line `-f body=`, and a `-F body=@file` refused once then allowed on retry) so the next reader treats a blocked post as expected rather than as a regression to debug. —Claude

scripts/workflow/_pr-gates.sh
timothyfroehlich: Fixed in ffb27ec4 — `?per_page=100` on all three paginated calls, matching the mirror.

This one had a tail. Adding it failed ten tests in `test_merge_pr_automerge.py`, whose stub matched `*"/issues/"*"/comments")` with no trailing `*`: the query string fell through to the catch-all, the markers were never returned, and the gate read `unreviewed`. That stub also had no `pulls/*/comments` branch at all, so review-comment lookups had been falling through silently since the first commit — invisible only because "no answer" and "no review comments" produce the same result. Both endpoints now match explicitly. —Claude

timothyfroehlich:
/audit-override four high advisories live on main (undici, fast-uri, ip-address, brace-expansion) are unrelated to this PR, which touches only workflow scripts, tests and docs — no dependency changes. #1821 clears all twelve repo-wide; arming this so #1819 can be reviewed against green CI instead of waiting on that merge order.

.claude/settings.json
```diff
@@ -1,5 +1,9 @@
 {
   "permissions": {
+    "allow": [
+      "mcp__github_inline_comment__create_inline_comment",
```
timothyfroehlich, line 4: **The gate's last human checkpoint is now auto-approved.** `commented` PASSes on *any* top-level inline comment pinned to head, and this rule auto-approves the exact surface that creates one — no prompt, for any repo/PR the MCP tool can reach.

Concrete scenario: an agent finishes a PR nobody has reviewed, calls `create_inline_comment` once ("nit: consider extracting this"), and `check_review_happened` flips from `unreviewed` to `PASS: reviewed: inline review comments pin head SHA …`. Before this PR the only ways to green that gate were `mark-claude-review.sh` (whose own header names posting-without-a-review as a false attestation) or a permission prompt Tim would see.

The PR body says the honour-system consequence was raised and accepted — but what was accepted was "the gate can't tell Tim's comments from an agent's". Removing the permission prompt is a separate, additive lever, and it isn't discussed anywhere in the diff. Worth at least scoping the allow to a narrower shape, or leaving `create_inline_comment` on `ask` so posting review evidence stays a deliberate act.

.claude/settings.json
```diff
@@ -9,6 +13,12 @@
       "Bash(killall:*)",
       "Bash(gh issue delete *)",
       "Bash(gh repo delete *)",
+      "Bash(gh api * --method DELETE*)",
+      "Bash(gh api * --method PATCH*)",
+      "Bash(gh api * --method PUT*)",
+      "Bash(gh api * -X DELETE*)",
```
timothyfroehlich: **These patterns assume the method flag comes after the path, which is not the usual invocation order.** Each glob is `gh api ` + `*` + ` --method DELETE` (or ` -X DELETE`) + `*`, so it only matches when something sits between `gh api ` and the flag — i.e. `gh api <path> -X DELETE`. Written the other way round, `gh api -X DELETE repos/...`, the remainder after `gh api ` begins with the flag and there is no leading space for the glob to line up with, so none of the six rules match. The attached forms (`-XDELETE`, `--method=DELETE`) match no rule in either position.

This matters because `scripts/workflow/AGENTS.md` (added in this PR) states the coverage as a fact: *"the `deny` list carries `--method`/`-X` entries for `DELETE`, `PATCH`, and `PUT`; deny wins over allow."* A reader takes that as method-level coverage of the allowed path, and it is really position-dependent coverage.

It also touches the allow rule above directly: `repos/timothyfroehlich/PinPoint/pulls/1/comments/123` matches `pulls/*/comments*`, so calls on comment sub-paths are allow-matched and these rules are the only thing narrowing them back to posting.

Empirical note from this review run: `-X POST` on the allowed path was refused while `--method POST` on the identical call went through, which is the same shape-sensitivity from the other direction and is worth knowing given the PR's own note about intermittent refusals.

Either write the entries so they match the flag wherever it appears, or reword the AGENTS.md sentence to describe what the rules actually cover.

scripts/workflow/_pr-gates.sh
```diff
@@ -36,61 +38,135 @@ _repo_slug() {
...
+# OPEN PR. At 30/page that is a real bite out of the 5000/hr quota shared with every
+# other session on the host, and a 403 here reads as "no evidence", so exhausting the
+# quota degrades into gate failures rather than into an error anyone would recognise.
+_review_record() {
   local pr=$1 owner_repo=$2 head=$3
-  gh api --paginate "repos/${owner_repo}/issues/${pr}/comments" \
-    | jq -rs --arg prefix "$CLAUDE_MARKER_PREFIX" --arg head "$head" \
+  local markers comments
+
+  markers=$(gh api --paginate "repos/${owner_repo}/issues/${pr}/comments?per_page=100" \
```
timothyfroehlich, line 144: **The `per_page=100` mitigation doesn't address the cost it's justifying.** The comment above says the function "went from one paginated call to two" and that `per_page=100` is the answer. It isn't: for a PR with fewer than 30 comments — the overwhelming majority — the old code already made exactly one request per endpoint, so raising the page size changes nothing while the request count still doubles.

Where that bites: `merge-pr.sh --automerge` re-runs `run_all_gates` every 30s for up to an hour (120 polls), and `pr-dashboard.sh` calls `_compute_review_state` once per open PR. Both now issue one extra request per invocation. Given the documented failure mode — a 403 reads as "no evidence", which degrades into gate failures rather than a recognisable error — this is the direction that matters, and the comment reads as if it were already handled.

Cheap fix if you want one: `pulls/{pr}/comments` only needs consulting when the marker lookup didn't already pin head.

scripts/workflow/_pr-gates.sh
```diff
@@ -36,61 +38,135 @@ _repo_slug() {
...
-           elif ($markers | length) > 0 then ($markers | last) + { state: "stale_marker" }
-           else { state: "unreviewed", sha: "", depth: "", at: "", summary: "" }
-           end
-         | [ .state, .sha, .depth, .at, .summary ] | @tsv'
+               summary: ($b | split("\n") | last | gsub("^\\s+|\\s+$"; "")) } ]')
+
+  # Grouped by the SHA reviewed, so a review that raised six findings is ONE piece of
+  # evidence rather than six — the summary then reports the count, which is the fact a
+  # reader of the handoff report wants.
+  comments=$(gh api --paginate "repos/${owner_repo}/pulls/${pr}/comments?per_page=100" \
```
timothyfroehlich, line 159: **A comment pins the commit GitHub anchored it to, not the commit the reviewer read.** `/code-review` reads the local working tree (`git diff @{upstream}...HEAD`), while a review comment posted without an explicit `commit_id` is anchored by GitHub at the PR's *current remote head*. Those are the same commit only when the reviewer's checkout is exactly level with origin.

Concrete failure: Tim's local branch is a commit or two behind origin (an agent pushed while he was reviewing, or he reviewed before pulling). `/code-review <depth> --comment <PR>` reads the older tree, posts its findings, GitHub stamps `original_commit_id` = the newer remote head — and this gate reports `commented` for a commit he never looked at. That is precisely the PR #1784 false-PASS class the SHA pin exists to prevent, arriving through the new evidence kind.

The marker path doesn't have this problem in the same way: `mark-claude-review.sh` is run deliberately by an agent that knows what was pushed, and the docs already tell it when re-attesting is legitimate. Nothing analogous guards the comment path, and the reviewer has no visibility into it. At minimum this deserves a line in the `_review_record` header alongside the `original_commit_id` / replies notes, since those two are documented at length and this one is the same class of hazard.

.claude/settings.json
timothyfroehlich: Documented rather than reverted — the widening is the point of PP-97tt, but you are right that it shipped undiscussed. scripts/workflow/AGENTS.md now carries a "Know what the allow rule costs" paragraph: the two changes landed together, one throwaway comment clears `unreviewed` with no human in the loop at that moment, the alternative (prompting per finding) makes `--comment` unusable for the batch it exists to post, and the residual control is that merging is human-only so a false clear still has to survive someone reading the handoff report. —Claude

.claude/settings.json
timothyfroehlich: Correct, and this was the real defect — thank you. Replaced the six pairs with three verb-anywhere globs (`Bash(gh api *DELETE*)` etc.), which cover both orders and the attached spellings, at the cost of also denying a read whose path contains one of the words (no such endpoint here). Case-sensitivity is still open: `-X delete` matches nothing, and denying lowercase substrings would start refusing ordinary reads. That gap is now stated in AGENTS.md rather than implied closed. —Claude

scripts/workflow/_pr-gates.sh
timothyfroehlich: Fixed. The comment now says the doubling is unmitigated, names where it bites (`--automerge`s 30s loop, once per open PR in pr-dashboard.sh), and demotes `per_page=100` to what it actually does — nothing below 30 comments, 4 requests into 1 on a busy PR. —Claude

scripts/workflow/_pr-gates.sh
timothyfroehlich: Real and now documented in the `original_commit_id` header. Nothing in the lookup can detect it (the field reads identically either way), so I recorded it as the markers one advantage over comments — mark-claude-review.sh stamps the SHA the attester names — plus the procedural mitigation, and pointed at why merge-handoff.sh prints the reviewed SHA and its distance from head rather than just "reviewed". —Claude

timothyfroehlich:
Claude review of head ebd9cb1 — `/code-review medium` — 4 findings: gh api deny globs missed the common flag order (fixed); per_page and original_commit_id comments overclaimed (corrected); allow-rule cost now documented

timothyfroehlich:
Closing unmerged — this implements the wrong half of the intention.

The gate this adds reads inline PR review comments, so a review that **found something** attests itself. But a clean `/code-review --comment` posts nothing, so it reads `unreviewed` and still needs `mark-claude-review.sh` — which is the case Tim actually wanted covered.

Second problem pointing the other way: the gate matches *any* top-level inline review comment on head, from anyone, for any reason, because nothing in a comment body marks which command produced it.

Not a quality call — the review at `medium` was clean and its 4 findings were addressed. The design just doesn't reach the case it was wanted for. Branch `worktree-feat/review-comment-gate` is left in place; PP-97tt keeps the requirement.

—Claude

PR #1820, opened 2026-08-03
docs(agents): cut the process corpus to what the scripts don't print (PP-22e4)

timothyfroehlich:
## Summary

Sweep task 5 of PP-22e4 — the four process skills. **1267 → 706 lines, net −604.**

Test 2 ("does a script, hook, or gate already do this?") dominated. A process skill's job is easy to confuse with narrating the script that already does the process.

| Skill | Before | After |
| --- | --- | --- |
| `pinpoint-pr-workflow` | 421 | 256 |
| `pinpoint-orchestrator` | 317 | 159 |
| `pinpoint-huddle` | 370 | 175 |
| `pinpoint-briefing` | 159 | 116 |

### What survived, and why

**`pr-workflow`** keeps Phase 3.4 whole — the review handoff, which pushes let you re-attest vs. needing a fresh `/code-review`, the trivial exception — plus the screenshot `--pages` gotchas and the Dependabot back-to-back lockfile casework. Nothing else in the corpus states any of those. Deleted: Phase 1's branch validation / pre-commit gate / push mechanics / E2E-selection matrix (all AGENTS.md §5, and the matrix was a *fourth* copy of the which-tests decision #1817 just made AGENTS.md-canonical), the merge-script flag reference, the status-token table, and the MCP call fences.

Kept the conventional-commit convention **and said why it's a convention**: verified there is no commit-msg hook and no commitlint, so nothing enforces it.

**`orchestrator`** keeps both upstream-bug blockquotes (CLAUDE.md points here for the technical record), the coordinator-not-implementer boundary, the lead review backstop, and the over-scoping casework. The inline agent-prompt template duplicated `references/agent-prompt-template.md` — which the same file already linked — so the copy went.

**`huddle`** had ~85 lines narrating hook output the agent is looking at while it reads them. Verified against `huddle-session-start.sh`: the notices print the bootstrap command, the rotation dispatch instruction, and the registration block *including* the naming format and worked examples. Kept what the hooks don't print — post-etiquette, auto-posted vs. manual (added scope is the most-skipped and most-needed), the em-dash self-filter, session_id ownership (PP-788v), the subagent registration refusal — and the rotation dispatch, because the hook explicitly routes here for it.

**`briefing`** keeps both audit-reading decisions verbatim: no `pnpm outdated` (Dependabot's soak is the supply-chain protection) and no `pipefail` on group B. A wrong reading of either produces a phantom finding.

### Deletions with no survivor

- `references/monitoring-commands.md` — zero inbound links, entirely command listings, and its "Batch PR Check" hand-rolls the `gh` polling loop we have a standing rule against. Recency check: last touched by #1508, not by this sweep.

### Pointer repairs

Three claims would have been stranded:

- `scripts/workflow/AGENTS.md` twice said the skill documents what to do per status token — that table is gone, and its own table already has an Action column.
- `docs/runbooks/gha-flake-log.md` cited "Phase 4.3 escape hatch", already off by one before this PR; now 4.2.
- `scripts/workflow/AGENTS.md` used `.agent/skills/` (singular) — same typo class #1817 fixed elsewhere.

**Phase 3 keeps its original 3.1–3.6 numbering on purpose.** Seven inbound pointers across AGENTS.md, CLAUDE.md, REVIEW.md, `ui-screenshot-reminder.cjs`, `pinpoint-superpowers-bridge` and `pinpoint-orchestrator` cite 3.4 and 3.6 by number. Renumbering saved nothing and would have broken all seven.

## Test Plan

- [x] `pnpm run check` green
- [x] `pnpm run check:python` green (498 passed)
- [x] Inbound-link check (`rg --hidden`) on every deleted file and section
- [x] Verified against source: no commit-msg hook / commitlint; `huddle-session-start.sh` notice bodies; `references/agent-prompt-template.md` is a superset of the deleted inline copy; `scripts/workflow/AGENTS.md` token table has an Action column
- [ ] Needs `/code-review` — this is the largest task in the sweep

## Related Issues

PP-22e4


timothyfroehlich:
## Review findings addressed (`/code-review high`)

All ten fixed in d7fdea3b. Four were the same class and the reviewer named it correctly: **facts deleted on the theory a script states them, where for those specific facts nothing did.**

| Finding | Verification | Fix |
| --- | --- | --- |
| Label stripped on FAIL undocumented | The merge script does it at lines 223-225; only its own header comment says so | Restored to 4.2, with the consequence — re-apply the label before re-handing the merge command |
| Huddle throttle/nudge knobs undiscoverable | `HUDDLE_NUDGE_SECONDS` / `HUDDLE_THROTTLE_SECONDS` appear only in hook source, `settings.json`, pytest — no hook prints them | Restored as a "Turning the volume down" section |
| Huddle snippets lost their caveats | `config.json` absent on a fresh clone; `today_bead.id` is a verified hint (PP-9lq5) | Both restored next to the snippets |
| `bd ready -n 50` in the wrong phase | The "default 10 hides work" caveat had migrated to Phase 5 | Back in Phase 1, where task selection happens |
| `scripts/workflow/AGENTS.md:112` circular | Line 78's claim was removed by this PR; 112 still advertised it | Rewritten to point at the table above it |
| Phase 4.2 hollow for the runbook | 4.2 said "log the flake first" without naming the script | Named `log-gha-flake.sh` + the runbook |
| `briefing-output.md` names scripts Step 1 doesn't run | Group A's mapping line was deleted | Restored, noting `orchestration-status.sh` wraps both |
| "Superset" claim overstated | Reference file had only the compressed form | Moved the re-attest-vs-fresh-review sentence into `references/agent-prompt-template.md` |
| Huddle intro factually wrong | The rotation notice points back here rather than carrying a command | Corrected — this was invented prose, the recurring failure mode in this sweep |

### On the description-cost finding

Confirmed, and worth recording rather than fixing. The four `description` fields now total **2,731 chars** (briefing 554, huddle 687, orchestrator 775, pr-workflow 715), up from roughly 1,000. All are under the 1,024 cap.

So the headline is more honestly: **−540 lines of load-on-demand body, +~1,700 chars of always-resident routing.** That's still strongly net-positive — a skill body costs nothing until loaded, and the sweep's premise is that these bodies were being loaded to re-read things the reader already had. But it isn't free, and task 9 of PP-22e4 is a description audit across all 15 skills, which is the right place to weigh the whole always-on tier at once.

timothyfroehlich:
Claude review of head d7fdea3 — `/code-review high` — 10 findings, all fixed: 4 facts deleted with no live survivor restored (ready-for-review stripped on gate FAIL, huddle throttle/nudge knobs, huddle config.json + today_bead.id caveats, bd ready -n 50 in Phase 1); 3 pointer repairs incl. a circular skill<->scripts/workflow/AGENTS.md loop this PR created; corrected my own false claim that all three huddle notices carry their command; description-cost tradeoff recorded in a PR comment

PR #1822, opened 2026-08-03
docs(agents): delete the two UI reference files the rules already state (PP-22e4)

timothyfroehlich:
Sweep task 6 of PP-22e4. Two `pinpoint-ui` reference files, 63% and 78% fenced code, deleted outright: **−298 lines, +6.**

PR #1793 simplified `pinpoint-ui/SKILL.md` and `styling-and-shadcn.md` but did not reach these two.

## Where each deleted section already lives

| Deleted | Already stated by |
| --- | --- |
| The anti-pattern catalog (email `type`, confirm-field `autocomplete`, bare `animate-*`, `<div role="button">`, `title` tooltip, non-LCP `priority`) | `docs/NON_NEGOTIABLES.md` — CORE-FORM-001/002, CORE-A11Y-002/004/005, CORE-PERF-003, each with a fuller Do/Don't than the BAD/GOOD pair here |
| `<Label htmlFor>` on the Radix wrapper | `references/accessibility.md` § Label-to-control association — as prose, with the "no error, no lint failure, visually identical render" consequence this version omitted |
| Page-layout + responsive-grid code blocks | `references/key-files.md` § Layout and CORE-RESP-001/003 |
| `window.innerWidth` / `sm:` structural prose | CORE-RESP-002 (with its two sanctioned exceptions) and CORE-RESP-003 |
| Global CSS resets, hardcoded spacing, inline styles | Generic teaching examples — test 3 |
| Direct-Server-Action and `useActionState` form shapes | CORE-ARCH-005, whose **Rationale** field carries the reason inline wrappers break: Next.js can't serialize a client-side wrapper |
| Dropdown Server Actions | `pinpoint-design-bible` §17 states the *why* more accurately: "Radix closes the dropdown before the form submits" |
| The Radix Select carve-out paragraph | `SKILL.md` § Server Action Forms — the compressed copy pointed at it already |

## The two facts with no survivor

Test 2 in the other direction, which is where the last two reviews found real problems: if nothing else states it, it should not have been deleted.

- **"MainLayout wraps the route — don't re-add it here"** → moved verbatim into `references/key-files.md` § Layout, next to the composition chain it qualifies. Verified before moving: only `src/app/(app)/layout.tsx`, `(site)`, and `(dev)` render `<MainLayout>` — no page does. Without it, key-files' "Every authenticated page should compose `<MainLayout>` → …" reads as an instruction to add one.
- **The dropdown rationale** → `SKILL.md` rule 5 now points at design-bible §17 rather than restating it.

`SKILL.md` also loses two rows from its reference table, repoints its layout sentence, and picks up the CORE-ARCH-005 citation on rule 4.

## Verification

- `rg --hidden` for both filenames repo-wide: only the frozen sweep spec/plan mention them.
- `pnpm run check` green.
- Checked, not assumed: CORE-ARCH-005's Rationale field, CORE-RESP-002's exception list, CORE-PERF-003's `sizes` requirement, and design-bible §17 line 72 all say at least as much as what was cut.

Bead: PP-22e4. Previous: #1817 (testing), #1820 (process).


timothyfroehlich:
Claude review of head 0683cb1 — `/code-review medium` — 2 low: rule 5 pointed at design-bible §17 instead of citing CORE-ARCH-006; CORE-ARCH-005's Don't did not cover the onSubmit+fetch shape the deleted file banned. Both fixed in 0683cb1b.

PR #1824, opened 2026-08-04
fix(worktrees): label Supabase's staged secrets for SELinux hosts (PP-9mg0)

timothyfroehlich:
Fixes PP-9mg0 — `supabase start` has been failing on the Bazzite host, taking `db:fast-reset`, `test:integration:supabase` and smoke E2E with it. The Supabase-dependent half of `preflight` was dead locally.

## Root cause

Not podman, and not an image regression — a Supabase CLI change meeting SELinux.

Older CLI versions wrote the pgsodium root key from *inside* the db container. CLI 2.111.0 stages it on the host at `supabase/.temp/start-secrets/<container>/secret-0` and bind-mounts it read-only at `/etc/postgresql-custom/pgsodium_root.key` — with **no `z`/`Z` relabel**.

The staged file inherits `user_home_t` from the repo; the container runs as `container_t`, so SELinux denies the stat. That makes `pgsodium_getkey.sh`'s `[[ ! -f "$KEY_FILE" ]]` **true** (`test -f` is false on EACCES), so it tries to *create* the key — writing through a read-only mount. Both halves of the log fall out of that:

```
pgsodium_getkey.sh: .../pgsodium_root.key: Read-only file system
FATAL:  invalid secret key
```

Kong and pooler stage their secrets the same way; db just fails first.

Evidence: the two postgres images (`17.6.1.143`, `17.6.1.156`) have byte-identical keygen scripts and a writable `/etc/postgresql-custom`, and the still-present container from before the CLI bump has `cat <<'EOF' > /etc/postgresql-custom/pgsodium_root.key` in its entrypoint — no host mount, no SELinux involvement.

## Fix

Label `start-secrets/` `container_file_t:s0` once. The directory persists across starts (the CLI only removes the per-container subdirectory inside it) and new files inherit their parent's label, so everything staged later is readable — including for a `supabase start` typed by hand, which an env var like `CONTAINERS_CONF_OVERRIDE` could not reach. It is also narrower than disabling container labeling wholesale: only these staged secrets get a shared label, and the containers stay confined.

- `scripts/selinux-label-supabase-secrets.sh` — idempotent, best-effort, no-op on macOS and CI. Carries the full root-cause note in its header.
- Runs from `worktree_setup.py` (every post-checkout) and from `db:_restart`.
- `dev-status.sh` names a wrong label when Supabase isn't up — the CLI's own error says nothing about it.

## Verification (on the Bazzite host)

- `supabase start` in a fresh worktree: all containers healthy.
- Staged db and kong secrets came out `container_file_t:s0` **without being relabeled directly** — inheritance works in the real flow.
- Key readable inside the db container (64 bytes).
- `pnpm run db:migrate` + `db:fast-reset` seed clean.
- `pnpm run test:integration:supabase` — 104/104 passing.
- 6 new pytest cases cover both no-op branches, labeling, `:Z`-residue relabeling, idempotence, and the best-effort failure path. `pnpm run check` and `check:python` (531 tests) green.

### Full `preflight` on Bazzite (added 2026-08-09, after merging main)

The bead's last open acceptance criterion was "preflight runs end to end locally", which the original verification pass skipped. Ran the whole `preflight:unlocked` sequence on the Bazzite host via crabbox, on this branch merged up to main:

| Step | Result |
| --- | --- |
| `typecheck`, `lint:local` | ✅ |
| unit | ✅ 2077 (224 files) |
| `db:fast-reset` | ✅ connects and seeds |
| `next build` | ✅ |
| `test:integration` (PGlite) | ✅ 573 (55 files) |
| `test:integration:supabase` | ✅ 104 (14 files) |
| smoke E2E (chromium, Mobile Chrome, firefox) | ✅ 200 passed, 2 flaky-on-retry, 8 skipped |

Two runner-environment notes, neither a defect in this change:

- **`check:config` cannot run on the crabbox runner.** Its lease checkout is a standalone clone, so `worktree_setup.py` early-returns and never generates `supabase/config.toml`. Skipped; it is a static config-drift check with no bearing on this fix.
- **Playwright browser binaries were stale** after main's Playwright bump, so smoke died in global setup before reaching Supabase. Installed them on the runner and re-ran. Filed as **PP-gprj**.

## Follow-up

Worth reporting upstream: the CLI's `secretFiles` binds should carry `:ro,z` on SELinux hosts. Not filed yet.


scripts/selinux-label-supabase-secrets.sh
```diff
@@ -0,0 +1,61 @@
...
+
+# Make the Supabase CLI's staged secret files readable by its own containers on
+# an SELinux-enforcing host (Bazzite / Fedora Atomic). No-op everywhere else --
+# macOS, CI, and any Linux box without SELinux exit immediately.
+#
+# Why this exists (PP-9mg0):
+#
+# Older CLI versions wrote the pgsodium root key from *inside* the db container
+# (`cat <<'EOF' > /etc/postgresql-custom/pgsodium_root.key` in the entrypoint).
+# CLI 2.111.0 stages it on the host instead, at
```
timothyfroehlich, line 12: **The mechanism this works around is gone in the CLI version the repo pins.**

Host-staging + `:ro` bind-mount of `secretFiles` is 2.111.0-only. I diffed the tags:

- `v2.111.0` `legacy/commands/start/lib/container-lifecycle.ts:588` — `const dir = join(workdir, "supabase", ".temp", "start-secrets", containerName)` then appends `<tempHostPath>:<containerPath>:ro` to the container's binds. Exactly the failure you describe.
- `v2.112.0`/`v2.113.0` `legacy/shared/db-bootstrap/container-lifecycle.ts` — `legacyCopyStartSecretFileIntoContainer` writes the content to a short-lived `os.tmpdir()` file and `docker cp`s it into the created-not-started container, deleting the temp file immediately. The doc comment on `secretFiles` in `docker-create-args.ts` says it outright: "never a host bind mount" (changed for remote-daemon support, supabase/cli#6022).

CI pins `2.113.0` (`.github/workflows/ci.yml:504`) and the Mac has `2.113.0`. At that version the only remaining consumer of `supabase/.temp/start-secrets/` is the edge_runtime container (`shared/functions/serve.ts:1526`), and `supabase/config.toml.template:142` has `[edge_runtime] enabled = false` — so nothing in `supabase start` touches that directory at all.

Concrete consequence: on a 2.113.0 Bazzite host this script creates and labels a directory that is never used; if `supabase start` still fails there, the labeling will not fix it and the added `db:_restart` step / dev-status hint will point at the wrong cause. If the failure reproduced on Bazzite, the CLI installed there is almost certainly older than the pin — bumping it is the actual fix. Worth confirming `supabase --version` on the host where this reproduced before landing.

—Claude

scripts/worktree_setup.py
```diff
@@ -784,6 +807,8 @@ def main() -> None:
 
     generate_launch_json(worktree_path, port_config)
 
+    label_supabase_secrets(worktree_path)
```
timothyfroehlich, line 810: **This never runs for the main worktree.** `main()` returns at line 754-755 when `Path.cwd().resolve() == main_wt.resolve()`, long before this call. So on an SELinux host the primary clone — which does run the stack, on the default 3000/54321/54322 ports (AGENTS.md §4 "Ports") — only ever gets labeled if someone runs `pnpm run db:reset` (the only caller of `db:_restart`). A plain `supabase start` there, which AGENTS.md explicitly tells agents to do self-service, stays unlabeled.

Secondary: the post-checkout hook only invokes this script when `$3 == 1` (branch checkout), so `git pull` on an already-checked-out worktree doesn't re-run it either — fine for the persistent-label design, but it means "every post-checkout" in the AGENTS.md bullet overstates the coverage.

—Claude

AGENTS.md
```diff
@@ -67,6 +67,7 @@ Each git worktree gets isolated Supabase ports automatically. The Husky `post-ch
...
 - **Config**: edit `supabase/config.toml.template`, not the generated file (which is chmod 444).
+- **SELinux hosts (Bazzite)**: since CLI 2.111.0 the pgsodium root key is staged on the host under `supabase/.temp/start-secrets/` and bind-mounted read-only into the db container — with no `z`/`Z` relabel, so an SELinux-enforcing host denies the container the read and Postgres dies on `FATAL: invalid secret key`. `scripts/selinux-label-supabase-secrets.sh` labels that directory `container_file_t:s0` once; everything the CLI later stages inside it inherits the label. It runs from `worktree_setup.py` (so every post-checkout) and from `db:_restart`, and is a no-op on macOS and CI. Run it by hand in a worktree that predates it, or if `pnpm run dev:status` says the label is wrong. (PP-9mg0.)
```
timothyfroehlich, line 70: Two claims here don't hold:

1. "since CLI 2.111.0 the pgsodium root key is staged on the host ... and bind-mounted read-only" — true of 2.111.0 only. 2.112.0 replaced it with `docker cp` from an `os.tmpdir()` temp file (supabase/cli#6022; the `secretFiles` doc comment in 2.113.0 says "never a host bind mount"). The repo pins 2.113.0 in CI. Reading this bullet later, someone will believe the current CLI still bind-mounts the key.
2. "It runs from `worktree_setup.py` (so every post-checkout)" — `worktree_setup.main()` returns early for the main worktree, so the primary clone is never labeled by the hook; only `db:_restart` covers it there. And the hook only fires on branch checkouts (`$3 == 1`), not every checkout.

—Claude

scripts/selinux-label-supabase-secrets.sh
timothyfroehlich: Confirmed, and this closes the PR rather than amending it. I verified it on the host rather than from the CLI source:

| Check on Bazzite | Result |
| --- | --- |
| CLI version | 2.113.0 |
| Supabase stack | healthy |
| `supabase/.temp/start-secrets` | does not exist |
| Host bind mounts into the repo, any container | none — db has only its data volume |
| This script in that worktree | absent, so nothing labeled anything |

#1837 bumped the pin to 2.113.0 earlier today, which fixed PP-9mg0 before this code could land. The script would label a directory nothing creates or reads.

Tim's call was to keep the record and drop the code: #1845 lands the diagnosis in the `pinpoint-deployment` skill — symptom, mechanism, why 2.112.0+ ended it, and the one-line `chcon` if an older CLI brings it back. Good catch; this would have been dead code running on every post-checkout. —Claude

scripts/worktree_setup.py
timothyfroehlich: Correct on both counts, and moot now — the PR is closed (see the reply on the script finding; the bind mount this works around is gone at the pinned CLI). Worth recording that you found it, though: it means the main worktree was only ever covered via `pnpm run db:reset`, so the "every post-checkout" claim was wrong in two directions, not one. —Claude

AGENTS.md
timothyfroehlich: Both wrong, agreed, and the bullet is not landing — the PR is closed. The replacement text in #1845 scopes the claim to 2.111.0 explicitly and says 2.112.0+ removed the bind mount, which is the part that dates. —Claude

timothyfroehlich:
Closed in favor of #1845 — the bug this fixes was already fixed by the CLI pin bump to 2.113.0 (#1837).

`/code-review medium` found that the host bind mount the script works around exists only in CLI 2.111.0; 2.112.0 replaced it with `docker cp`. Verified on Bazzite at 2.113.0: stack healthy, `supabase/.temp/start-secrets` never created, no container holding a host bind mount into the repo, and this script not present in that worktree. It would label a directory nothing reads.

#1845 keeps what still has value — the diagnosis, in `pinpoint-deployment`. The cause was a CLI version, so an older CLI brings it back, and the error message ("Read-only file system" + "FATAL: invalid secret key") names neither SELinux nor the CLI.

PR #1825, opened 2026-08-04
fix(pinballmap): guard the credentials RPC, harden the FORCE_PRODUCTION flags (PP-rnup)

timothyfroehlich:
Fast follow on #1818 (PP-o355.30), from a post-merge agy review. Three items; all three verified against source before acting.

## 1. HIGH — `get_pinballmap_credentials()` had no in-body role guard

`0061` created a `SECURITY DEFINER` function returning a **decrypted Vault secret** — the PinballMap outbound write token — protected only by `REVOKE`/`GRANT`. Its header cites `0028_natural_vengeance.sql` as the model. That is the **unhardened** version: `0029_discord_config_role_check.sql` exists *specifically because* REVOKE-only was not enough. Supabase re-grants `EXECUTE` on `public.*` functions to `authenticated` at connection time, and PostgREST exposes every public function as `POST /rest/v1/rpc/<name>`. `0057_superb_switch.sql` — the other PinballMap token RPC — already carries the guard and its comment says the `auth.role()` check "is the actual gate".

`0062` adds that check via `CREATE OR REPLACE`. **`0061` is left untouched** (it is applied history) — the repo's precedent for exactly this case is `0029`, which hardened `0028` and left `0028`'s text alone.

`0057`'s `get_pinballmap_api_token()` turned out **not** to be a live reference implementation: it was dropped in `0059` when `api-token.ts` moved to `process.env`. `get_pinballmap_credentials()` is now the only credential RPC in the schema, so the new `src/test/integration/supabase/pinballmap-credentials-rpc.test.ts` covers it alone: anon refused, an authenticated member refused, a member who **has** been granted `EXECUTE` still refused, `service_role` allowed.

That third case is the only one that proves anything about `0062`. With the `REVOKE` intact Postgres refuses on privilege before the body runs, so the first two pass against the unhardened function too — which is precisely the false comfort that let `0028` and `0061` ship without a guard. Isolating the body check means `GRANT EXECUTE ... TO authenticated` over a direct pg connection first (supabase-js cannot run DDL), then revoking in a `finally`.

## 2. MEDIUM — the `FORCE_PRODUCTION` flags were enabled by any truthy string

All three prod opt-ins read `Boolean(process.env[...])`. `Boolean("0")` and `Boolean("false")` are both `true`, so the two spellings an operator reaches for to turn a flag **off** in fact turned the bypass **on**. On `seed-pinballmap-creds.mjs` that writes a dev operator token into prod's Vault.

The bead said "follow the existing `MARK_MIGRATION_FORCE_PRODUCTION` convention" — but the existing convention *is* the defect, so fixing only the PinballMap script would have left two different `FORCE_*` semantics in the repo. All three (`DRIZZLE_`, `MARK_MIGRATION_`, `SEED_PINBALLMAP_CREDS_`) now route through a shared `isForceProductionEnabled()` in `scripts/lib/db-target.mjs` that accepts only `1`/`true`. Exactly three call sites, so Rule of Three is satisfied rather than anticipated. Failure direction is a refusal that names the token, never an unintended prod write.

## 3. Unlist deleted nothing when the stored lmx had drifted

Reported as "needs judgement, reachability not established". **It is reachable — and the suggested fix does not close it.**

PBM re-mints a title's lmx after a delete + re-add outside its resurrection window. The hourly sync puts the live id in the snapshot while the machine row keeps the dead one — the `lmx_drifted` state auto-link heals as `reconnected`. If a human clicks Unlist before that heal runs:

1. `removeMachine({lmxId: <stale>})` → PBM answers `not_found`
2. the action proceeds (that branch exists for the genuinely-already-gone case)
3. `withLmxRemoved(snapshot, <stale>)` is a no-op — the snapshot holds the *live* id
4. next auto-link pass sees the title still on the stored lineup and **re-lists the cabinet**

The human's unlist silently un-happens, **and the machine never actually left PinballMap**. `auto-link.ts:23-31` predicted exactly this scenario as the thing PP-o355.21 had to decide.

Adding the `machineId` filter to `withLmxRemoved` (the bead's suggestion) buys ≤1 hour: the next cron refresh re-fetches the live row from PBM and auto-link re-lists anyway.

So the fix is at the source — the unlist action resolves the **live** lmx from the snapshot by title (the same `findLmxForMachine` lookup auto-link uses, so the two agree on which row is live) and deletes that, falling back to the stored id when the title is absent from the lineup entirely. The timeline event records what was actually deleted. `withLmxRemoved` *also* gains the `machineId` filter `withLmxAdded` has, which now closes a real race rather than papering over one: the snapshot is re-read `FOR UPDATE` inside the transaction and could drift between the two reads.

## Verification

| Gate | Result |
| --- | --- |
| `pnpm run check` | ✅ |
| `pnpm run test` (unit) | ✅ 2074 |
| `pnpm run test:integration` (PGlite) | ✅ 574 |
| `pnpm run test:integration:supabase` (real stack, Bazzite) | ✅ 108 / 15 files |

The drift test is load-bearing — reverted to the pre-fix action it fails with `expected [ { id: 777, machineId: 7 } ] to deeply equal []`.

**Now verified against a real Supabase stack** (2026-08-09, after merging main). The original pass could not apply `0062` or run the RPC test — local Supabase would not start, **PP-9mg0**. Ran it on the Bazzite host via crabbox instead, after a full `db:reset` so `0062` was actually applied: **108 tests / 15 files**, up from main's **104 / 14**. The four new guard tests ran and passed, including the granted-`EXECUTE` case.

One flake worth noting, unrelated to this diff: the first unit run emitted an unhandled `dispatchEvent` error from `multi-select.test.tsx` (Radix focus-scope timer firing after jsdom teardown). It did not recur on a re-run and that file passes alone.

## Also: the rule, written down

Acceptance asked for `0061`'s stale `0028` citation to be corrected "so the next person does not mirror the unhardened version a third time". Editing an applied migration to fix a comment is not how this repo works — `0029` didn't touch `0028` — so the goal is met somewhere more durable: a new **`SECURITY DEFINER` RPCs that return a secret** section in the `pinpoint-security` skill.

It states the rule, says copy `0029`/`0062` and never `0028`/`0061` (and why those two still read wrong), and records the testing trap above. Grepped first: that rule appeared nowhere — not `NON_NEGOTIABLES.md`, not `SECURITY.md`, not the skill. Twice-relearned, never written down.

Closes PP-rnup.

https://claude.ai/code/session_01NqwhT5XVhknvki5jHUKTkV


timothyfroehlich:
**CI green** on `3f184b22` — all jobs pass, including `Supabase Integration Tests` and `Verify Migrations`, which are the two that could not run locally (PP-9mg0 blocks local Supabase on Bazzite). Migration `0062` applies cleanly and the guard test passes against a real stack.

Two corrections landed in `3f184b22` after the first CI round, both worth flagging for review:

1. **The test covered a function that no longer exists.** It parameterized over `get_pinballmap_api_token()` as well — that RPC was **dropped in `0059_overrated_inhumans.sql`** when `api-token.ts` moved to reading `process.env`. The bead cited it as a live reference implementation; it hasn't existed for three migrations. `get_pinballmap_credentials()` passed all three of its cases in that same run, so the `0062` guard was sound from the start.

2. **The assertions weren't testing the guard.** With the `REVOKE` intact, Postgres refuses on privilege *before the function body runs* — so every assertion would have passed against `0061`'s unhardened function too. That is exactly the false comfort that let `0028` and `0061` ship without an in-body check, and I nearly shipped a test that reproduced it.

   The new case isolates it: `GRANT EXECUTE` to `authenticated` over a direct pg connection — reproducing the Supabase connection-time re-grant that *is* the threat model — then asserts the member is still refused with `permission denied`. It can only pass if `auth.role()` inside the body fires. Restored in a `finally`, plus a defensive `REVOKE` in `afterAll`.

The first run's `E2E Full` failure (`machine-owner-picker.spec.ts` + 2 flaky) was unrelated flake — that job passed this round with no changes to any E2E surface.

Ready for review.

src/app/(app)/m/pinballmap-actions.ts
```diff
@@ -528,13 +528,29 @@ export async function unlistMachineFromPinballMapAction(
...
+  // resurrection window) our stored id goes stale — the `lmx_drifted` state
+  // auto-link heals as `reconnected`. Unlisting on the stale id deletes nothing:
+  // PBM answers `not_found`, we clear the local flag, and the title is still on
+  // the public lineup, so the next auto-link pass re-lists the machine. The
+  // human's unlist silently un-happens and the cabinet never left PinballMap.
+  // Resolving through the snapshot by title is the same lookup auto-link uses,
+  // so the two agree on which row is live. Falls back to the stored id when the
+  // title is absent from the lineup entirely — that IS the already-gone case the
+  // `not_found` branch below exists to finish.
+  const liveLmxId =
```
timothyfroehlich, line 641: The live-lmx resolution only closes the drift window in one direction — when the stored handle is stale **and** the snapshot is already fresh. The mirror case is just as reachable and still fails exactly the way the PR description says it must not:

1. 12:00 hourly sync stores `{id: 777, machineId: T}`; the machine row also holds 777, so nothing looks drifted.
2. 12:10 someone deletes + re-adds the title on pinballmap.com; PBM re-mints it as 888.
3. 12:20 a human clicks Unlist. `findLmxForMachine` returns the **stale** 777 (next sync is 13:00), so `liveLmxId === storedLmxId === 777`.
4. `removeMachine` answers `not_found`, the action takes the already-gone branch, clears the local columns, reports success, and drops the title from the stored snapshot.
5. 13:00 sync re-fetches 888, `resolveAutoLink` sees the title on the lineup and re-lists the cabinet.

Net result: the unlist silently un-happens and the machine never left PinballMap — the same outcome the drift fix is aimed at. Resolving from the snapshot narrows the window rather than closing it, because the snapshot is only trustworthy up to one hour. The `not_found` reply is the signal that the handle we used is not what PBM currently has, so concluding "already gone" from it is the step that is unsound; a forced re-sync + re-resolve (throttle permitting) before taking that branch would cover both directions.

src/lib/pinballmap/snapshot-edit.ts
```diff
@@ -51,11 +51,29 @@ export function withLmxAdded(
...
+ * already gone.
+ *
+ * Drops the title's row by `machineId`, not just the `id` we deleted — the same
+ * asymmetry-closing rule `withLmxAdded` applies, and for the same reason: PBM
+ * gives a title exactly one lmx at our location, and every consumer resolves it
+ * by `machineId`. Filtering on `id` alone left a re-minted row behind, which
+ * `resolveAutoLink` reads as "the title is still on the lineup" and re-lists
+ * within the hour — silently undoing a human unlist (PP-rnup).
+ *
+ * `lmxId` is still matched first so this stays a no-op in the ordinary case
```
timothyfroehlich: This sentence does not describe the code below it. The filter is a plain OR removal — both predicates are evaluated for every row, nothing is "matched first" — and in the ordinary case (caller handle == stored row) the function *removes* the row, which is the opposite of a no-op. A reader who trusts the doc will assume `lmxId` is authoritative and `machineId` is only a fallback.

The behaviour the doc should state is the one that actually matters at the call site: every row carrying `pinballmapMachineId` is dropped, regardless of its `id`. So if the stored snapshot ever holds two rows for one title (PBM anomaly, or a snapshot written from two overlapping sources), both disappear locally while only one was deleted on PBM, and the snapshot disagrees with the lineup until the next hourly sync.

src/test/integration/supabase/pinballmap-credentials-rpc.test.ts
```diff
@@ -0,0 +1,126 @@
...
+      email: memberEmail,
+      password: "TestPassword123",
+    });
+  });
+
+  afterAll(async () => {
+    await adminClient.auth.admin.deleteUser(memberUser.id);
+    // Defensive: the grant test restores this itself, but a failure mid-test
+    // must not leave `authenticated` holding EXECUTE for the rest of the run.
+    await sql`REVOKE ALL ON FUNCTION public.get_pinballmap_credentials() FROM anon, authenticated`;
```
timothyfroehlich: Ordering hazard: `adminClient.auth.admin.deleteUser(memberUser.id)` runs first, and `memberUser` is only assigned at the end of `beforeAll`. If `createUser` fails or `signInWithPassword` throws, `memberUser` is `undefined`, this hook throws a TypeError on `.id`, and neither the defensive REVOKE nor `sql.end()` runs — so the postgres connection stays open (vitest blocks until the teardown timeout) and, if the grant test was where things died, `authenticated` keeps EXECUTE on the credential RPC for the rest of the run. Guard the delete (`if (memberUser)`) and put `sql.end()` in its own `finally`.

src/test/integration/supabase/pinballmap-credentials-rpc.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+if (!supabaseUrl || !supabaseAnonKey || !serviceRoleKey || !databaseUrl) {
+  throw new Error("Missing Supabase env vars for PinballMap credential tests.");
+}
+
+const adminClient = createClient(supabaseUrl, serviceRoleKey);
+const anonClient = createClient(supabaseUrl, supabaseAnonKey);
+// Needed to hand `authenticated` the EXECUTE privilege the guard is supposed to
+// survive — supabase-js cannot run DDL.
+const sql = postgres(databaseUrl);
```
timothyfroehlich: This porsager client is built without `prepare: false`. `databaseUrl` falls back to `POSTGRES_URL`, which in any pooled environment is the Supavisor **transaction** pooler on `:6543`, where prepared statements are unsupported — the exact configuration AGENTS.md §7 / PP-d8l8 forbids, and the reason `scripts/lib/pg-client.mjs` applies `prepare:false` last so callers cannot re-enable it. Against a local direct connection this is harmless, but a run pointed at a pooled URL fails on the GRANT/REVOKE rather than on the assertion, which reads as a broken guard test.

src/app/(app)/m/pinballmap-actions.ts
timothyfroehlich: Confirmed and fixed in `7a01ff7e`. You are right that resolving from the snapshot narrows the window rather than closing it, and right about which step is unsound — `not_found` is evidence the handle is wrong, not evidence the entry is gone.

The action now re-fetches through `syncLocationSnapshot` (the sanctioned chokepoint, so CORE-PBM-001's manual throttle still applies) and re-resolves before believing a 404: absent from the fresh lineup → finish the unlist; present under a different id → delete that one; present under the id we tried → PBM contradicting itself, refuse; lineup unreachable → refuse rather than report an unlist we cannot confirm. The refresh is skipped when the stored snapshot is already newer than the throttle interval, since that is the freshest lineup we are allowed to fetch anyway.

The old comment's worry that refusing would strand the machine no longer holds: the genuinely-already-gone case resolves on the refresh instead of 404ing forever, so the only refusals left are transient or need a human anyway.

Three integration tests cover it, and I verified all three fail against the pre-fix action before keeping them. —Claude

src/lib/pinballmap/snapshot-edit.ts
timothyfroehlich: Fixed in `7a01ff7e`. The doc did describe a lookup-with-fallback the code does not implement. Rewritten to state the actual behaviour — an OR evaluated over every row, dropping every row carrying `pinballmapMachineId` regardless of its `id` — and it now names the two-rows-for-one-title consequence you raised, plus why that is the safer direction to err in (the opposite error re-lists a cabinet a human just withdrew). —Claude

src/test/integration/supabase/pinballmap-credentials-rpc.test.ts
timothyfroehlich: Fixed in `7a01ff7e`. Reordered worst-consequence-first so every step is reachable: the defensive REVOKE runs before the user delete, the delete is guarded on `memberUser`, `memberUser` is now typed `| undefined` so the compiler enforces the guard, and `sql.end()` is in a `finally`. —Claude

src/test/integration/supabase/pinballmap-credentials-rpc.test.ts
timothyfroehlich: Fixed in `7a01ff7e` — `prepare: false` added, with a comment citing AGENTS.md §7 / PP-d8l8. Agreed it is a repo-wide nit rather than a new class, so I left `timeline-invited-conversion.test.ts` alone as out of scope here. —Claude

timothyfroehlich:
Claude review of head 7a01ff7 — `/code-review medium` — 4 findings, all fixed in 7a01ff7e: remove-404 treated as 'already gone' without checking the live lineup (MEDIUM, re-fetch + re-resolve, 3 new tests); withLmxRemoved JSDoc described a lookup the code doesn't do; RPC test afterAll could skip its REVOKE and leak the pg connection; that test's pg client lacked prepare:false (PP-d8l8).

PR #1821, opened 2026-08-03
chore(deps): clear all 12 open pnpm advisories (PP-chqx)

timothyfroehlich:
`pnpm audit` was reporting **12 vulnerabilities — 4 high, 8 moderate** — across 6 packages, all transitive. Every one is fixed by raising an override floor in `pnpm-workspace.yaml`. No direct dependency changes, no source changes.

## What moved

| Package | Override | Advisory | Sev |
|---|---|---|---|
| `undici` | `^7.28.0` → `^7.29.0` | [GHSA-4cwx-7wf7-3272](https://github.com/advisories/GHSA-4cwx-7wf7-3272) + 4 moderates | high |
| `fast-uri` | `>=3.1.4` → `>=3.1.5` | [GHSA-7p8r-x3mc-p8w7](https://github.com/advisories/GHSA-7p8r-x3mc-p8w7) | high |
| `ip-address` | *(new)* `>=10.3.1` | [GHSA-mwp4-54f8-5fhr](https://github.com/advisories/GHSA-mwp4-54f8-5fhr) + 2 moderates | high |
| `brace-expansion` | `>=5.0.8` → `>=5.0.9` | [GHSA-rgw5-rvv9-x895](https://github.com/advisories/GHSA-rgw5-rvv9-x895) | high |
| `postcss` | `>=8.5.18` → `>=8.5.23` | [GHSA-fxqj-rqcc-2cmp](https://github.com/advisories/GHSA-fxqj-rqcc-2cmp) | moderate |
| `hono` | *(new)* `>=4.12.34` | [GHSA-8j4g-w8fx-2239](https://github.com/advisories/GHSA-8j4g-w8fx-2239) | moderate |

Resolved versions move by a patch each, except `ip-address` 10.2.0 → 10.4.0 (the `>=10.3.1` floor pulls current 10.x).

## The one worth reading closely

**`ip-address`** reaches us through `@modelcontextprotocol/sdk` and `mcp-handler` via `express-rate-limit`. `Address4` decodes leading-zero octets as decimal while resolvers decode them as octal — SSRF and trust-boundary bypass. The MCP server is the exposed surface, and the vulnerable code sits in its rate limiter's IP parsing.

`brace-expansion` 5.0.9 is notable for a different reason: 5.0.8 *was* the fix for CVE-2026-14257, and this advisory is a bypass of that mitigation.

## Tradeoff you should know about

[...]

PR #1832, opened 2026-08-07
chore(deps): clear the nanoid high and the deferred hono moderates (PP-dj1w)

timothyfroehlich:
`pnpm audit --audit-level=high` exits 1 on main as of 2b2b557a. The `pnpm audit` job on main's CI run [31226892212](https://github.com/timothyfroehlich/PinPoint/actions/runs/31226892212) is **red**, and it will redden every open PR the moment that PR merges main in. Two override floors in `pnpm-workspace.yaml` fix it. No direct dependency changes, no source changes.

## nanoid `>=3.3.17 <4` — the new one, and why main is red

GHSA-2v37-7h3g-55p8 (**high**): a custom alphabet generator loops indefinitely when called with size zero.

Dependabot opened this alert at `2026-08-07T23:21:56Z` — the same minute #1821 merged. It is **not a regression from #1821**; it landed on top of it. It reaches us through the `postcss >=8.5.23` floor that #1821 raised: postcss 8.5.25 depends on nanoid, which resolved to 3.3.16.

Resolves to **3.3.17** (published 2026-08-03), not the newer 3.3.18 (published 2026-08-07T16:41Z, roughly 7 hours old). 3.3.18 is still inside the minimum-release-age window and pnpm's soak filter excluded it unprompted — so this floor needs no `minimumReleaseAgeExclude` entry. The soak gate working is what makes this a clean bump instead of a bypass.

## hono `>=4.12.34 <5` — the item #1821 split out on purpose

GHSA-8j4g-w8fx-2239 (moderate) plus three more; four advisories in total.

#1821 dropped this override rather than land a standing soak-bypass grant, and tracked the remainder in **PP-dj1w**. hono 4.12.34 published `2026-08-03T02:30Z`, so the 24h window closed four days ago and the override now installs with no exemption. That window was the only thing PP-dj1w was waiting on.

Resolves to **4.13.0**, which published `2026-08-03T21:54Z` and has since aged past the same soak window that filtered it out during #1821.

## Both ranges are bounded

`<4` and `<5`, per the unbounded-override finding from #1821's review: pnpm applies overrides to peer ranges too, so an open-ended `>=` deletes the upstream compatibility ceiling. Confirmed the lockfile records `hono: '>=4.12.34 <5'` against `@hono/node-server` rather than an unbounded range, and that no `minimumReleaseAgeExclude` key was added.

## Result

Advisory count **5 → 0**. `pnpm audit` is clean at `--audit-level=low`, two levels stricter than the CI gate.

## Verification

- `pnpm audit --audit-level=low` — clean, exit 0
- `pnpm run check` — all checks passed
- `next build` — success

Two gates could not produce a usable local signal, both for reasons that predate this change:

- **Unit tests.** 41 tests across 5 files fail on this host with `localStorage` undefined. I verified this against **main's own lockfile** — same failures, before this change — and the same suite passes as a required job in CI on main. That is a local vitest environment problem on this machine, not a repo regression; I'm filing it separately. CI covers unit here.
- **Supabase integration + smoke E2E.** The Supabase Postgres container will not start on this host (PP-9mg0). CI covers both.

Closes PP-dj1w.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -15,6 +15,7 @@ overrides:
   yaml: '>=2.8.3'
   uuid: '>=14.0.0'
   postcss: '>=8.5.23'
+  nanoid: '>=3.3.17 <4'
   shell-quote: '>=1.8.4'
   ws: '>=8.21.0'
   vite: '>=8.0.16'
@@ -23,6 +24,7 @@ overrides:
   fast-uri: '>=3.1.5 <4'
   sharp: '>=0.35.0'
   '@hono/node-server': '>=2.0.5'
+  hono: '>=4.12.34 <5'
   ip-address: '>=10.3.1 <11'
 
 importers:
@@ -839,7 +841,7 @@ packages:
     resolution: {integrity: sha512-bjD221KPLoJTWUwso1J6fGKiTXEUFedG/s0visavY4zakFPkeGURMRNly+FhBHs7T8Dz4qHaZIMX9ZoJHSJtKA==}
     engines: {node: '>=20'}
     peerDependencies:
-      hono: ^4
+      hono: '>=4.12.34 <5'
 
   '@hookform/resolvers@5.4.0':
     resolution: {integrity: sha512-EIsqr/t/qbinPIhGjMdtvutIN1Kk4uwbROE9/UQ93CAVGR7GkA7Y92+fX80OzXi/OB67jVFYwKGO1WzkxmkFZw==}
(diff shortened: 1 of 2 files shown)


timothyfroehlich:
Claude review of head 2aff505 — `/code-review medium` — 0 findings — advisory ranges verified against both GHSA entries, no unintended consumers, lockfile internally consistent, peer-range rewrite is a narrowing; CI green incl. pnpm audit + full E2E.

PR #1833, opened 2026-08-07
feat(lint): require Server Action modules to be named so path globs find them (PP-22e4.4)

timothyfroehlich:
Prepares the `.claude/rules/` path-scoped tier landing in the next PR, and is useful on its own.

## The problem

Server Actions are deliberately scattered. A route-local action colocates with its route under `src/app/**` — the App Router convention, and the import graph confirms it: every colocated action is imported only from within its own feature subtree. A genuinely cross-cutting one lives in `src/server/actions/` (`avatar.ts`, `images.ts`, both imported from three unrelated places).

Collecting all 24 into one directory was considered and **rejected** — it would flatten a working convention and cost ~24 file moves plus 60+ import rewrites, to solve a problem four globs already solve.

The cost of keeping the split is that *"is this a Server Action module?"* is a file-**content** question — a `"use server"` directive — while the consumers that need to answer it can only match on **path**:

```
**/actions.ts   **/*-action.ts   **/*-actions.ts   src/server/actions/**
```

All 24 current action modules already match, so this rule is clean on `main` today. What it prevents is the silent case: a new action named `foo.ts` drops out of every path consumer and **nothing fails** — the rules just quietly stop loading. This makes that loud.

## Scope

Deliberately narrow:

- Constrains what an action module is **called**, never where it lives. Moving actions around stays a free choice.
- Matches only a module-level directive (`Program > ExpressionStatement[directive="use server"]`). An inline `"use server"` in a function body is a different construct owned by CORE-ARCH-005, and is not this rule's business.
- A `"use server"` in a comment or string literal is not a directive and never matches — three files under `src/app/(auth)` mention it in prose for exactly this reason, and each has a test.

## Verification

- `pnpm exec eslint src/` — clean (all 24 action modules conform).
- Negative control: a probe file `src/app/(app)/issues/zz-probe-mutations.ts` **is** flagged by the real `eslint.config.mjs`, not just by the test harness.
- 23 unit tests; `pnpm run check` green.

## Two harness gotchas, recorded in the test

Both bit during development, both are specific to a filename-driven rule, and neither applies to the sibling transaction rule:

1. **`files:` must be set.** A flat-config block with none matches only ESLint's default `**/*.{js,mjs,cjs}` — so every `.ts` fixture lints to zero messages and every assertion passes **vacuously**. The first version of this test "passed" 22/23 for exactly that reason.
2. **Fixture paths must be repo-relative.** Flat config resolves `files` against the config basePath; an absolute `/repo/...` path lands outside it and ESLint returns *"No matching configuration found"* instead of linting.

### DIFF
--- a/eslint-rules/server-action-file-naming.mjs
+++ b/eslint-rules/server-action-file-naming.mjs
@@ -1,7 +1,7 @@
 // ===== Server Action file naming =====
 //
 // A module whose directive prologue is `"use server"` must be named
-// `actions.ts(x)` or `<something>-action(s).ts(x)`, or live under
+// `actions.ts` or `<something>-action(s).ts`, or live under
 // `src/server/actions/`.
 //
 // ── Why a naming rule exists at all ──────────────────────────────────────────
@@ -44,8 +44,17 @@
 // literal is not a directive and never matches — three files in `src/app/(auth)`
 // mention it in prose for exactly this reason.
 
-/** Basename of a conforming action module: `actions.ts` or `*-action(s).ts`, `.tsx` too. */
-export const ACTION_FILENAME_PATTERN = /^(?:.*-)?actions?\.tsx?$/;
+// The pattern must accept EXACTLY what the four globs above match — no more.
+// Two near-misses this deliberately rejects, because accepting either would
+// pass a file the globs then drop, which is the failure this rule exists to
+// catch:
+//   - bare singular `action.ts` — `**/actions.ts` is plural, and the other two
+//     globs require the literal `-`.
+//   - any `.tsx` — every glob ends in `.ts`. A module that is nothing but
+//     Server Actions has no JSX, so `.tsx` is the wrong extension regardless;
+//     widening the globs to cover it would buy nothing.
+/** Basename of a conforming action module: `actions.ts` or `*-action(s).ts`. */
+export const ACTION_FILENAME_PATTERN = /^(?:actions|.*-actions?)\.ts$/;
 
 /** Shared (non-route-local) actions live here and are exempt from the basename rule. */
 export const SHARED_ACTIONS_DIR_PATTERN = /(?:^|[/\\])src[/\\]server[/\\]actions[/\\]/;
--- a/eslint-rules/server-action-file-naming.mjs
+++ b/eslint-rules/server-action-file-naming.mjs
@@ -1,0 +1,107 @@
+// ===== Server Action file naming =====
+//
+// A module whose directive prologue is `"use server"` must be named
+// `actions.ts(x)` or `<something>-action(s).ts(x)`, or live under
+// `src/server/actions/`.
+//
+// ── Why a naming rule exists at all ──────────────────────────────────────────
+// Server Actions are deliberately NOT collected into one directory: a
+// route-local action colocates with its route under `src/app/**` (that is the
+// App Router convention and the import graph confirms it — each colocated
+// action is imported only from within its own feature subtree), while a
+// genuinely cross-cutting action lives in `src/server/actions/`. That split is
+// worth keeping.
+//
+// The cost of keeping it is that "is this a Server Action module?" becomes a
+// file-CONTENT question, and several things that need to answer it can only
+// match on PATH:
+//
+//   - `.claude/rules/*.md` frontmatter (`paths:` globs) — the path-scoped rule
+//     tier that loads CORE-ARCH-005 / CORE-ARCH-008 / CORE-ARCH-011 / CORE-ARCH-012
+//     when an agent opens an action file. Globs cannot read a directive.
+//   - Any future CODEOWNERS entry, path filter in CI, or reviewer checklist
+//     keyed on actions.
+//
+// Those consumers all use the same four globs:
+//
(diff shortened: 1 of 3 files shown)


timothyfroehlich:
/audit-override nanoid GHSA-2v37-7h3g-55p8 is inherited from main (21 transitive paths under webpack/postcss); this PR does not touch pnpm-lock.yaml and #1832 carries the real fix. Every other check on this PR is green.

eslint-rules/server-action-file-naming.mjs
```diff
@@ -0,0 +1,107 @@
...
+// as a Server Action module, and only that is matched here
+// (`Program > ExpressionStatement[directive="use server"]`). An inline
+// `"use server"` inside a function body is a different construct with a
+// different rule (CORE-ARCH-005 bans it as a form-action wrapper) and is not
+// this rule's business. A `"use server"` appearing in a comment or a string
+// literal is not a directive and never matches — three files in `src/app/(auth)`
+// mention it in prose for exactly this reason.
+
+/** Basename of a conforming action module: `actions.ts` or `*-action(s).ts`, `.tsx` too. */
+export const ACTION_FILENAME_PATTERN = /^(?:.*-)?actions?\.tsx?$/;
```
timothyfroehlich: **The accepted set is wider than the glob set, in the exact way the rule exists to prevent.**

`actions?` makes the bare singular `action.ts` conforming — the optional `(?:.*-)?` group means no hyphen is required. But none of the four globs this rule is protecting matches it:

- `**/actions.ts` — plural only
- `**/*-action.ts` / `**/*-actions.ts` — require the literal `-`
- `src/server/actions/**` — different directory

Concrete scenario: someone adds `src/app/(app)/issues/action.ts` with a module-level `"use server"`. ESLint passes, and the file drops out of `.claude/rules/server-actions.md`'s `paths:` — the silent failure the PR description says this rule makes loud. Verified against the actual regex (`isConformingActionFilename("src/app/x/action.ts") === true`).

The singular form should only be reachable through the hyphenated shape (`export-action.ts`), which is what the globs allow:

```suggestion
export const ACTION_FILENAME_PATTERN = /^(?:actions|.*-actions?)\.tsx?$/;
```

(Leaves the `.tsx` question — see the separate comment — alone.)

—Claude

src/test/eslint/server-action-file-naming.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    ["src/components/issues/issue-actions.tsx", ".tsx"],
+    ["src/server/actions/avatar.ts", "shared actions dir, any basename"],
+  ])("stays silent on %s (%s)", (filename) => {
+    expect(violations(USE_SERVER_MODULE, filename)).toHaveLength(0);
+  });
+
+  it('stays silent on an off-pattern file with NO "use server" directive', () => {
+    const code = `import { db } from "~/server/db";
+
+export async function doThing(): Promise<void> {
```
timothyfroehlich, line 101: **This case asserts a gap rather than closing it.** `\.tsx?` accepts `actions.tsx` / `issue-actions.tsx`, but every glob in the set the rule protects ends in `.ts`:

```
**/actions.ts   **/*-action.ts   **/*-actions.ts   src/server/actions/**
```

So `src/components/issues/issue-actions.tsx` with a module-level `"use server"` lints clean *and* matches none of the `paths:` entries in `.claude/rules/server-actions.md` — same silent drop-out as an off-pattern name, just via the extension. All 24 action modules on `main` are `.ts`, so nothing is broken today; the rule just isn't holding the line it claims to.

Two valid directions, and this one's your call:

1. Reject `.tsx` — tighten to `\.ts$` and flip this fixture to expect 1 violation. A `"use server"` module can only export async functions, so `.tsx` buys nothing.
2. Keep `.tsx` and add `**/actions.tsx` / `**/*-action.tsx` / `**/*-actions.tsx` to the frontmatter in the rules-tier PR.

Either way the two sets should be identical — that identity is the rule's entire contract.

—Claude

eslint-rules/server-action-file-naming.mjs
timothyfroehlich: Fixed in 0cfe3f1e, taking the wider version of your suggestion: `/^(?:actions|.*-actions?)\.ts$/`. It closes this and the `.tsx` finding at once, so the accepted set and the matched set are now identical. Verified against all 8 distinct `"use server"` basenames in `src/` — every one still conforms. —Claude

src/test/eslint/server-action-file-naming.test.ts
timothyfroehlich: Fixed in 0cfe3f1e by taking the first of your two directions: the regex now rejects `.tsx` rather than the rules file gaining three `.tsx` globs. A module that is nothing but Server Actions has no JSX, so `.tsx` is the wrong extension regardless — widening the globs would buy coverage for a file that should not exist. Both near-misses now have a test. —Claude

timothyfroehlich:
Claude review of head 0cfe3f1 — `/code-review high` — 2 findings, both fixed: the filename regex accepted `action.ts` and `.tsx`, neither of which any protected glob matches

PR #1834, opened 2026-08-07
docs(agents): move the 20 non-negotiables into a path-scoped .claude/rules tier (PP-22e4.4)

timothyfroehlich:
Closes the last open piece of the PP-22e4 context-corpus sweep, with a design that replaces the original task-8 plan.

## What changed

`AGENTS.md` §2.1 restated all 20 implementation rules as one-line summaries, loaded on **every** session whatever the session was doing — the responsive-layout rules during a migration, the migration rule during a CSS change. They now live in `.claude/rules/`, where Claude Code loads a file by its frontmatter:

- no `paths:` → loaded at launch
- a `paths:` glob list → loaded on demand, the first time a matching file is read

Four rules load always (`always.md` — type safety, path aliases, Rule of Three, email privacy; no glob would narrow them). Sixteen are path-scoped across seven files. §2.1 becomes a pointer at the catalog.

## Why not the original plan

Task 8 routed the same rules through `.github/instructions/*.instructions.md`. **Those six files were deleted by #1801** (the Copilot review retirement) one day after the plan was written, so that tier had no source left. `.claude/rules/` is a better fit anyway: a rule file can carry the rule rather than a one-line summary, and there is one loading mechanism instead of two.

Tradeoff, taken deliberately: `.claude/rules/` is **Claude Code-only**. Codex, Gemini, and Antigravity lose the always-loaded index and must read the catalog. `.github/instructions/` had the same single-tool limit — the plan just never said so.

## What is and is not copied

`docs/NON_NEGOTIABLES.md` stays canonical. The rule files carry the §2.1 summaries **moved verbatim** plus a pointer — that is what keeps this from becoming a third copy of the catalog to drift. `scripts/check_rule_ids.py` already globbed `.claude/rules/*.md` (written ahead of the directory existing), so every `CORE-*` citation in the new files is validated from the first run.

## Grouping

Drawn by which glob set a rule needs, not by topic, so no file loads to deliver a rule that doesn't apply:

| File | Globs | Rules |
|---|---|---|
| `always.md` | *(none)* | 4 |
| `ui.md` | components / app / hooks / css | 5 |
| `server-actions.md` | action filename globs + permissions | 3 |
| `testing.md` | test / e2e / config | 3 |
| `database.md` | drizzle / db / migration scripts / supabase | 1 |
| `supabase-auth.md` | supabase seam, auth routes, `middleware.ts` | 1 |
| `pinballmap.md` | the PBM seam + vendored terms | 1 |
| `config-and-env.md` | `next.config.ts`, ENV_VARS, workflows | 2 |

`ui.md` is the broad one — roughly two-thirds of the source tree. That's deliberate rather than sloppy: it still excludes `src/lib`, `src/server`, tests, `scripts/`, and `drizzle/`, so a backend or migration session never loads it.

Every glob was checked against a real path first. Two drafts pointed at files that don't exist (`src/proxy.ts`, `src/middleware.ts` — the real one is `middleware.ts` at the repo root).

`server-actions.md` matches action modules by **filename**, because a `"use server"` directive isn't something a glob can read. #1833 adds the lint rule that keeps filenames and globs agreeing; the two are independent but that one is what stops this from rotting silently.

## The budget gate

`check_rule_ids.py` grows a fourth check: the always-loaded files must stay under **40 KiB**. Currently 38,085 bytes — ~2 KB headroom.

This is a budget on a measurable resource, not a prose-quality lint. It asks one question at the moment someone is adding always-loaded text: *does this need to be always-loaded, or does it have a path?* A rule file declaring `paths:` isn't counted, so moving text behind a glob is always the cheap fix.

**Honest accounting:** the immediate always-loaded saving is small — §2.1 was ~2.3 KB and the pointer plus `always.md` costs ~1.5 KB back. The win is that sixteen rules now arrive with more context than a one-liner, only when they apply, and that the budget stops §2.1 growing back. **If you want to cut something from this PR, the gate is the piece to cut** — the tier stands without it.

## Also refreshed

Four claims in `check_rule_ids.py` that this PR falsifies: `.claude/rules/` "does not exist yet"; §2.1 "is being replaced"; the orphan count (42 of 62 → 17 of 67, since the tier cites more of the catalog than the old list did); and the fragile-citation error message, which told the reader to look the ID up in "AGENTS.md §2.1's own list for now".

## Verification

`pnpm run check` green, `pnpm run check:python` green (509 passed, 11 new tests covering `declares_paths` and the budget). All 20 rules accounted for: 4 + 5 + 3 + 3 + 1 + 1 + 1 + 2 = 20.


timothyfroehlich:
/audit-override nanoid GHSA-2v37-7h3g-55p8 is inherited from main (21 transitive paths under webpack/postcss); this PR does not touch pnpm-lock.yaml and #1832 carries the real fix. Every other check on this PR is green.

scripts/check_rule_ids.py
```diff
@@ -142,6 +153,77 @@
...
+# non-negotiables here, including the responsive-layout rules during a
+# migration and the migration rule during a CSS change.
+#
+# Only files with no `paths:` frontmatter count. A rule file that declares
+# globs is on-demand and is deliberately NOT budgeted, so splitting a rule out
+# by path is always the cheap fix.
+ALWAYS_LOADED_SOURCES: tuple[str, ...] = (
+    "AGENTS.md",
+    "CLAUDE.md",
+    ".claude/rules/*.md",
```
timothyfroehlich: The always-loaded budget under-counts: `ALWAYS_LOADED_SOURCES` globs only `.claude/rules/*.md`, but Claude's rules loader walks `.claude/rules/` **recursively** (verified against the 2.1.226 bundle — the loader recurses into subdirectories, and any `.md` with no `paths:` frontmatter is loaded at launch).

Failure: add `.claude/rules/deployment/prod.md` with no `paths:` frontmatter and 10 KB of prose. It enters context on every session, and `check:rule-ids` still reports the total as if it weren't there — the gate silently stops measuring the thing it exists to measure. `CITING_SOURCES` already carries both `*.md` and `**/*.md`; this list should too.

```suggestion
    ".claude/rules/*.md",
    ".claude/rules/**/*.md",
```

—Claude

scripts/check_rule_ids.py
```diff
@@ -142,6 +153,77 @@
     {"scripts/check_rule_ids.py", "scripts/tests/test_check_rule_ids.py"}
```
timothyfroehlich, line 167: Scope gap: `LEGACY_CITATION_SOURCES` (just above, lines 131–146) does not include `.claude/rules/**`, so the directory this PR makes the primary home for rule prose is exempt from the fragile-citation gate.

Failure: a future `.claude/rules/database.md` edit that says "see AGENTS.md §2.1" or "per rule 12" passes `pnpm run check` clean, which is exactly the drift PP-22e4 built this check to stop. `CITING_SOURCES` covers the tier for `CORE-*` IDs; the legacy scan should cover it too.

Adding the globs surfaces a second, pre-existing bug — see the comment on `NUMBERED_RULE_CITATION`.

—Claude

scripts/check_rule_ids.py
```diff
@@ -72,16 +81,18 @@
     r"\b(?:AGENTS\.md`?\s+)?(?:[Rr]ule|[Cc]ommandment)\s*#?\d+\b"
```
timothyfroehlich: `NUMBERED_RULE_CITATION` uses `\s*` / `\s+`, which match newlines, so the pattern spans lines.

Failure: `.claude/rules/README.md` contains `## Adding or changing a rule` followed by a blank line and `1. **State it in ...**`. That matches as the citation `"rule\n\n1"`. Any markdown heading ending in the word "rule" above an ordered list is a false positive. It is latent today only because `.claude/rules/**` is not in `LEGACY_CITATION_SOURCES`; add the tier there (see the other comment) and `pnpm run check` fails on this PR's own README.

A citation lives on one line — use horizontal whitespace only:

```suggestion
NUMBERED_RULE_CITATION = re.compile(
    r"\b(?:AGENTS\.md`?[ \t]+)?(?:[Rr]ule|[Cc]ommandment)[ \t]*#?\d+\b"
)
```

(`SECTION_21_CITATION` on line 88 has the same `\s*` and wants the same treatment.)

—Claude

.claude/rules/README.md
```diff
@@ -0,0 +1,52 @@
...
+## Globs
+
+Patterns are repo-relative and match the same way everywhere else
+(`**` spans directories, `*` does not). Route-group parentheses are literal —
+`src/app/**` covers `src/app/(app)/…` without needing to escape anything, so
+prefer it over spelling the group out.
+
+One glob set is load-bearing beyond this directory. `server-actions.md` matches
+action modules by **filename**, because a Server Action is marked by a
+`"use server"` directive and a glob cannot read one. `pinpoint/server-action-file-naming`
```
timothyfroehlich, line 51: This states in the present tense that `pinpoint/server-action-file-naming` exists in `eslint.config.mjs`. It does not — `rg -l server-action-file-naming` on this branch hits only these two rule files, and #1833 (which adds it) is still OPEN.

Failure: if #1834 merges first, both `README.md` and `server-actions.md` tell the reader that a lint rule guarantees filenames and globs agree, while nothing does. Someone adds `src/app/(app)/foo/bar.ts` with `"use server"`, `pnpm run check` stays green, and the three CORE-ARCH rules silently stop loading for that file — the precise failure the paragraph claims is prevented.

(All 24 current `"use server"` modules do match the globs, so this is about the future guarantee, not today's state.) Fix is merge order: land #1833 first, or state the dependency rather than the fact.

—Claude

.claude/rules/server-actions.md
```diff
@@ -0,0 +1,49 @@
...
+## Why these globs are filenames
+
+A Server Action module is marked by a `"use server"` directive, and a glob
+cannot read one. Actions are also deliberately scattered — a route-local
+action colocates with its route under `src/app/**`, and only genuinely
+cross-cutting ones live in `src/server/actions/` — so there is no single
+directory to point at either.
+
+The filename globs above are therefore the seam, and
+`pinpoint/server-action-file-naming` in `eslint.config.mjs` is what keeps them
```
timothyfroehlich, line 36: Same as the `README.md` note: `pinpoint/server-action-file-naming` is not in `eslint.config.mjs` on this branch — #1833 is still open. "is what keeps them honest" is a claim about a rule that does not exist yet, in the one paragraph whose whole job is to say why the filename globs cannot rot silently.

—Claude

timothyfroehlich:
Two findings that have no line in this diff to anchor to (`docs/NON_NEGOTIABLES.md` is untouched by the PR; the other is outside a hunk):

**1. `docs/NON_NEGOTIABLES.md:6` — the sync contract now points at the deleted index.**

> **Sync contract**: `AGENTS.md`'s rule index is a one-line summary of these rules. Every index entry cites the canonical `CORE-*` ID(s) here. When a rule changes, update both.

This PR deletes that index. Concrete failure: someone adds `CORE-XXX-001` to the catalog, follows the contract to AGENTS.md §2.1, finds a pointer paragraph with nothing to sync, and stops. The rule never reaches `always.md` or any path-scoped file, and it shows up only in `check:rule-ids --orphans`, which never fails the build. The PR body lists four claims it refreshed because this PR falsifies them — this is a fifth, and it is the one that is load-bearing for keeping the new tier in sync.

**2. `scripts/check_rule_ids.py:390` — stale inline comment.** Still reads "§2.1 is being replaced by a grouped index, so both forms break the moment that happens." The module docstring and the user-facing message were both updated to past tense; this one was missed.

—Claude

scripts/check_rule_ids.py
timothyfroehlich: Fixed in 471b7dda: `ALWAYS_LOADED_SOURCES` now carries `.claude/rules/**/*.md` alongside the flat glob, with a test that a nested no-`paths:` file is budgeted. Both collectors dedupe by relative path, so the overlap costs nothing. —Claude

scripts/check_rule_ids.py
timothyfroehlich: Fixed in 471b7dda: added `.claude/rules/*.md` and `.claude/rules/**/*.md` to `LEGACY_CITATION_SOURCES`. This is also what surfaced your next finding — with the surface covered, the README's own heading tripped the newline-spanning regex. —Claude

scripts/check_rule_ids.py
timothyfroehlich: Fixed in 471b7dda: both patterns now use `[ \t]` instead of `\s`, since a citation lives on one line. Added a test for the exact shape — a heading ending in "rule" above an ordered list. Good catch that it was latent only until the scope fix landed. —Claude

.claude/rules/README.md
timothyfroehlich: Declining a wording change — the resolution is merge order, and I am enforcing it rather than leaving it to memory: I will not hand this PR over until #1833 is merged, so `pinpoint/server-action-file-naming` is in `eslint.config.mjs` on `main` before this sentence reaches it. A PR-number reference here would be noise the moment #1833 lands. —Claude

.claude/rules/server-actions.md
timothyfroehlich: Declining a wording change — the resolution is merge order, and I am enforcing it rather than leaving it to memory: I will not hand this PR over until #1833 is merged, so `pinpoint/server-action-file-naming` is in `eslint.config.mjs` on `main` before this sentence reaches it. A PR-number reference here would be noise the moment #1833 lands. —Claude

timothyfroehlich:
Claude review of head 4dc80a5 — `/code-review high` — 5 findings (2 medium, 3 low): CORE-ARCH-008/011 globs missed src/services and src/lib/mcp entirely — both rules moved to always.md; stale 'AGENTS.md contains the non-negotiables' claims in antigravity.md and design-bible; 'Four checks' docstring listing three; ui.md's wrong claim that the globs exclude tests; wrapped rule-number citations no longer caught. All fixed.

PR #1837, opened 2026-08-09
chore(ci): bump the pinned Supabase CLI to 2.113.0 (PP-nlv6)

timothyfroehlich:
Weekly-chores version-drift item (PP-nlv6, chores bead PP-qehv).

## What changed

All **9** `supabase/setup-cli` pins move from `2.109.1` → `2.113.0` (the current latest stable), together:

| File | Sites |
| --- | --- |
| `.github/workflows/ci.yml` | 6 |
| `.github/workflows/preview-control.yaml` | 1 |
| `.github/workflows/preview-sync.yaml` | 1 |
| `.github/workflows/preview-reaper.yaml` | 1 |

All nine or none — a partial bump leaves jobs on mismatched CLI versions and shows up as a job-specific failure that reads like a flake.

## Version range crossed

`2.110.0` → `2.111.0` → `2.112.0` → `2.113.0`. Release notes reviewed for anything that changes `supabase start` / `db reset` / `db push` / `migration` / `branches` behavior our workflows depend on.

**The one breaking change does not apply to us.** 2.112.0 stopped forwarding `NPM_AUTH_TOKEN` into Docker bundling for `functions deploy` (restoring Go CLI behavior). PinPoint ships no edge functions and has no `.npmrc`, so nothing here touches it.

**The 2.112.0 deprecations also do not apply**: `db diff --use-pg-schema` and `db pull --experimental` are unused in this repo.

### Notable, in our favor

- **2.112.0** — `supabase start` no longer hangs when analytics migrations fail; `start` reuses existing volumes instead of failing when they already exist; Kong reloads after `db reset` so auth requests stop 502-ing when container IPs rotate. All three are exactly the class of infra flake the `setup-supabase` composite action retries around today.
- **2.112.0** — migrations now run in true version order (`db push` / `migration up` no longer misreport an already-applied migration as missing when one version string is a prefix of another).
- **2.112.0** — `-o toml/yaml/json` machine-readable output is back at Go-CLI parity. Worth naming because `.github/actions/setup-supabase/action.yml` parses `supabase status --output json` with `jq` (`API_URL`, `ANON_KEY`, `SERVICE_ROLE_KEY`, `DB_URL`, `INBUCKET_URL`) — this moves toward the shape we expect, not away from it, and CI exercises it on every job.
- **2.110.0 / 2.113.0** — TypeScript-port progress: `start`, `stop`, `status`, `db push`, `db reset` are now served by the TypeScript shell. Upstream states behavior matches the Go CLI; CI is the check.

## Verification

- `pnpm run check` green (includes actionlint, yamllint, zizmor — the real gate for workflow edits).
- `rg '2\.109\.1'` returns one remaining hit, deliberately left alone: `scripts/backup-production.sh:60` is a prose comment recording that `supabase db reset` was *verified* local-only against CLI 2.109.1. That is a historical verification record, not a pin — rewriting it to say 2.113.0 would assert a re-verification that did not happen.
- Full E2E is CI's job here, not local.

https://claude.ai/code/session_01JRJXTCULMF915M6LN5zJgq

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -501,7 +501,7 @@ jobs:
         uses: supabase/setup-cli@46f7f98c7f948ad727d22c1e67fab04c223a0520  # ratchet:supabase/setup-cli@v3.0.0
         with:
           # Pinned Supabase CLI — bump deliberately (was: latest). PP-d5zn.
-          version: 2.109.1
+          version: 2.113.0
 
       - name: Setup Supabase
         uses: ./.github/actions/setup-supabase
@@ -563,7 +563,7 @@ jobs:
         uses: supabase/setup-cli@46f7f98c7f948ad727d22c1e67fab04c223a0520  # ratchet:supabase/setup-cli@v3.0.0
         with:
           # Pinned Supabase CLI — bump deliberately (was: latest). PP-d5zn.
-          version: 2.109.1
+          version: 2.113.0
 
       - name: Setup Supabase
         uses: ./.github/actions/setup-supabase
@@ -619,7 +619,7 @@ jobs:
         uses: supabase/setup-cli@46f7f98c7f948ad727d22c1e67fab04c223a0520  # ratchet:supabase/setup-cli@v3.0.0
         with:
           # Pinned Supabase CLI — bump deliberately (was: latest). PP-d5zn.
-          version: 2.109.1
+          version: 2.113.0
 
       - name: Cache Playwright browsers
         uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # ratchet:actions/cache@v6.1.0
@@ -701,7 +701,7 @@ jobs:
         uses: supabase/setup-cli@46f7f98c7f948ad727d22c1e67fab04c223a0520  # ratchet:supabase/setup-cli@v3.0.0
         with:
           # Pinned Supabase CLI — bump deliberately (was: latest). PP-d5zn.
-          version: 2.109.1
+          version: 2.113.0
 
       - name: Cache Playwright browsers
         uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # ratchet:actions/cache@v6.1.0
@@ -783,7 +783,7 @@ jobs:
         uses: supabase/setup-cli@46f7f98c7f948ad727d22c1e67fab04c223a0520  # ratchet:supabase/setup-cli@v3.0.0
         with:
           # Pinned Supabase CLI — bump deliberately (was: latest). PP-d5zn.
-          version: 2.109.1
+          version: 2.113.0
 
       - name: Cache Playwright browsers
         uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # ratchet:actions/cache@v6.1.0
@@ -866,7 +866,7 @@ jobs:
         uses: supabase/setup-cli@46f7f98c7f948ad727d22c1e67fab04c223a0520  # ratchet:supabase/setup-cli@v3.0.0
         with:
           # Pinned Supabase CLI — bump deliberately (was: latest). PP-d5zn.
-          version: 2.109.1
+          version: 2.113.0
 
       - name: Cache Playwright browsers
         uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # ratchet:actions/cache@v6.1.0
--- a/.github/workflows/preview-reaper.yaml
+++ b/.github/workflows/preview-reaper.yaml
@@ -53,7 +53,7 @@ jobs:
       - uses: supabase/setup-cli@46f7f98c7f948ad727d22c1e67fab04c223a0520  # ratchet:supabase/setup-cli@v3.0.0
         with:
           # Pinned Supabase CLI — bump deliberately (was: latest). PP-d5zn.
-          version: 2.109.1
+          version: 2.113.0
 
       - name: Reap preview branches
         run: bash scripts/workflow/preview/preview-reap.sh
--- a/.github/workflows/preview-control.yaml
+++ b/.github/workflows/preview-control.yaml
@@ -115,7 +115,7 @@ jobs:
         if: steps.parse.outputs.action == 'start' || steps.parse.outputs.action == 'stop'
         with:
           # Pinned Supabase CLI — bump deliberately (was: latest). PP-d5zn.
-          version: 2.109.1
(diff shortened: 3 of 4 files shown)


.github/workflows/preview-reaper.yaml
```diff
@@ -53,7 +53,7 @@ jobs:
       - uses: supabase/setup-cli@46f7f98c7f948ad727d22c1e67fab04c223a0520  # ratchet:supabase/setup-cli@v3.0.0
         with:
           # Pinned Supabase CLI — bump deliberately (was: latest). PP-d5zn.
-          version: 2.109.1
+          version: 2.113.0
```
timothyfroehlich, line 56: **Low — the green CI on this PR does not cover the three preview workflows, which are the ones most exposed to this bump.**

Every job that ran here uses the CLI through `.github/actions/setup-supabase` (`supabase start`, `supabase status --output json`). The preview trio instead calls `supabase branches list --output json`, `branches get -o env`, and `branches delete` — and `-o toml/yaml/json` machine-readable output is exactly what 2.112.0 changed (#6002, "matches the Go CLI again"). `Supabase Preview` shows `skipping` on this PR, so none of that ran at 2.113.0.

Concrete failure mode if the shape did move: `scripts/workflow/preview/preview-destroy.sh:33` and `preview-resync.sh:32` both swallow a CLI error into `|| echo '[]'`, so a failed `branches list` reads as "No Supabase branch named 'pr-N' (already gone)" and the branch is never deleted — a billed Supabase branch leaks silently with a green workflow. This reaper degrades safely (`jq 'length'` on a non-array yields 0 or the name regex skips), so the risk is leaked branches rather than wrongful deletion, but leaked branches are the thing this workflow exists to prevent.

Cheap fix: run `/preview` on this PR once (start then stop) before merging. That exercises create → get → list → delete against 2.113.0 and turns the PR body's parity argument into an observation. Worth it given 2.113.0 landed 2026-08-08, ~24h before this PR, and the reaper runs hourly against the production project ref.

—Claude

.github/workflows/ci.yml
```diff
@@ -501,7 +501,7 @@ jobs:
         uses: supabase/setup-cli@46f7f98c7f948ad727d22c1e67fab04c223a0520  # ratchet:supabase/setup-cli@v3.0.0
         with:
           # Pinned Supabase CLI — bump deliberately (was: latest). PP-d5zn.
-          version: 2.109.1
+          version: 2.113.0
```
timothyfroehlich, line 504: **Low — `supabase db reset` is reimplemented in this range and CI's green run almost certainly never executed it.**

2.113.0 moved `db reset` onto the TypeScript shell (#6026). The only `supabase db reset` in the repo's CI path is `e2e/global-setup.ts:413`, and it is the *fallback* arm: step 4 runs `pnpm run db:fast-reset` and `return`s on success; the full `supabase db reset --yes` + migrate + seed chain only runs after fast reset throws. On a CI runner the composite action has already migrated and seeded, so fast reset succeeds and the fallback is skipped. So the green E2E jobs are evidence for `start` / `status` at 2.113.0, not for `db reset`.

Scenario: the next time fast reset legitimately fails (fresh checkout, drifted schema, a seed script erroring), CI takes the recovery arm against a freshly-ported `db reset` for the first time, and a behavior difference there fails the exact path that exists to recover from failure — surfacing as an E2E setup error on an unrelated PR.

One local `supabase db reset --yes` at 2.113.0 (or a CI run with `db:fast-reset` forced to fail) closes this. Same rationale covers the `scripts/backup-production.sh:60` note you deliberately left at 2.109.1: that comment's prod-safety claim — a bare `db reset` is local-only — was verified against the Go implementation, and `db reset` is now different code. Leaving the prose alone is right; re-verifying the claim once is worth more than the version string in it.

—Claude

.github/workflows/preview-reaper.yaml
timothyfroehlich: Verified live against the prod project rather than via `/preview`, because **`/preview` cannot test this pre-merge**: `preview-control.yaml` is `issue_comment` and `preview-reaper.yaml` is `schedule`/`workflow_dispatch`, and GitHub runs both from the **default branch** — so a `/preview` on this PR would have exercised main's 2.109.1, not 2.113.0. (`preview-sync.yaml` does run from the PR branch, but its paths filter is `drizzle/**` + seeds, which this PR does not touch.)

What I ran instead, against `udhesuizjsgxfeotqybn`, read-only, at both versions:

- `branches list --output json` — output is **byte-identical** between 2.109.1 and 2.113.0. Keys: `created_at, git_branch, id, is_default, name, parent_project_ref, persistent, preview_project_status, project_ref, status, updated_at, with_data`. `.name` — the only key `preview-destroy.sh`, `preview-resync.sh`, `preview-create.sh` and `preview-reap.sh` read — is present and unchanged.
- `branches get <name> -o env` — **identical key set** at both versions, including the `POSTGRES_URL` that `preview-create.sh:84` polls for.

So the `|| echo '[]'` leak path you identified is real as a design weakness, but 2.113.0 does not trigger it. Not filing it as a blocker for this bump.

**Separately, and more useful than the finding:** CLI **2.112.0 is genuinely broken** for this command — it fails with `SchemaError(Expected a string matching the RegExp ...)` on `created_at`, because its client-side validator demands a `Z` suffix while the API returns `+00:00`. 2.113.0 fixes it. That is an argument for taking this bump, not against it. Anyone on a local 2.112.0 should upgrade. —Claude

.github/workflows/ci.yml
timothyfroehlich: Re-verified the underlying safety claim rather than the version string, as suggested.

`supabase db reset --help` diffed between 2.109.1 and 2.113.0: the only changes are sentence-casing of the global flag descriptions and one new `--wizard` flag. **The remote-capable flags are unchanged** — `--linked` ("Resets the linked project with local migrations") and `--db-url` are both still explicit opt-ins, and a bare `supabase db reset` still targets the local stack only.

So `scripts/backup-production.sh:60`'s prod-safety claim still holds at 2.113.0, and leaving its prose at 2.109.1 stays correct — it records the version a verification was performed against, not a pin. The `e2e/global-setup.ts:413` fallback arm remains untested by CI for the reason you gave, but the property that matters for prod safety is verified. —Claude

timothyfroehlich:
Claude review of head 00bfd8d — `/code-review high` — 2 low findings, both verification gaps, both closed with live read-only checks: branches list JSON byte-identical 2.109.1 vs 2.113.0, branches get -o env keys identical, db reset remote flags unchanged

PR #1838, opened 2026-08-09
fix(huddle): detect subagents by transcript path, not env markers (PP-uxnn)

timothyfroehlich:
## Problem

`huddle-whoami.sh register` refused **every top-level Claude Code session**:

```
$ bash scripts/hooks/huddle-whoami.sh register Claude-HerdrRecon <sid>
huddle-whoami.sh: refusing to register — this is a dispatched subagent.
```

No session could claim a name, so the self-filter map stayed empty and huddle attribution had nothing to key off.

## Root cause

Not herdr. **Claude Code sets both markers on every shell the Bash tool spawns**, top-level session and dispatched subagent alike. From the bundled CLI (2.1.226, identical in 2.1.224/225):

```js
function T2t(e){
  let t={CLAUDECODE:"1", CLAUDE_CODE_SESSION_ID:e.sessionId,
         CLAUDE_CODE_CHILD_SESSION:"1", CLAUDE_PID:String(process.pid)};
  if(e.source==="agent") t.AI_AGENT=w6i("agent");   // -> claude-code_2-1-226_agent
  ...
}
```

and the Bash tool's env builder calls it unconditionally as `T2t({sessionId: Ft(), effortLevel: s, source: "agent"})`. `source: "agent"` means *"the model is spawning this process"* — any tool call — not *"dispatched subagent"*. So the PP-788v guard's predicate (`CLAUDE_CODE_CHILD_SESSION` set **and** `AI_AGENT` ending in `_agent`) is true for everyone.

**Measured env delta between the top-level session and a dispatched subagent: none.** `AI_AGENT`, `CLAUDE_CODE_CHILD_SESSION` and `CLAUDE_CODE_SESSION_ID` are byte-identical (the subagent inherits the parent's session id). There is no env-level discriminator.

herdr's contribution is zero: `ps -Ewww` on the `claude` process shows it was exec'd with only `HERDR_ENV` / `HERDR_PANE_ID` / `HERDR_SOCKET_PATH` / `HERDR_TAB_ID` / `HERDR_WORKSPACE_ID`, and `~/.claude/hooks/herdr-agent-state.sh` exports nothing — it reads hook stdin and reports over a unix socket.

## Fix

Detect by **transcript path** — the same `*/subagents/*` signal `huddle-poll.sh` and `huddle-session-start.sh` already use. They get the path from their hook payload; a script invoked from the Bash tool has to find its own record, so the guard scans the trailing records of `<project>/<session>.jsonl` and every `<project>/<session>/subagents/*.jsonl` for the newest one naming `huddle-whoami.sh`. The harness flushes the `tool_use` record for the running command to the **calling** agent's transcript before the command executes, so that record identifies the caller.

Details that fell out of measuring the real thing:

- The scan covers a short window of trailing records, not just the last line: a `PreToolUse` `hook_success` attachment can land between the `tool_use` and execution (observed once in 34 live Bash calls). That record carries the command string too.
- Ties across candidate transcripts are broken by the newest matching record's timestamp, so a stale match elsewhere cannot outrank the live one.
- An indeterminate read — no matching record, another harness, an invocation wrapped in a different script — is treated as "not a subagent", the same fail-open posture the old guard had. The PP-788v rebind guard in `register` remains the backstop.

`--force` semantics and the PP-788v no-silent-rebind rule are untouched.

## Verification

Both directions, live, replaying **real** transcript records under a fake `HOME`:

- the top-level session's own recorded `register Claude-HerdrRecon …` call → **allowed**
- a real subagent record, newer, added alongside it → **refused**
- this PR's authoring subagent running `discover` against the real tree → **refused**

Plus pytest coverage for both directions (`scripts/tests/test_huddle_whoami.py`), including the herdr-shaped env that triggered the bug, the trailing-attachment case, and the no-transcripts fail-open path.

`pnpm run check` and `pnpm run check:python` (501 tests) both clean.

## Docs

`.agents/skills/pinpoint-huddle/SKILL.md` §"Subagent sessions" rewritten in place to state the transcript-path contract and why the env marker is not a discriminator.

Closes PP-uxnn.

https://claude.ai/code/session_01HiinjHi8c1VC1c98BZwoa5


scripts/hooks/huddle-whoami.sh
```diff
@@ -121,20 +123,209 @@ discover_session_id() {
...
+# but reading the file is not invoking it — and an agent researching this guard
+# does exactly that. Both misreadings are live: a subagent that greps the file
+# makes its parent look like a subagent, and a parent that cats it makes a real
+# subagent look top-level.
+#
+# So the signature must sit in COMMAND POSITION: at the start of the command or
+# after a `;`/`&`/`|` separator, optionally behind a `bash`/`sh` prefix, and
+# optionally behind a directory path. As an argument to some other program it
+# does not count.
+WHOAMI_INVOCATION_RE="(^|[;&|][[:space:]]*)[[:space:]]*(([^[:space:]]*/)?(ba)?sh[[:space:]]+)?([^[:space:]]*/)?$(
```
timothyfroehlich: **Multi-line Bash commands never match, so the guard silently fails open.**

`WHOAMI_INVOCATION_RE` only accepts the signature at string start or after a `;`/`&`/`|`. jq's regex engine (Oniguruma, Perl syntax) runs with `ONIG_OPTION_SINGLELINE` by default, so `^` is `\A` — *string* start, not line start — and a newline is not in `[;&|]`.

Verified:
```
$ jq -n '"cd /x\nbash scripts/hooks/huddle-whoami.sh discover" | test("(^|[;&|][[:space:]]*)[[:space:]]*(([^[:space:]]*/)?(ba)?sh[[:space:]]+)?([^[:space:]]*/)?huddle-whoami\\.sh([[:space:]]|$)")'
false
```

This is not a corner case: over the 8 most recent transcripts in `~/.claude/projects/-Users-froeht-Code-PinPoint`, **117 of 669 Bash `tool_use` commands (17.5%) contain a newline**. Concrete scenario: a dispatched subagent runs
```
cd "$REPO"
bash scripts/hooks/huddle-whoami.sh register Claude-Sub $SID
```
No candidate transcript yields a match, `caller_transcript` returns 1, `caller_is_subagent` returns false, and the register proceeds — the PP-788v clobber path the guard exists to close. Every `INVOCATIONS` case in the test suite is single-line, so the suite stays green.

Adding a newline to the separator class fixes it without loosening the command-position requirement (re-verified: all six `MENTIONS` still return `false`).

```suggestion
WHOAMI_INVOCATION_RE="(^|[;&|\n][[:space:]]*)[[:space:]]*(([^[:space:]]*/)?(ba)?sh[[:space:]]+)?([^[:space:]]*/)?$(
```

Worth adding a multi-line case to `INVOCATIONS` too. (Same class, lower frequency: a prefix other than `bash`/`sh` — `timeout 5 bash …`, `env bash …` — also misses.)

scripts/hooks/huddle-whoami.sh
```diff
@@ -121,20 +123,209 @@ discover_session_id() {
...
+WHOAMI_JQ_NEWEST_EPOCH_MS='
+  def epoch_ms:
+    capture("^(?<base>[^.]+?)(?:\\.(?<frac>[0-9]+))?Z$")
+    | (((.base + "Z") | fromdateiso8601) * 1000)
+      + (((.frac // "0") + "000")[0:3] | tonumber);
+  [ inputs
+    | fromjson? // empty
+    | select(.timestamp != null)
+    | select(
+        (.message.content // [])
```
timothyfroehlich: **One badly-shaped record aborts the whole jq run, silently disabling detection.**

`.message.content` is an unguarded index. Every other risky step here is defensive (`fromjson? // empty`, `epoch_ms? // empty`, `.input.command // ""`, the `if type == "array"` guard) — this one is not, and jq aborts the entire program on the first offending record rather than skipping it:

```
$ printf '%s\n' '{"timestamp":"2026-08-09T00:00:00.000Z","message":"hi"}' | jq -n -R '[inputs|fromjson? // empty|select(.timestamp!=null)|select((.message.content // [])|type=="array")]'
jq: error (at <stdin>:1): Cannot index string with string "content"
exit=5
```

The caller swallows it (`2>/dev/null || true`), so `ms` comes back empty, the candidate is skipped, and — if it happens on the transcript that holds the caller's record — the read goes indeterminate and fails open. A subagent registers.

I checked 2946 records across the 8 newest live transcripts and `.message` was an object in all of them, so this is latent rather than currently firing. But it is a whole-guard disable triggered by a single unexpected record shape from a format nobody controls, and the fix is one character:

```suggestion
        (.message?.content? // [])
```

.agents/skills/pinpoint-huddle/SKILL.md
```diff
@@ -36,7 +36,11 @@ don't register, just sign your posts.
...
+
+**Detection is by transcript path — the same `*/subagents/*` signal the two hooks use.** A hook gets its transcript path in the payload; a script invoked from the Bash tool has to find its own record, so `huddle-whoami.sh` scans the trailing records of `<project>/<session>.jsonl` and every `<project>/<session>/subagents/*.jsonl` for the newest one naming the script. The harness flushes a tool_use record to the **calling** agent's transcript before the command runs, so that record identifies the caller. An indeterminate read — no matching record, another harness, an invocation wrapped in a different script — is treated as "not a subagent"; the rebind guard in `register` is the backstop.
```
timothyfroehlich: **This restates the assumption the PR itself found to be false.** "The harness flushes a tool_use record to the calling agent's transcript before the command runs" is exactly the certainty the script's own comment walks back:

> "Usually" is load-bearing and was originally written as a certainty. The flush is not guaranteed — a top-level `discover` was observed running with its own record still unwritten.

The skill is the durable reference an agent reads when this misbehaves, and as written it would send someone back down the same path. It also omits the `WHOAMI_FRESH_SECONDS` bound, which is now half the detection: a stale record is ignored entirely, not merely outranked, and "no *fresh* matching record" is a distinct indeterminate case from the three listed.

Suggest: "The harness *usually* flushes a tool_use record … before the command runs; the flush is not guaranteed, so a matching record only counts when it is also recent (see `WHOAMI_FRESH_SECONDS`)."

scripts/hooks/huddle-whoami.sh
timothyfroehlich: Fixed in 15c8780b. Confirmed against jq directly before changing anything — `^` is start-of-string under ONIG_OPTION_SINGLELINE, and the multi-line case returned false while the single-line one returned true. `\n` added to the separator class; all six MENTIONS still return false, and a multi-line mention (`cd "$REPO"\nrg -n discover …`) was added to that list so the fix cannot become a blanket pass. Two multi-line INVOCATIONS cases added, both verified failing against the pre-fix regex.

The related `timeout 5 bash …` gap is left open deliberately and now documented in the source: it fails toward "not a subagent", which the rebind guard catches, whereas widening the pattern risks false positives, which have no backstop. —Claude

scripts/hooks/huddle-whoami.sh
timothyfroehlich: Fixed in 15c8780b — `.message?.content?`. Reproduced the abort first (`{message:"a string"} | (.message.content // [])` raises, and `//` does not catch it), which confirms one odd record would take out the whole run rather than just that record. New test puts two non-object-`.message` records ahead of a genuine invocation in the same transcript, so it only passes if the run survives them; verified failing against the unguarded version. —Claude

.agents/skills/pinpoint-huddle/SKILL.md
timothyfroehlich: Fixed in 15c8780b. Rewrote the paragraph rather than patching the one sentence: it now says the flush happens "usually" before the command runs, and documents both bounds that exist because of it — the 120s freshness window (with the 18-minute-stale-record incident that motivated it) and command-position-not-mention matching. Also states why the bias runs toward fail-open, since that is the part a future reader most needs. —Claude

timothyfroehlich:
Claude review of head 388a7c1 — `/code-review medium` — 3 findings, all fixed: multi-line commands never matched the invocation regex (jq ^ is string-start; 17.5% of real commands are multi-line, fail-open to PP-788v clobber); unguarded .message.content aborted the whole jq run on one odd record; skill doc restated the disproven flush-ordering assumption and omitted the freshness bound.

PR #1839, opened 2026-08-09
chore(pinballmap): refresh vendored llms.txt for the new Attribution section (PP-2h1b)

timothyfroehlich:
PinballMap changed its published `llms.txt` in a way that touches our CORE-PBM-001 attribution obligation. This refreshes the vendored copy and propagates the new rule into the docs that state it. **No behavior changes** — we already comply with the hard requirement.

## What actually drifted

`Last updated:` moved 2026-07-22 → 2026-08-06, and a new `## Attribution` section was added. That section is the *only* substantive change; the rest of the file is byte-identical.

Its substance:

- PBM data is CC BY-SA 4.0 and requires attribution.
- A generic `https://pinballmap.com` link is fine for general credit (an about page), **but a surface displaying data for a specific location must link to that location's listing** — `https://pinballmap.com/map?by_location_id=<location_id>`.
- They additionally **encourage** (explicitly not require) explanatory language such as *"update this location listing at Pinball Map \<link\>"*, because PBM is a source of truth many people rely on and they want readers routed back to correct it.

`robots.txt` was re-diffed against live and is unchanged — left alone.

## The hard requirement: already satisfied

The investigation was prompted by a suspicion that `public-url.ts` built the link as `?n=`. It does not, and it never did — `git log -S` across all history finds no such form. `pinballmapLocationUrl()` has emitted `by_location_id` since PP-o355.3 (#1659).

The one open question was the trailing slash: we emit `/map/?by_location_id=`, PBM documents `/map?by_location_id=`. Checked empirically with two `HEAD` requests (a descriptive User-Agent, no crawling):

| URL | Result |
| :-- | :-- |
| `https://pinballmap.com/map/?by_location_id=26454` (ours) | `HTTP/2 200`, no redirect |
| `https://pinballmap.com/map?by_location_id=26454` (documented) | `HTTP/2 200`, no redirect |

Rails treats them identically, so the slash is incidental and the URL is left as-is rather than churning four test fixtures for no functional gain. That finding is recorded in the `public-url.ts` docstring so nobody has to re-derive it.

## What changed

A verbatim refresh nobody reads is inert, so the new obligation was propagated into the places that assert it:

- **`docs/external/pinballmap-llms.txt`** — refreshed verbatim. SHA-256 of the committed file matches the `curl` output byte-for-byte, re-verified after `format:fix` and after the pre-commit hooks (prettier's glob is `{js,jsx,ts,tsx,json,css,md}`, so `.txt` is never a candidate — safe by extension rather than by an explicit `.prettierignore` entry).
- **`docs/external/README.md`** — dated refresh note in the established blockquote format; provenance date bumped.
- **`docs/external/pinballmap-api-terms.md`** — the Attribution section now quotes the location-specific requirement instead of only the vaguer FAQ line, and points at `pinballmapLocationUrl()` as the single place a pinballmap.com URL may be built.
- **`docs/NON_NEGOTIABLES.md` CORE-PBM-001 + the AGENTS.md §2.1 summary** — "a link back to pinballmap.com" tightened to a location-specific link built by `pinballmapLocationUrl()`, never hand-written. Also corrected a factual error in the same paragraph: it claimed the vendored files are "kept current by the drift GHA", but there is no drift GHA — `docs/external/README.md` says so, and it is the weekly chores routine that does it.
- **`src/lib/pinballmap/public-url.ts`** — docstring only. Records that `by_location_id` is a documented requirement rather than a stylistic choice, and that hand-writing a PBM URL elsewhere is how the requirement gets silently dropped.

## What was deferred, and why

**PP-2h1b (P1, `pinballmap`)** — the encouraged explanatory copy. Neither of our two rendered link-backs carries it:

- `machine-pinballmap-card.tsx` — link text "View on Pinball Map" (the public Info tab, visible to anonymous visitors; the primary attribution surface)
- `edit/page.tsx` — the Manage tab's "Pinball Map" section heading is itself the link

(A third consumer, `PinballmapListingControl.tsx`, is not currently rendered — PP-o355.19 replaced its only call site with a placeholder.)

Deferred rather than done here because it is user-facing copy, which is Tim's call and not an agent's (AGENTS.md §6); because the Manage-tab surface is slated for a wholesale rewrite in PP-o355.21, so copy written now would churn; and because PBM frames it as "if possible, we encourage", so nothing is out of compliance meanwhile. The bead carries the exact quote, both file locations, and a suggested shape for the wording.

## Verification

- `pnpm run check` — green
- `pnpm run test` — 223 files, 2052 tests, all passing
- `pnpm exec vitest run src/lib/pinballmap/public-url.test.ts` — 2/2 passing (unchanged; the code change is a comment)

Docs + one docstring, so no E2E run.

https://claude.ai/code/session_01JRJXTCULMF915M6LN5zJgq


timothyfroehlich:
Claude review of head c3345d6 — `/code-review low` — No findings — docs-only plus one JSDoc rewrite, no executable code changed

PR #1843, opened 2026-08-09
docs(skills): fold three Vercel deploy diagnostics into pinpoint-deployment (PP-p4ek)

timothyfroehlich:
## What

Adds a **Production Deploys (Vercel)** section to `.agents/skills/pinpoint-deployment/SKILL.md` — the skill that already owns the deploy surface — carrying three facts that were about to be lost.

The 2026-08-09 memory review demoted two Vercel beads memories (`vercel-env-registry-ordering`, `vercel-cli-setup`) out of the always-loaded tier. Three operational facts in them had no home in the repo:

1. **Build order.** `vercel-build` is `migrate:production && next build`, so a build that dies later — on the env assertion, on a compile error — has **already migrated prod**. Vercel doesn't promote a failed build, so prod keeps serving the *old* code against the *new* schema. Verify forward-compatibility rather than assuming nothing happened.
2. **Diagnosing without build logs.** The `claude_ai_Vercel` MCP's `get_deployment_build_logs` 401s. Infer from a very short `buildingAt`→`ready` delta (~19s means it died at module evaluation, not on a compile error) plus whether the previous production deploy succeeded.
3. **The SENSITIVE footgun.** `vercel env add <name> production` defaults to Vercel's write-only SENSITIVE type — `vercel env pull` returns the literal `[SENSITIVE]` (length 11) and the dashboard can't show it either. The app works while the secret is unretrievable by humans. Pass `--no-sensitive` for anything Tim also needs a copy of.

Also updates the frontmatter description and the intro paragraph so the skill actually gets loaded on "production deploy failed" / "set a Vercel env var" triggers.

## Absence check (`rg --hidden`)

- Fact 1 — absent everywhere. ✅
- Fact 2 — the *ordering* is stated in a comment in `drizzle/0059_overrated_inhumans.sql` and defined in `package.json`, but the **consequence** (a failed build already migrated prod) is in no doc or skill.
- Fact 3 — **partially present already**: `docs/ENV_VARS.md` states it as an aside on `PINBALLMAP_API_TOKEN`'s entry. The skill now carries the general rule plus the verify-by-length-then-shred procedure, and points at ENV_VARS.md rather than duplicating it.

## Scope

Docs-only, one file, 25 lines added. Deliberately does **not** restate the CORE-SEC-009 registry rule — duplicating that is what got these memories demoted in the first place.

`pnpm run check` green.

Bead: PP-p4ek


timothyfroehlich:
Claude review of head 44622d6 — `/code-review low` — No findings — docs-only, one skill file.

PR #1844, opened 2026-08-09
docs(agents): move this Mac's hardware facts into user context (PP-ljx7)

timothyfroehlich:
## What

`AGENTS.md` encoded one machine's specs as project facts — "a 16 GB laptop" in four places, plus per-OS `brew`/`apt` install commands. None of that is true of PinPoint; it is true of the machine Tim runs it on.

## The line

- **Repo** states what a command _costs and needs_: three parallel Chromium workers, a Supabase stack, a Next server, peak RSS. True on any host, useful to any tool.
- **User context** states what _this hardware_ can absorb and what to run instead.

The offload answer (run it on the Bazzite desktop) lives in a stowed Claude skill, not here — so agents that cannot reach that runner correctly fall through to CI without being told about a machine they cannot use.

## Changes

| File                                                  | Change                                                                                                                                        |
| ----------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
| `AGENTS.md` §4                                        | Host prerequisites state the requirement; the macOS/Linux command split is gone                                                                |
| `AGENTS.md` §5                                        | `e2e:full` / `e2e:all` state their footprint instead of judging the reader's laptop                                                            |
| `AGENTS.md` §5                                        | Lint-engine note ends at the RSS numbers                                                                                                      |
| `AGENTS.md` §5                                        | "Which tests to run" keeps the real rules (never bare `playwright test`, always `--project=chromium`, report flaky) and drops the hardware clause |
| `scripts/guard/mem-precheck.sh`                       | Hard-block message names `crabbox` as the preferred option, guarded on the CLI existing                                                        |
| `scripts/workflow/heavy-run.sh`, `preflight-locked.sh` | Header rationale says "memory-constrained host", not "16 GB Mac" — the guards themselves are unchanged                                         |

No behavior change beyond the added suggestion line in `mem-precheck.sh`.

## Verification

- `rg -i '16 ?GB|resource-constrained' AGENTS.md` → nothing
- `pnpm run check` clean

Docs and one shell message; nothing to screenshot.


timothyfroehlich:
Claude review of head 1c066ab — `/code-review medium` — 4 findings, all fixed: two checked-in skills still asserted the deleted 16 GB rule (missed because rg skips dot-dirs); pytest lost its only install mention with no runtime hint to replace it; the crabbox suggestion line used bare pipes.

PR #1845, opened 2026-08-10
docs(deployment): record the SELinux/pgsodium stack failure and its CLI cause (PP-9mg0)

timothyfroehlich:
Replaces #1824, which is closed. Same bug, no code.

## What happened

`supabase start` on Bazzite (Fedora Atomic, SELinux enforcing, `docker` shimming rootless podman) never completed — the `supabase_db_*` container restart-looped on:

```
pgsodium_getkey.sh: /etc/postgresql-custom/pgsodium_root.key: Read-only file system
FATAL:  invalid secret key
```

Two messages, one cause, and neither of them says SELinux or names a CLI version. It blocked `db:fast-reset`, `test:integration:supabase`, and smoke E2E on that host, and diagnosing it took a full session.

## Why #1824 is closed instead of merged

#1824 shipped a script to label the CLI's secret-staging directory `container_file_t:s0`. A `/code-review medium` pass found that the bind mount it works around exists **only in CLI 2.111.0** — 2.112.0 replaced it with `docker cp` from a temp file, and 2.113.0's `secretFiles` doc says it is never a host bind mount ([supabase/cli#6022](https://github.com/supabase/cli/issues/6022)).

**#1837 bumped the pin to 2.113.0 earlier today**, so the bug was already gone before any of that code could land. Verified on Bazzite at 2.113.0:

| Check | Result |
| --- | --- |
| CLI version | 2.113.0 |
| Supabase stack | healthy |
| `supabase/.temp/start-secrets` | never created |
| Host bind mounts into the repo, any container | none — db has only its data volume |
| The labeling script in that worktree | absent, so nothing labeled anything |

The script would have labeled a directory nothing creates or reads. The review also found two real defects in it (unreachable in the main worktree; the AGENTS.md bullet stated two things that were no longer true), both moot once the premise went.

## What this PR keeps

The part that has ongoing value: the cause was a **CLI version**, so an older CLI brings the bug straight back, and nothing about the error message would lead anyone to this explanation a second time. A new section in `pinpoint-deployment` records the symptom, the mechanism (staged file inherits `user_home_t`, container runs as `container_t`, SELinux denies the `stat`, so `test -f` is false on `EACCES` and the keygen tries to *create* the key through a read-only mount), why 2.112.0+ ended it, and the one-line `chcon` if it ever returns.

Routed to the skill rather than AGENTS.md per §8 — actionable "what/how" in AGENTS.md, deep dives in skills — and the skill's `description` frontmatter now names the symptom so the lookup works from the error message.

Closes PP-9mg0.

https://claude.ai/code/session_01Dt1h4mHucYX7oQQyjUL4LT


timothyfroehlich:
Claude review of head 0f76b8a — `/code-review low` — No findings at review; re-attested after a merge-from-main conflict resolution (#1843 rewrote the same skill description) — reviewed content byte-identical, only the frontmatter description and intro sentence changed.

PR #1848, opened 2026-08-10
docs(agents): point subagent quality gate at AGENTS.md §5, not `check` alone (PP-lql4)

timothyfroehlich:
## Problem

PP-4zcj (#1800) made `pnpm run check` a **pure static gate** on 2026-08-02: types, lint, format, and the shell/YAML/Python linters. **It no longer runs unit tests or pytest.** AGENTS.md was updated thoroughly; the `pinpoint-orchestrator` skill and its dispatch prompt template were not — and those are what actually reach subagents.

Hooks don't fire for subagents, so the dispatch prompt IS the enforcement. A lead following the template verbatim dispatched agents whose only instructed gate cannot fail on a broken unit test: the agent returns "check green", opens a PR, and CI is the first thing that notices. Observed 2026-08-02 on PRs #1807/#1809, where both agents ran `pnpm run test` on their own initiative — diligence covering for the instruction, not the instruction working.

## Approach

Every site now **points at AGENTS.md §5 "Which tests to run"** as the single source of truth rather than restating a command list. A restated list is a second place to drift, which is the exact bug being fixed. The one fact each site does state inline is the correction itself: `check` is static and cannot fail on a broken test.

## Files changed

| File | What |
| :--- | :--- |
| `.agents/skills/pinpoint-orchestrator/SKILL.md` | Phase 3 dispatch checklist item 3 |
| `.agents/skills/pinpoint-orchestrator/references/agent-prompt-template.md` | the prompt body (Quality Gates) and the Key Points rationale |
| `CLAUDE.md` | "Quality Enforcement" under Parallel Subagent Workflow — the surviving anti-pattern site the bead placed at `SKILL.md:304` |
| `.agents/skills/pinpoint-superpowers-bridge/SKILL.md` | tiered test commands + the overrides table (found by the sweep, beyond the bead's five sites) |
| `.claude/agents/investigator.md` | command block claimed `check` runs `test`, and that `preflight` runs smoke — both wrong (sweep) |

## Notes

- **`.claude/skills` is a symlink to `.agents/skills`** (mode `120000` in git), so "the two trees agree" holds by construction — no duplicate edit needed.
- The bead named a fifth site at `SKILL.md:304` (anti-patterns). That section no longer exists in the current `SKILL.md`; the surviving equivalent is CLAUDE.md's "Quality Enforcement" line, which is fixed here. The bead's other line numbers are also stale relative to the rewritten file.
- **`AGENTS.md` and `docs/NON_NEGOTIABLES.md` deliberately untouched** — open PR #1834 rewrites both. No fix here needed them; AGENTS.md §5 is already correct and is what everything now points to.
- Dated artifacts under `docs/plans/`, `docs/superpowers/`, and `docs/testing/unit-test-audit-2026-06.md` still mention the old `check` semantics; per AGENTS.md §8 those are records, left alone.

Docs-only. `pnpm run check` green.

Closes PP-lql4.

### DIFF
--- a/.agents/skills/pinpoint-orchestrator/SKILL.md
+++ b/.agents/skills/pinpoint-orchestrator/SKILL.md
@@ -75,7 +75,7 @@ Present options to the user. Before proceeding, verify tasks are independent:
 
 1. The bead ID + "First run `bd show <id>` && `bd update <id> --claim`, then work from the bead." The bead carries scope, files, line numbers, and acceptance criteria — do NOT restate them in the prompt (two places to drift).
 2. Only context that ISN'T already in the bead — cross-bead conflicts, a reference PR, sequencing constraints. Omit when there's nothing to add.
-3. Quality gate: "Run `pnpm run check` before returning."
+3. Quality gate: point the agent at **AGENTS.md §5 "Which tests to run"** and tell it to run the tiers matching what it touched. Do **not** write "run `pnpm run check`" and stop — `check` is a static gate that runs no unit tests and no pytest (PP-4zcj), so that instruction cannot fail on a broken test and CI becomes the first thing that notices (PP-lql4).
 4. Full PR lifecycle: "Create PR, verify CI green."
 5. Structured return format: branch, PR#, CI status, blockers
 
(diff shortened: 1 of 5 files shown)


timothyfroehlich:
Claude review of head 8f0108a — `/code-review low` — No findings — docs-only PR (5 markdown/skill files) redirecting the subagent quality gate from 'pnpm run check' alone to AGENTS.md §5's tiered list. No runtime code changed.

PR #1849, opened 2026-08-10
fix(e2e): retire the dead root suite, move the orphaned profile-edit spec into full (PP-8oeq, PP-stut)

timothyfroehlich:
Fixes PP-8oeq and PP-stut. They are two halves of the same rot — a suite pointing at a spec that no longer exists, and a spec that no suite points at — so they land together.

## What was broken

**PP-8oeq — `pnpm run e2e:all` could not pass.** `scripts/workflow/e2e-all-isolated.sh` ran a third "root" suite:

```
pnpm exec playwright test --config=playwright.config.ts e2e/machines-filtering.spec.ts
```

That spec was deleted on 2026-05-16 in #1343. Playwright exits non-zero when a path filter matches no tests, and the script is `set -euo pipefail` with `|| exit $?`, so the script failed at the end of every run — after already spending ~10 minutes on full + smoke. Nobody noticed because the script has no caller: CI invokes the configs directly, and AGENTS.md said not to run `e2e:all` locally. The prohibition is exactly what hid the rot.

**PP-stut — `e2e/profiles/profile-edit.spec.ts` had never run.** Added 2026-06-20 in #1572, matched by nothing: not `playwright.config.full.ts` (`testMatch: **/full/**/*.spec.ts`), not `playwright.config.smoke.ts` (`testDir: e2e/smoke`), not any CI job. The only invocation that would have picked it up is `pnpm run e2e:mobile` — `playwright test --project='Mobile Chrome'` with no spec path and no `--config`, i.e. the whole-tree form AGENTS.md forbids for cross-contaminating seed state.

## The decision PP-8oeq asked for: does a "root" suite still have a job?

No. Retired, not repointed.

Both beads' notes suggested repointing the root invocation at `e2e/profiles/`. That fixes the exit code but not the coverage: **`e2e:all` has no CI caller**, so profile-edit would still never run in CI — it would just be dead coverage in a new place, and PP-stut's acceptance explicitly requires a CI job to exercise it.

So instead:

- `e2e/profiles/profile-edit.spec.ts` → `e2e/full/profile-edit.spec.ts`, where `playwright.config.full.ts` matches it and the required `test-e2e-full-chromium` job runs it. The relative import (`../support/auth-state.js`) is unchanged by the move.
- The third invocation is deleted. With every spec now under `full/` or `smoke/`, those two suites are the exact partition of the spec corpus — which is *why* a third suite has nothing left to do. A comment in the script records that, so the next spec doesn't land at `e2e/` root or in a sibling directory and quietly repeat this.

`e2e:all` keeps its actual job (running full and smoke as separate Playwright processes so global-setup re-resets the DB between them); it just no longer claims a third suite.

## The spec was dormant, not broken — but it was not suite-safe (commit 2)

PP-stut warned it might already be broken. Run alone it passes in 11.7s. Run **in the suite** it came back **flaky on both chromium and firefox**, which is not done. Two distinct races, both fixed:

1. **It raced its own save.** The spec clicked Save and then issued its own `page.goto` back to the read view. `updateProfileAction` already calls `revalidatePath` and `redirect()` on success, so that manual navigation was racing the in-flight Server Action and could abort it. It now waits for the redirect the action performs — which drops `?edit`, so the URL assertion is itself the proof that the write landed before the read.

2. **It raced other copies of itself.** The full config declares four browser projects and Playwright runs them concurrently, and this journey writes the seeded member's **own profile row — a singleton**. Two projects interleave and each asserts the other's value. The runner's failure snapshot is unambiguous: the page showed `they/them 1786329226292` while that copy expected `…226307`. It is now pinned to `chromium`, the project the required `test-e2e-full-chromium` job runs. A text form has no cross-browser risk worth triple-writing a shared row for.

## `e2e:all` pinned to chromium (commit 3, from review)

Review caught that removing the dead suite fixed the symptom but not the script: neither invocation passed `--project`, so both configs still ran all four browser projects concurrently against one database — the same contention as race 2 above. Both lines now pass `--project=chromium`, mirroring the required CI jobs.

Chose that over documenting the limitation because the multi-project run isn't merely expensive, it's uninformative: several specs write singleton seeded rows, so a red run says nothing about the code. Cross-browser coverage belongs to CI, where each browser gets its own job and its own database. The reasoning lives in the script header.

## Verification

Measured:

- `pnpm run check`, `pnpm run check:python` (530 passed), `pnpm run test` (2099 passed)
- profile-edit alone on the Bazzite runner: passes
- profile-edit across chromium + firefox + Mobile Chrome **simultaneously**, after the fix: 4 passed, 2 skipped, no flake
- the whole `full` suite on the runner with the spec in place (all projects, 8.4m) — which is how both races were found

- **`pnpm run e2e:all`, complete, on the Bazzite runner: green.** Exit 0, both legs, ending on the script's own `✅ All E2E suites passed (full + smoke, chromium)` banner — 2m9s wall for the pair. (The shared runner lease was held by another session for the first four attempts; this landed on the fifth.)

So PP-8oeq's acceptance is met on **both** of its branches, not just the removal one: the dead invocation is gone *and* the script runs green end to end.

**Pre-existing condition worth knowing, not introduced here:** the whole-suite runner run also failed 9 unrelated tests (`machine-info`, `machine-owner-picker`, `machine-timeline`, `machine-settings`) on firefox and Mobile Chrome only — the same shared-row contention. The required CI job is chromium-only so it never sees this; the post-merge `E2E Comprehensive` job does run three projects and is `continue-on-error`. Out of scope here, but it is a real reason that job's signal is noisy.

## AGENTS.md corrections

Unblocked now that #1834 has merged:

- The `e2e:all` row no longer claims a "roots" suite — the last place saying one existed — and states the chromium pinning.
- The `e2e:full` row was also wrong in the other direction: it claimed "Three parallel Chromium workers", but `e2e:full` runs **every** browser project. Corrected, with a pointer to `--project=chromium`.
- The "never invoke playwright with no spec path" paragraph now explains why `--project=chromium` matters for whole-suite runs, not just targeted ones.

## Follow-up, deliberately not in this PR

**PP-2g7m** — a static gate that fails when any `e2e/**/*.spec.ts` is matched by zero suite config, so this class can't recur. Filed from PP-stut's notes; kept out of here to keep this PR to the two bug fixes. That bead also carries the second-order defect found on the way: `pnpm run e2e:mobile` is itself the forbidden no-spec-path invocation.


scripts/workflow/e2e-all-isolated.sh
```diff
@@ -38,9 +46,8 @@ run_suite() {
 
 run_suite "full" pnpm exec playwright test --config=playwright.config.full.ts || exit $?
```
timothyfroehlich: `e2e:all` still cannot pass, for a different reason. Neither `run_suite` passes `--project`, so both invocations run every project in `playwright.config.ts` (chromium, firefox, Mobile Chrome, Mobile Safari) concurrently — exactly the multi-project shared-row contention this PR documents (the 9 `machine-*` failures on firefox/Mobile Chrome, and the reason `profile-edit` itself had to be pinned to chromium). So after this change a local `pnpm run e2e:all` will still exit non-zero on most runs; PP-8oeq's symptom is fixed at the removed-suite level but not end to end.

If the intent is that `e2e:all` mirrors the required CI job, both lines want `--project=chromium` (CI line 810 uses exactly that). If the intent is genuinely multi-project coverage, this is worth saying in the header comment so the next person doesn't read a red run as a regression they caused.

—Claude

timothyfroehlich: Correct, and fixed in c9d3c1f7 — both lines now pass `--project=chromium`.

Took the mirror-CI option rather than documenting the limitation, because the multi-project run isn't merely expensive, it's uninformative. The four projects run concurrently against one database and several specs write singleton seeded rows (the member's own profile, machine owners, machine settings), so two projects interleave on the same row and each asserts the other's value. A red run therefore carries no signal about the code — which is exactly the "next person reads it as a regression they caused" problem, except documenting it wouldn't stop the run from being wrong. Cross-browser coverage is CI's, where each browser gets its own job and its own DB. Reasoning is in the script header so the `--project` flags don't look like an arbitrary narrowing.

**Scoping the claim, since you're right that PP-8oeq's acceptance is about end-to-end behavior:** I did *not* run `pnpm run e2e:all` to completion. The Bazzite runner's lease was held by another session throughout (retried on a backoff, never freed), and it's far too heavy for this Mac. So what's verified is the invocation itself — `--project=chromium` on both lines, matching ci.yml — plus the individual pieces measured earlier on the runner: the full config at `--project=chromium` and the moved spec across three projects simultaneously. I'm not claiming the script now runs green end to end; that's inferred, not measured, and it isn't stated anywhere in the PR as fact. PP-8oeq's acceptance is satisfied via its other branch (dead invocation removed) with this as an additional improvement.

—Claude

timothyfroehlich: Upgrading the previous reply: the runner lease freed and I got the measurement, so the caveat above no longer applies.

`pnpm run e2e:all` on Bazzite, complete, **exit 0** — both legs, ending on the script's own banner:

```
✅ All E2E suites passed (full + smoke, chromium)
command complete in 2m8.777s ... exit=0
```

Smoke leg: 68 passed, 4 skipped, 56.8s. So PP-8oeq's acceptance is now satisfied on both of its branches — the dead invocation is removed *and* the script runs green end to end — rather than only the first. PR body updated to state the measurement rather than the gap.

—Claude

timothyfroehlich:
Claude review of head c9d3c1f — `/code-review medium` — One finding: e2e-all-isolated.sh ran all four Playwright projects concurrently against one DB, so e2e:all still exited non-zero. Fixed by pinning both suites to --project=chromium, mirroring CI's required job. Thread resolved. Second finding (AGENTS.md 'roots' wording) also fixed, unblocked by #1834 merging.

PR #1850, opened 2026-08-10
fix(gates): lint and format .mjs before push, not only in CI (PP-ru2p)

timothyfroehlich:
Closes PP-ru2p.

## The bug

PR #1796 (PP-ojv5) extended `pnpm run lint` from `src/` to `src/ e2e/ scripts/`, which is 11 `.mjs` files. CI Gate started linting them; nothing before push did.

- `lint-staged` globs were `*.{ts,tsx}` and `*.{js,jsx,json,css,md}` — neither matches `.mjs`
- `.husky/pre-commit` runs lint-staged + typecheck + format; none of those reach `.mjs`
- `.husky/pre-push` runs no lint at all
- Prettier's `format` glob (`**/*.{js,jsx,ts,tsx,json,css,md}`) excludes `.mjs`, so `.mjs` was formatted by nothing, anywhere

Net effect: edit `scripts/db-fast-reset.mjs`, introduce a `no-undef`, commit and push — every local hook passes, then CI Gate goes red on a file no local gate ever touched.

## The fix

**`package.json` — lint-staged.** New group `{src,e2e,scripts}/**/*.{mjs,cjs,js,jsx}` running `eslint --fix` + `prettier --write`, and the old `*.{js,jsx,json,css,md}` group narrowed to `*.{json,css,md}`.

Two deliberate constraints on that glob:

- **Scoped to the three directories CI lints**, not a bare `*.{mjs,cjs}`. A bare glob would pull in `.claude/hooks/*.cjs` and `.agents/hooks/*.cjs`, which produce 109 `no-undef`/`no-unused-vars` errors under the current config and which CI does not lint. That would make the local hook *stricter* than CI — the same class of local/CI disagreement this bead is about, pointed the other way.
- **Disjoint from the other two groups.** lint-staged sets `matchBase: !pattern.includes('/')`, so a bare pattern matches its extensions in every directory. Overlapping groups run concurrently, which would race `eslint --fix` against `prettier --write` on the same file. `{json,css,md}` and `{mjs,cjs,js,jsx}` share nothing, so the three groups stay non-overlapping.

`.mjs`/`.cjs` outside those three directories (`supabase/`, `eslint-rules/`, `.agents/`, root `*.config.mjs`) is not auto-fixed by lint-staged. It is gated by `pnpm run format`, which pre-commit runs — see the `--log-level` change below for why that gate is only now actionable.

**`package.json` — prettier globs.** `format` / `format:fix` gain `mjs,cjs`. This reformats 12 pre-existing files (139 lines total); that noise is the one-time cost of the extension never having been covered.

**`package.json` — `format` log level** (review finding 1). `format` drops `--log-level error` for `--log-level warn`; `format:fix` keeps `error`. At `error`, `prettier --check` prints **nothing at all** on failure — the pre-commit abort and CI's `Prettier Format Check` log were both blank, for every file type, not just `.mjs`. At `warn` the offending filenames and the remedy line are printed, and success is still silent. The flag stays on `format:fix` because `prettier --write` at any lower level prints **1130 lines** (measured), which would drown the nine other tasks `pnpm run check` runs in parallel. The flag was added to both scripts at once in #735, where only the `format:fix` half had a reason.

**`eslint.config.mjs`** (NOTES sibling 1). The Node-globals block keyed on `scripts/**/*.{mjs,ts}` while the relaxed-rules block above it uses `scripts/**/*`, so a future `scripts/foo.cjs` or `.js` would fall outside the globals block and re-trigger the 133-error `no-undef` wall. Widened to `scripts/**/*.{mjs,cjs,js,jsx,ts,tsx}` — the JS-family extensions only, deliberately not bare `scripts/**/*`, which would make ESLint treat `scripts/*.py` and `scripts/*.sh` as lintable source.

**`scripts/check-prod-backups.mjs`** (NOTES sibling 2). Fifth cause-dropping `throw` — the bare `catch {}` form meant `preserve-caught-error` never flagged it (`requireCatchParameter` is off). Now `catch (error)` + `{ cause: error }`.

Both NOTES siblings were genuinely small, so both are folded in rather than deferred to a follow-up bead.

**`.js`/`.jsx` under the same three directories** (review finding 2). CI's `eslint src/ e2e/ scripts/` lints them via ESLint 9 directory expansion; nothing local did — PP-ru2p one extension over, and inconsistent with the `eslint.config.mjs` widening above. Closed rather than deferred: `git ls-files | rg '\.(js|jsx)$'` returns **zero** files repo-wide, so narrowing `*.{js,jsx,json,css,md}` to `*.{json,css,md}` removes auto-fix from an empty set and needs no third "outside those dirs" group to avoid the concurrency race. Should a `.js` ever land outside the three directories, it falls to `pnpm run format`, which now names it.

## Evidence the gate now fires

Not "the config looks right" — the bead is about a gate that does not fire, so this was proven by injecting a real error.

Appended to `scripts/db-fast-reset.mjs`:

```js
const  x=deliberatelyUndefinedGlobal( 1,2 );
console.log(x);
```

…which is both a `no-undef` and badly formatted. Staged it, then ran lint-staged twice — once against the **old** glob set (reconstructed in a temp config file), once against the new one in `package.json`:

| Gate | Result |
| :-- | :-- |
| lint-staged, **old** globs | `→ lint-staged could not find any staged files matching configured tasks.` — **exit 0**, bug reproduced |
| lint-staged, **new** globs | `✖ eslint --fix` → `75:10 error 'deliberatelyUndefinedGlobal' is not defined no-undef` — **exit 1** |
| `pnpm run lint:local` (what `pnpm run check` runs) | same `no-undef`, **exit 1** |
| `pnpm run format` (new glob), clean tree | **exit 0**, silent |
| lint-staged, staged `scripts/pp-ru2p-probe.js` (finding 2) | `✖ eslint --fix` → `1:11 error 'notDefinedAnywhere' is not defined no-undef` — **exit 1** |

The deliberate error was then reverted; the tree is clean and the commit contains none of it.

### The diagnostic is actionable, not just non-zero (finding 1)

Appended an unformatted line to `supabase/seed-users.mjs` — a `.mjs` *outside* the three directories, the exact case the reviewer raised — staged it, and ran the real `git commit`:

```
→ lint-staged could not find any staged files matching configured tasks.
$ tsc --noEmit -p tsconfig.app.json
$ prettier --check "**/*.{js,jsx,mjs,cjs,ts,tsx,json,css,md}" --log-level warn
[warn] supabase/seed-users.mjs
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
husky - pre-commit script failed (code 1)
```

Under `--log-level error` that same block was the `husky` line alone — abort, no filename, no remedy. CI's `Prettier Format Check` runs the same script, so its log gains the filenames too.

On the lint-mirror question: the lint-staged hook invokes **authoritative** `eslint` (no `PINPOINT_LINT_SLIM`), and `lint:local`'s two engines were already pointed at `scripts/`, so this PR adds no new mirror surface. `no-undef` comes from `js.configs.recommended`, which slim mode does not touch (it only strips type-aware rules), and oxlint does not configure `no-undef` at all — so the `.mjs` coverage lands entirely on the ESLint side of the mirror, identical to CI. No `.oxlintrc.json` change is needed.

## Verification

- `pnpm run check` — pass
- `pnpm run lint` (authoritative full ESLint, what CI runs) — pass
- `pnpm run format` — pass
- `pnpm run test` — 2099 passed / 224 files
- `pnpm run check:python` — 525 passed (touched `scripts/`)


package.json
```diff
@@ -22,8 +22,8 @@
     "lint:_oxlint": "oxlint src/ e2e/ scripts/",
     "lint:_slim": "PINPOINT_LINT_SLIM=1 eslint src/ e2e/ scripts/ --quiet",
     "fix:lint-format": "npm-run-all --sequential lint:fix format:fix",
-    "format": "prettier --check \"**/*.{js,jsx,ts,tsx,json,css,md}\" --log-level error",
-    "format:fix": "prettier --write \"**/*.{js,jsx,ts,tsx,json,css,md}\" --log-level error",
+    "format": "prettier --check \"**/*.{js,jsx,mjs,cjs,ts,tsx,json,css,md}\" --log-level error",
```
timothyfroehlich: **`.mjs`/`.cjs` become the only extensions in the format glob with no repo-wide lint-staged auto-fix — and the resulting failure prints nothing.**

Every other extension here is auto-fixed by lint-staged *anywhere in the repo*, because `*.{ts,tsx}` and `*.{js,jsx,json,css,md}` are bare patterns and lint-staged sets `matchBase: !pattern.includes('/')` (verified in `node_modules/lint-staged/lib/matchFiles.js`). The new group is path-anchored to `{src,e2e,scripts}`, so `.mjs`/`.cjs` under `supabase/`, `eslint-rules/`, `.agents/hooks/` and root configs are format-*gated* but never format-*fixed* by the hook.

Combined with `--log-level error`, that failure is silent. Verified locally:

```
$ pnpm exec prettier --check "**/*.{js,jsx,mjs,cjs,ts,tsx,json,css,md}" --log-level error
exit=1        # zero output
$ ... --log-level warn
[warn] supabase/seed-timeline-demo.mjs
[warn] Code style issues found in 12 files.
```

Scenario: edit `supabase/seed-users.mjs`, add a line over 80 chars, `git commit`. lint-staged matches nothing, `pnpm run typecheck` passes, `pnpm run format` exits 1 having printed **no file name and no remedy** — the commit aborts with a blank diagnostic. CI's `Prettier Format Check` job runs the same script, so the CI log is blank too. The PR body's claim that these files "fail the commit with an actionable message" does not hold as written.

Two ways out: drop `--log-level error` from `format` (keep it on `format:fix`, which has nothing useful to say), or add a prettier-only lint-staged group for `.mjs`/`.cjs` outside the three directories so the class is auto-fixed like every other extension in this glob.

package.json
```diff
@@ -212,6 +212,10 @@
       "eslint --fix",
       "prettier --write"
     ],
+    "{src,e2e,scripts}/**/*.{mjs,cjs}": [
```
timothyfroehlich: **The same hole this PR closes for `.mjs`/`.cjs` is left open for `.js`/`.jsx` in the same three directories.**

`pnpm run lint` is `eslint src/ e2e/ scripts/`, and ESLint 9 directory expansion lints `.js`/`.cjs`/`.mjs` by default — so CI's ESLint job would lint a future `scripts/foo.js`. Locally, `.js` only matches lint-staged's third group, which runs `prettier --write` and no `eslint --fix`. That is exactly PP-ru2p's failure mode, one extension over.

The PR is internally inconsistent about this: `eslint.config.mjs` was widened to `scripts/**/*.{mjs,cjs,js,jsx,ts,tsx}` with a comment naming "a future `scripts/foo.cjs` or `.js`" as the reason, but this glob stops at `{mjs,cjs}` — so the config is hardened for a `.js` file the hook still cannot lint.

No `.js`/`.jsx` exists under `src/`, `e2e/`, or `scripts/` today (`git ls-files src e2e scripts | rg '\.(js|jsx)$'` is empty), so this is latent rather than live. Closing it needs both halves, or you reintroduce the concurrency race the PR body warns about: widen this glob to `{mjs,cjs,js,jsx}` **and** narrow `*.{js,jsx,json,css,md}` to `{json,css,md}` plus a `.js`/`.jsx`-outside-those-dirs group, otherwise `eslint --fix` and `prettier --write` run concurrently on the same file.

package.json
timothyfroehlich: Fixed in 24b17f21, taking the first option: `format` now uses `--log-level warn`, `format:fix` keeps `error`.

You were right that the PR body's claim did not hold — I've corrected that sentence. Checking the flag's history before changing it: it was added to **both** scripts in one line of #735 (a large unrelated feature PR), and only the `format:fix` half has a reason. `prettier --write` over this glob prints **1130 lines** at any level below `error`, which would drown the nine other tasks `pnpm run check` runs in parallel. `prettier --check` has the opposite property — the `[warn]` list *is* the diagnostic. So the flag was deliberate for one script and collateral damage on the other.

Picked `warn` over dropping the flag entirely because it keeps success silent (default level adds `Checking formatting...` + `All matched files use Prettier code style!` to every `check` run) while printing everything actionable on failure.

Your reasoning for preferring this over the second option holds up — I confirmed the blank log is not `.mjs`-specific. Verified end-to-end on your exact scenario, a real `git commit` with an unformatted `supabase/seed-users.mjs` staged:

```
→ lint-staged could not find any staged files matching configured tasks.
$ tsc --noEmit -p tsconfig.app.json
$ prettier --check "**/*.{js,jsx,mjs,cjs,ts,tsx,json,css,md}" --log-level warn
[warn] supabase/seed-users.mjs
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
husky - pre-commit script failed (code 1)
```

Under `--log-level error` that block was the `husky` line alone. CI's `Prettier Format Check` runs the same script, so its log gains the filenames too.

—Claude

package.json
timothyfroehlich: Closed rather than deferred, in 24b17f21 — the race you flagged does not materialize, and the reason is your own observation taken one step further.

You noted no `.js`/`.jsx` exists under the three directories. It's stronger than that: `git ls-files | rg '\.(js|jsx)$'` is empty **repo-wide**, so the `js,jsx` half of `*.{js,jsx,json,css,md}` matches nothing anywhere today. That means the two halves of your prescription collapse into one edit with no third group:

- `{src,e2e,scripts}/**/*.{mjs,cjs}` → `{src,e2e,scripts}/**/*.{mjs,cjs,js,jsx}`
- `*.{js,jsx,json,css,md}` → `*.{json,css,md}`

`{json,css,md}` and `{mjs,cjs,js,jsx}` are disjoint, so no file can match both groups and `eslint --fix` never runs concurrently with `prettier --write`. The narrowing removes lint-staged auto-fix from an empty set; `pnpm run format`'s glob still covers `js,jsx` repo-wide, so the *gate* is unchanged, and after the finding-1 fix it now names the file. A `.js` landing outside the three directories later gets the same treatment as `.mjs` there.

Confirmed both directions. ESLint 9 does expand `.js` under `scripts/` — a probe `scripts/pp-ru2p-probe.js` calling an undefined global is reported by `pnpm exec eslint scripts/`, which is what made this a real hole rather than a theoretical one. And staged, lint-staged now catches it locally:

```
✖ eslint --fix:
scripts/pp-ru2p-probe.js
  1:11  error  'notDefinedAnywhere' is not defined  no-undef
```

Probe deleted; the internal inconsistency with `eslint.config.mjs` is gone.

—Claude

timothyfroehlich:
Claude review of head 24b17f2 — `/code-review high` — Two findings, both fixed in 24b17f21. (1) format used --log-level error, so a .mjs outside {src,e2e,scripts} aborted the commit with a blank diagnostic; format now uses warn, format:fix keeps error. PR body claim corrected. (2) .js/.jsx left uncovered in the same dirs; closed by widening the lint-staged group and narrowing the other, which stay disjoint so no eslint-fix/prettier-write race. Both threads resolved.

PR #1851, opened 2026-08-10
feat(mcp): issue read, find, comment, and update tools (PP-u4ab.14)

timothyfroehlich:
Four MCP tools so Claude can work an issue end to end: read one in full, find issues across the collection, comment, and edit its fields. Closes the gap where the MCP surface could file an issue but never read, find, or change one afterwards.

**Bead:** PP-u4ab.14 · **Spec:** `docs/superpowers/specs/2026-08-09-mcp-issue-tools-design.md` · **Plan:** `docs/superpowers/plans/2026-08-09-mcp-issue-tools.md`

## The tools

| Tool | What |
| :--- | :--- |
| `get_issue` | One issue in full — description, status/severity/priority/frequency, reporter + assignee names, comment thread |
| `list_issues` | Cross-machine or scoped; filters on status (a **set**), severity, assignee; paged with `total`/`hasMore` |
| `add_issue_comment` | Post a comment, deduped on retry within a 10-minute window |
| `update_issue` | Change any of title, status, severity, frequency, priority, assignee |

All four take `machine` (initials or UUID) + `number`. Issue UUIDs are deliberately not accepted — no tool returns one, so it would be an argument the model could never populate.

## Three decisions worth reviewing

**1. `addIssueComment` gained an explicit `deduped` flag.** The plan proposed deriving "did this write anything" from an empty delivery plan. That proxy is unsound: a genuinely new comment also produces zero deliveries when the commenter is the only watcher, when no notification channels are configured, or when the best-effort notify block swallows an error — so it would report "already posted" for a real write. The service now returns the flag the way `createIssue` already does, and an integration test pins the notifies-nobody case specifically.

**2. `update_issue` is not atomic, and the response says so.** Each field is its own service function with its own transaction. Wrapping them would mean rewriting six service functions and touching the web edit path — the exact cost this bead was scoped to avoid. Instead a mid-loop failure returns a *success* payload with `partial: true` and the field that stopped it, plus per-field `{from, to, changed}`. Telling the caller nothing landed when half of it did is the CORE-ARCH-012 shape to avoid. Assignee resolution runs ahead of the loop so an unknown name fails the whole call cleanly.

**3. `list_issues` takes a set of statuses, not one.** `list_machines` shipped `presence` as a single value and needed widening immediately (PP-u4ab.13). Same failure applies here, so the set form ships first. An unresolvable assignee filter throws rather than falling through to "no filter" — a filter name that narrows nothing returns the whole collection under an authoritative-looking `total`.

## Permissions

Per-field, because the matrix splits issue edits two ways: `issues.update.reporting` for title/status/severity/frequency, `issues.update.triage` for priority/assignee. Only supplied fields are gated, so a status-only call is not denied for lacking triage rights it never uses. A unit test asserts the map, including that each permission id actually exists in the matrix — a typo would fail closed and silently.

`get_issue` never selects `reporterEmail`; a test asserts no `@` survives into the payload (CORE-SEC-007).

## Tests

- **Unit** — comment idempotency key (6 cases), `update_issue` permission map (4)
- **Integration** — 40 new cases in `src/test/integration/mcp-tools.test.ts`, 78 total in the file, all green
- **New** — a catalog test asserting all 12 registered tools by name. Every other test calls `run*` directly, so registration was the one place a finished tool could be silently unreachable.

## Verification

`pnpm run check` clean · full unit suite 2109 passed · integration 609 passed and `next build` clean (both on Bazzite via crabbox).

## Not in scope

`issues.reassign` (changes the issue's URL, and its permission carries an `owner` clause needing `OwnershipContext`), comment edit/delete, watchers, image attachments.

https://claude.ai/code/session_01Nr3bZsDP4mAG4iChVdSmLU


src/lib/mcp/tools/add-issue-comment.ts
timothyfroehlich, line 1: **A literal NUL byte in the source makes this file binary to git and ripgrep.**

`parts.join(...)` on this line contains a raw `0x00` byte (offset 1723) rather than the `"\u0000"` escape that `create-issue.ts` uses for the identical scheme. The runtime behaviour is correct, but the file is now byte-for-byte binary as far as the toolchain is concerned:

- `git diff origin/main...HEAD` reports `Binary files /dev/null and b/src/lib/mcp/tools/add-issue-comment.ts differ` — the whole file is unreviewable in this PR, and `git blame` / future diffs on it will be too.
- `rg -n "checkPermission" src/lib/mcp/tools/` silently skips it; asking for it by name prints `binary file matches (found "\0" byte around offset 1723)` instead of the match. AGENTS.md makes `rg` the default search tool, so every codebase-wide sweep from here on quietly misses this file.

Fix is a one-character change to match `create-issue.ts` line 107:

```suggestion
  const digest = createHash("sha256").update(parts.join("\u0000")).digest();
```

src/lib/mcp/tools/get-issue.ts
```diff
@@ -0,0 +1,134 @@
...
+          eq(issueComments.issueId, issue.id),
+          // System rows carry `eventData`, not prose. Rendering them into
+          // readable history is real work for marginal value: the transitions
+          // they describe are already visible as the issue's current state.
+          eq(issueComments.isSystem, false)
+        ),
+        columns: { content: true, createdAt: true },
+        with: { author: { columns: { name: true } } },
+        orderBy: (c, { asc }) => [asc(c.createdAt)],
+        limit: args.commentLimit ?? DEFAULT_COMMENT_LIMIT,
```
timothyfroehlich: **`asc(createdAt)` + `limit` returns the OLDEST N comments and drops the newest ones with no signal.**

On a thread with 30 comments and the default `commentLimit` of 20, the caller gets comments 1–20 and never sees 21–30 — and the payload carries no `commentCount` / `hasMore`, so nothing distinguishes a complete thread from a truncated one. The tool description tells the model "use this before commenting or updating, so you are acting on the issue you think you are", which is exactly the case this breaks: the most recent comment ("parts ordered, don't touch it") is the one that gets dropped, and the agent then comments or closes on stale context.

The plan doc (`docs/superpowers/plans/2026-08-09-mcp-issue-tools.md:451`) says "newest last", which reads as intent to return the *tail* of the thread. Either take the last N (order `desc`, then reverse before returning) or add `commentCount` + `hasMoreComments` so truncation is visible.

src/lib/mcp/tools/update-issue.ts
```diff
@@ -0,0 +1,292 @@
...
+            to: next,
+            changed: issue.assignedTo !== next,
+          });
+          break;
+        }
+      }
+    } catch (error) {
+      failure = {
+        field,
+        reason: error instanceof Error ? error.message : "unknown error",
```
timothyfroehlich: **This catch relays raw internal error text to the MCP client and skips Sentry entirely.**

`failed.reason` is set to `error.message` verbatim and returned in the tool payload. That contradicts the posture `runTool` establishes in `shared.ts` (lines 92–102), where an unexpected error is passed to `reportError` and replaced with a generic "Internal error running the tool" precisely so internal detail never reaches the caller. Because `runUpdateIssue` swallows the error here, neither half of that happens:

- A Postgres/driver error (constraint text, column names, connection detail) goes straight out over the MCP wire.
- `reportError` is never called, so a partial-write failure on `update_issue` produces **no Sentry event**. A recurring failure on, say, `updateIssuePriority` would be invisible in observability while the tool keeps returning HTTP 200 with `partial: true`.

Suggest calling `reportError(error, { action: "mcp.tool.update_issue", userId: ctx.userId })` here and putting a fixed string in `failed.reason` (the field name already tells the caller which field stopped it, which is the actionable part).

src/lib/mcp/tools/update-issue.ts
```diff
@@ -0,0 +1,292 @@
...
+    result: {
+      machine: issue.machineInitials,
+      number: issue.issueNumber,
+      url: issueUrl(issue.machineInitials, issue.issueNumber),
+      applied,
+      // Each field commits in its own transaction, so a mid-loop failure leaves
+      // the earlier fields WRITTEN. Returning a blanket error would tell the
+      // caller nothing landed when some of it did, so this stays a success
+      // payload that names what failed (CORE-ARCH-012).
+      ...(failure ? { partial: true, failed: failure } : { partial: false }),
```
timothyfroehlich, line 323: **A partially-failed update is audited as `outcome: "ok"`.**

`runUpdateIssue` returns a success `ToolOutcome` when a field write throws mid-loop, so `runTool` logs `outcome: "ok"` with no reason. The MCP audit log is the only server-side record of what these tools wrote to production; a call that wrote `severity` and then failed on `priority` is indistinguishable in that log from one that applied everything. Combined with the missing `reportError` above, a systematic failure of one field leaves no trace anywhere. Consider threading a `partial` flag through `ToolOutcome` into `logMcpToolCall`.

src/lib/mcp/tools/update-issue.ts
```diff
@@ -0,0 +1,292 @@
...
+  "assignee",
+] as const;
+
+type UpdatableField = (typeof UPDATABLE_FIELDS)[number];
+
+/**
+ * Which matrix permission each field requires.
+ *
+ * `issues.update.reporting` covers what a reporter naturally owns — guests hold
+ * it on their OWN issues. `issues.update.triage` is the organizational half and
```
timothyfroehlich: **The comment describes ownership-scoped behaviour that this code can never produce.**

"guests hold it on their OWN issues" is true of the matrix (`issues.update.reporting` is `"own"` for guest), but `checkPermission` is called at line 137 with no `OwnershipContext` — and `helpers.ts` returns `false` for any conditional value when `context?.userId` is absent. So the `"own"` grant never fires here; a guest is denied on `title`/`status`/`severity`/`frequency` regardless of who reported the issue.

It fails closed and MCP is admin-gated today, so nothing is exploitable — but the comment sets the wrong expectation for whoever extends this surface, and the integration test that pins it (`mcp-tools.test.ts`, "denies a mixed update before writing any field") has a docstring describing a *member* scenario while actually exercising a *guest*, which suggests the confusion is already live. Either pass `{ userId: ctx.userId, reporterId: issue.reportedBy }` (which needs `resolveIssue` moved ahead of the gate) or state plainly that ownership-conditional grants are not honoured on this surface.

src/lib/mcp/tools/add-issue-comment.ts
timothyfroehlich: Fixed in `0fe73f19`. The file is text again — `git diff` shows it and `rg` no longer skips it.

Worth recording how it got there, since it will recur: writing the escape through the Write tool re-emitted a literal NUL both times I tried. I had to do the byte replacement in Python. The separator is now a named `KEY_SEPARATOR` constant with a doc comment saying why it has to stay an escape, which at least gives the next person a place to notice.

—Claude

src/lib/mcp/tools/get-issue.ts
timothyfroehlich: Fixed in `0fe73f19`. Took both halves of your suggestion: the query is `desc(createdAt)` + limit, reversed before returning, so the window is the tail; and the payload now carries `commentCount` (full thread length) and `commentsTruncated`. The `commentLimit` and tool descriptions say so explicitly.

Two integration tests pin it — one seeds three comments with explicit timestamps (three posted in a row can land in one clock tick, and the test is meaningless if the order it asserts is accidental) and asserts `["middle","newest"]` at `commentLimit: 2`; the other asserts `commentsTruncated: false` on a thread that fits. The count query carries the same `isSystem` filter as the page query, which the "excludes system rows" test now also checks.

—Claude

src/lib/mcp/tools/update-issue.ts
timothyfroehlich: Fixed in `0fe73f19`, exactly as suggested. `reportError(error, { action: "mcp.tool.update_issue", userId: ctx.userId })`, and `failed.reason` is now a fixed string naming the field.

One carve-out: an `McpToolError` passes its message through unchanged, since those are already curated user-facing text and reducing them would lose the actionable part. Everything else is reported and generic.

The existing partial-failure integration test now asserts the thrown text does *not* appear in the response.

—Claude

src/lib/mcp/tools/update-issue.ts
timothyfroehlich: Fixed in `0fe73f19`. `ToolOutcome` gains `auditOutcome` + `auditReason`, `runTool` reads them instead of hardcoding `"ok"`, and a partial update sets `auditOutcome: "error"` with `auditReason: "partial:<field>"` — so the log line names which field stopped it.

Also added `src/lib/mcp/tools/shared.test.ts`, the first direct test of `runTool`: four cases covering the default `ok`, the override, `denied` (asserting `reportError` is *not* called), and an unexpected error (asserting its text stays out of the response). That was the cheapest place to pin this, since it needed no service mocking.

—Claude

src/lib/mcp/tools/update-issue.ts
timothyfroehlich: Fixed in `0fe73f19` — took the second option (state plainly that ownership-conditional grants are not honoured here), not the first.

Reasoning against passing an `OwnershipContext`: the MCP caller is a bearer token acting for one operator, not a person browsing an issue they filed. There is no request-scoped reporter identity to compare against, so `{ userId: ctx.userId, reporterId: issue.reportedBy }` would grant on "the token's operator happens to be the reporter", which is a different fact than the one the `"own"` grant is meant to express. It also costs the clean property that the gate runs before `resolveIssue` and therefore before any write.

The comment now says both halves: which permission covers which field, and that both are checked on access level alone so a guest is denied either way, fails closed, and the door already requires admin.

The test docstring was the more useful catch — you're right that the confusion was live. It now describes the guest correctly, naming both reasons a guest fails (triage is flatly false; reporting is conditional and unresolvable without a context).

—Claude

timothyfroehlich:
Claude review of head 6a62ffd — `/code-review medium` — list_issues assignee filter used the write-eligibility resolver; get_issue could not distinguish a withheld thread from an empty one; update_issue kept dead snapshot fallbacks that fail open; unreachable null branch in the assignee filter

PR #1852, opened 2026-08-10
fix(docs): correct three false claims in the non-negotiables catalog (PP-fj8s, PP-sbo6, PP-tug9)

timothyfroehlich:
Three unrelated stale assertions in `docs/NON_NEGOTIABLES.md` and the tiers that copy from it. Grouped into one PR because they are the same class of defect in the same file — statements that were true when written and went false silently. Each was re-derived against `src/` or a live tool, not against another doc.

Swept all four tiers for each claim (`docs/NON_NEGOTIABLES.md`, `.claude/rules/`, `.agents/skills/`, `AGENTS.md`) with `rg --hidden`. `AGENTS.md` no longer restates the rules after #1834, so it needed no change. Dated records under `docs/superpowers/` left alone per AGENTS.md §8.

## PP-fj8s — CORE-A11Y-001 said the skip link was not implemented

It shipped. `src/app/layout.tsx:90-95` renders the anchor as the first child of `<body>`; `src/components/layout/MainLayout.tsx:137-140` and `src/app/(auth)/layout.tsx:17` carry the `id="main-content" tabIndex={-1}` target. The risk of leaving it: an agent told the rule is unimplemented could add a second skip link with a duplicate target id, or "fix" a working `MainLayout`.

**CORE-A11Y-006 (`inert`) re-derived at the same time and left alone** — it is still genuinely unimplemented. Zero `inert` usages in `src/`; `RowEditSheet.tsx:117` mentions it only in a comment, and its test asserts `aria-hidden`, which is the weaker behavior the rule exists to replace. Its Status line now carries that evidence so the next reader doesn't re-derive it. Those two are the only Status lines in the CORE-A11Y-001..006 block.

`.agents/skills/pinpoint-ui/references/accessibility.md` already carried the corrected version (fixed in PR 7 / PP-22e4), so no skill copy needed changing.

## PP-tug9 — CORE-ARCH-004 named a column that does not exist

The rule said: *"Enforce `machine_id` foreign key with CHECK constraint."* Both halves are false. `issues` joins `machines` on `machine_initials` (a natural key) — `src/server/db/schema.ts:281-283`, relation at `schema.ts:1022` — and enforcement is `NOT NULL` + FK `ON DELETE CASCADE`. There is no CHECK constraint and no `machine_id` column.

The `schema.ts` doc comment above the table said the same wrong thing (\"enforced by CHECK constraint\") and is presumably where the drift came from, so it is corrected too — comment-only, no schema change, no migration.

`docs/TECH_SPEC.md` also has a `machine_id UUID` + `CONSTRAINT issues_machine_required CHECK (...)` block, but that is a whole-schema snapshot from 2025-11-10 that has drifted far beyond this one rule (its `issues` table has no `issue_number`, no `priority`). Out of scope here — filed as a follow-up bead rather than half-fixed.

## PP-sbo6 — `fetchpriority` was on the wrong side of our own Baseline gate

**Live lookup, run today** (`npx -y modern-web-guidance@latest retrieve \"optimize-image-priority\"`, the mechanism CORE-UI-006 names for exactly this):

> Baseline status for Fetch priority: **Newly available**. It's been Baseline since **2024-10-29**.

So it is not Widely available until ~Apr 2027 — but CORE-UI-005, `.claude/rules/ui.md`, `pinpoint-ui/SKILL.md`, and design-bible §19 all listed it among the primitives to *reach for directly*. The bead recorded three different wrong dates previously in the repo (Sep 2023 ×2, Sep 2025); none were reused — the §19 table that held them has since been deleted, and the remaining defect was the tier, in prose.

### The opt-in question — decided as: adopt below the floor

`fetchpriority` **is** in real use: `src/app/(site)/page.tsx:35` passes `priority` to a `next/image`, which emits `fetchpriority=\"high\"`. So the choice was a real opt-in or stop using it.

Recorded as an opt-in in design-bible §19's existing **\"Adopted below the Baseline floor (safe-no-op progressive enhancements)\"** table. The reasoning: a browser that doesn't understand the attribute ignores it and falls back to its own image-priority heuristics — identical to omitting it. The worst case is \"the LCP image loads at default priority\", never a broken render, which is exactly the no-op criterion that section already defines. Dropping `priority` instead would be a genuine LCP regression for zero correctness gain. We also don't write the attribute by hand; Next.js does.

**Tim: this is the one judgment call in the PR — say so if you'd rather defer the feature than adopt it.** Everything else here is a fact correction.

`text-wrap: balance` moved into the same table while I was there. It was sitting in the *deferred* list annotated \"partially adopted (we use it per §9)\", which is the same inconsistency one step further along, and CORE-UI-005 was separately calling it \"grandfathered\" — two different words for one un-recorded opt-in.

### Not fixed here: `:has()` may be mis-tiered too, in the other direction

The same live tool reports `:has()` as **Newly available since 2023-12-19**. Thirty months from that date is 2026-06-19, which has passed — so either MWG's snapshot is stale or the tier really hasn't flipped. Every file in this PR lists `:has()` as Widely available. I did not change it: substituting a possibly-stale tool answer for a claim that is probably now correct would be trading one wrong assertion for another. Filed as a follow-up bead.

## Files touched

| File | Beads |
| :--- | :--- |
| `docs/NON_NEGOTIABLES.md` | all three (CORE-A11Y-001, CORE-A11Y-006, CORE-ARCH-004, CORE-PERF-003, CORE-UI-005) |
| `src/server/db/schema.ts` | PP-tug9 (doc comment only) |
| `.claude/rules/ui.md` | PP-sbo6 |
| `.agents/skills/pinpoint-ui/SKILL.md` | PP-sbo6 |
| `.agents/skills/pinpoint-design-bible/references/19-browser-support.md` | PP-sbo6 (the opt-in record) |
| `.agents/skills/pinpoint-design-bible/references/21-22-images-and-guidance.md` | PP-sbo6 |

## Verification

- `pnpm run check` — green (includes `check:rule-ids`, which #1834 added and which validates the rule files this touches)
- `pnpm run check:python` — green, 530 passed
- No behavior change: the only `src/` edit is a JSDoc comment.


docs/NON_NEGOTIABLES.md
```diff
@@ -473,7 +473,8 @@
...
+- **Do:** Reach for Baseline Widely available primitives directly: `<dialog>`, container queries, `:has()`, `:user-invalid`, `inert`, `aspect-ratio`, native form validation, CSS Grid `auto-fit/minmax`. No feature detection or polyfill needed. Derive a feature's tier **live** from `modern-web-guidance` (CORE-UI-006) rather than from a cached date — this list names the primitives, not their dates, for exactly that reason.
+- **Do:** Two Newly-available features are adopted below the floor with a recorded opt-in in `pinpoint-design-bible` §19 — `fetchpriority` (via `next/image`'s `priority`) and `text-wrap: balance` / `text-pretty` (used selectively per §9). Both degrade to a harmless no-op. Adding a third requires the same opt-in, in §19, in the PR that adopts it.
```
timothyfroehlich: `text-pretty` is not Newly available — it is Baseline **Limited**, so this new line records the same class of wrong tier the PR is fixing.

Live lookup just now (`npx -y modern-web-guidance@latest retrieve "improve-text-layout-and-legibility"`):

> Baseline status for text-wrap: balance: Newly available. It's been Baseline since 2024-05-13.
> text-wrap: pretty has limited availability.
> Supported by: Chrome 117 (Sep 2023), Edge 117 (Sep 2023), and Safari 26 (Sep 2025).

No Firefox, so `text-wrap: pretty` has never been Baseline at all. Concrete misread this causes: an agent reading this bullet concludes `text-pretty` cleared the same bar as `text-wrap: balance` and `fetchpriority` (both genuinely Newly available with a Baseline date), when it is in the same tier as `interestfor` and `closedby` — which the very next `Don't` bullet lists as needing more than a Newly-available opt-in. `text-pretty` is in production use (`src/components/ui/dialog.tsx:140`, `src/app/(site)/page.tsx:42`, `src/app/(auth)/reset-password/page.tsx:53`), so this needs to be recorded as a Limited-availability adoption, not folded in with `balance`.

—Claude

.agents/skills/pinpoint-design-bible/references/19-browser-support.md
```diff
@@ -40,10 +39,14 @@ Widely-available primitive already covers the same need as the floor. Each such
...
-| `interactive-widget=resizes-content` (viewport) | Limited (Chromium-only) | The browser default (`resizes-visual`) | `scrollIntoView`-on-focus + `scroll-margin` in the settings editors |
+| Feature                                         | Status                       | Degrades to                                 | Cross-browser floor                                                 |
+| :---------------------------------------------- | :--------------------------- | :------------------------------------------ | :------------------------------------------------------------------ |
+| `interactive-widget=resizes-content` (viewport) | Limited (Chromium-only)      | The browser default (`resizes-visual`)      | `scrollIntoView`-on-focus + `scroll-margin` in the settings editors |
+| `fetchpriority` (via `next/image` `priority`)   | Newly available (2024-10-29) | The browser's own image-priority heuristics | Next.js `<Image>` srcset/lazy-loading, which is unaffected          |
+| `text-wrap: balance` / `text-pretty`            | Newly available              | Normal line breaking                        | The `text-balance` utility is cosmetic-only (§9)                    |
```
timothyfroehlich: Same tier error as `docs/NON_NEGOTIABLES.md:477`, and this table is the canonical opt-in record, so it is the one that matters most. `text-wrap: balance` is Newly available (Baseline 2024-05-13); `text-wrap: pretty` is **Limited** — Chrome/Edge 117, Safari 26, no Firefox — per the live MWG lookup for `improve-text-layout-and-legibility`. Collapsing them into one undated "Newly available" row means the next reader who greps §19 for `text-pretty` gets a wrong answer from the file the policy names as authoritative. Split the row (or state Limited and give the reason it is still safe: unsupported browsers just wrap normally).

Related: the row for `fetchpriority` carries its Baseline date but this one does not, which reads as an oversight rather than a decision.

—Claude

docs/NON_NEGOTIABLES.md
```diff
@@ -325,8 +325,8 @@
 
 - **Severity:** Critical
 - **Why:** Core product requirement
-- **Do:** Enforce `machine_id` foreign key with CHECK constraint
-- **Don't:** Allow issues without machines
+- **Do:** Enforce the link on `issues.machine_initials` — `NOT NULL` plus a foreign key to `machines.initials` with `ON DELETE CASCADE` (`src/server/db/schema.ts`, `issuesRelations`). The join key is the natural key `machine_initials`, not a surrogate id.
+- **Don't:** Allow issues without machines. Don't reach for a `machine_id` column or a CHECK constraint on `issues` — neither exists, and `NOT NULL` + FK already enforces the rule.
```
timothyfroehlich: This is overbroad in a way that will misfire: `issues` **does** carry a CHECK constraint — `reporter_check` (`src/server/db/schema.ts:339-341`, added in `drizzle/0001_invited-users-rename.sql`), which enforces the reporter XOR rule. "Don't reach for a `machine_id` column or a CHECK constraint on `issues` — neither exists" therefore states something false about the table as a whole, in the exact catalog this PR is fixing for false statements.

Scenario: an agent adding a reporter-related invariant reads this line, concludes the codebase policy is "no CHECK constraints on `issues`", and either rewrites `reporter_check` as application-level validation or files it as drift. Scope the sentence to the machine link — e.g. "no CHECK constraint enforces the machine link, and there is no `machine_id` column."

—Claude

src/server/db/schema.ts
```diff
@@ -272,7 +272,9 @@ export const pinballmapCatalog = pgTable(
  * Issues Table
  *
  * Issues reported for pinball machines.
- * Every issue MUST have exactly one machine (enforced by CHECK constraint).
+ * Every issue MUST have exactly one machine (CORE-ARCH-004). Enforced by
+ * `machine_initials` being NOT NULL with a foreign key to `machines.initials`
+ * (ON DELETE CASCADE) — there is no CHECK constraint and no `machine_id`.
```
timothyfroehlich: Same overbroad claim as `docs/NON_NEGOTIABLES.md:329`, and here it is more misleading because the contradiction is 60 lines below in the same table declaration: `reporterCheck: check("reporter_check", ...)` at line 339. A doc comment saying "there is no CHECK constraint" sits directly above a table that has one.

Narrow it to the machine link: no CHECK enforces the issue→machine relationship (`NOT NULL` + FK does), and there is no `machine_id` column.

—Claude

docs/NON_NEGOTIABLES.md
```diff
@@ -495,7 +496,7 @@
 - **Why:** WCAG 2.4.1 Level A. Without it, keyboard users tab through the AppHeader navigation on every single page load before reaching content. PinPoint's header has 6+ tab stops; that's a hard daily-driver cost for any keyboard-first user.
 - **Do:** The first child of `<body>` (in `src/app/layout.tsx`) must be `<a href="#main-content" className="sr-only focus:not-sr-only ...">Skip to main content</a>`. The `<main>` element in `MainLayout.tsx` must carry `id="main-content"` and `tabIndex={-1}`.
 - **Don't:** Ship a layout that puts content behind the header in tab order with no skip affordance.
-- **Status:** Not yet implemented; implementation tracked under beads epic PP-kqbk (child PP-kqbk.3). New layouts must satisfy this rule.
+- **Status:** Implemented (PP-kqbk.3). `src/app/layout.tsx` renders the anchor as the first child of `<body>`; the `<main id="main-content" tabIndex={-1}>` target lives in `src/components/layout/MainLayout.tsx` and `src/app/(auth)/layout.tsx`. Both halves are load-bearing — a new layout shell must port the pair, not just the anchor.
```
timothyfroehlich, line 499: Minor: the new Status names three files, but the `Do:` line directly above it (unchanged) still says only "The `<main>` element in `MainLayout.tsx` must carry `id=\"main-content\"`", and the skill tier (`.agents/skills/pinpoint-ui/references/accessibility.md:11`) likewise describes the target as living only in `MainLayout.tsx`. `src/app/(auth)/layout.tsx:17` is a second, equally real target.

Since the PR's claim is a four-tier sweep per rule, the auth layout being visible only in the Status line leaves the actionable half (`Do:`) and the skill copy still describing a single target — the same partial-truth shape the PR is correcting elsewhere.

—Claude

docs/NON_NEGOTIABLES.md
timothyfroehlich: Confirmed via live `modern-web-guidance` lookup: `text-wrap: pretty` is Baseline Limited (Chrome/Edge 117, Safari 26, no Firefox — no Baseline date at all), distinct from `text-wrap: balance` (Newly available, Baseline since 2024-05-13). Split the CORE-UI-005 Do: bullet into three named features with their own tiers instead of grouping `balance`/`text-pretty` together.

—Claude

.agents/skills/pinpoint-design-bible/references/19-browser-support.md
timothyfroehlich: Split the §19 table row into two — `text-wrap: balance` (Newly available, 2024-05-13) and `text-wrap: pretty` (Limited, no Firefox) — and added a bullet explaining why Limited is still safe to adopt here: same no-op fallback behavior as `balance`.

—Claude

docs/NON_NEGOTIABLES.md
timothyfroehlich: Scoped the Don't bullet to the machine link specifically and added a parenthetical naming `reporter_check` so the sentence doesn't read as "issues has zero CHECK constraints."

—Claude

src/server/db/schema.ts
timothyfroehlich: Same fix as docs/NON_NEGOTIABLES.md:329 — scoped the doc comment to the machine link and noted `reporter_check` exists below for a different purpose.

—Claude

docs/NON_NEGOTIABLES.md
timothyfroehlich: Updated the Do: line in docs/NON_NEGOTIABLES.md and the pinpoint-ui accessibility reference to name both `MainLayout.tsx` and `(auth)/layout.tsx` as targets, matching the Status line.

—Claude

timothyfroehlich:
Claude review of head a401525 — `/code-review medium` — Five findings, all fixed in a4015256: text-pretty was given the wrong Baseline tier (Limited, no Firefox - not Newly available) in both NON_NEGOTIABLES.md and design-bible 19, the same defect class the PR fixes; the 'no CHECK constraint on issues' claim was overbroad (reporter_check exists) in both the catalog and the schema.ts doc comment, now scoped to the machine link; and the CORE-A11Y-001 skip-link target list was inconsistent with accessibility.md. Re-attested at the new head - the push was the review's own findings.

PR #1853, opened 2026-08-10
feat(hooks): point at the runner when an agent overrides the memory gate (PP-9ka9)

timothyfroehlich:
## What

When an agent hits the memory-pressure hard block and overrides it with `FORCE_MEM_PRECHECK=skip`, the hook now attaches a note pointing at the crabbox runner. **It does not block** — the override stays a legitimate escape hatch.

The block message already names crabbox (PP-ljx7), but by the time an agent has typed the override it has stopped reading that message — it's re-running the same command with a prefix. The override is the moment the reminder is worth delivering.

## Where it lives

**Extends `block-heavy-under-pressure.cjs` rather than adding a second hook.** That hook already owns the heavy-command classifier and already detects the inline override — the early return added so it doesn't deny a command the user explicitly told it to skip. A separate hook would duplicate the classifier and cost another process on every Bash call.

## What it deliberately won't say

A reminder is only honest when a job actually exists:

| Command | Reminder |
| --- | --- |
| `pnpm run test:integration` | `crabbox job run integration` |
| `pnpm run test:integration:supabase` | `crabbox job run integration-supabase` |
| `pnpm run smoke` | `crabbox job run smoke` |
| `pnpm run e2e` / `e2e:full` / `e2e:all` | `crabbox job run e2e-full` |
| `pnpm run build` | **silent** — heavy, but crabbox has no build job |
| `vitest run`, `playwright test` | **silent** — heavy, but the suite isn't knowable from the command |

And it's guarded on the CLI being on `PATH`, so a host with no runner — CI, Bazzite, another contributor's machine — sees nothing rather than advice it can't act on.

## Mechanism

Emits `hookSpecificOutput.additionalContext` with **no `permissionDecision`**, so the permission flow is untouched and the command runs either way. `additionalContext` is injected into the agent's context at the tool call.

## Verification

- 66 tests in `block-heavy-under-pressure.test.ts` (was 45), covering the job mapping, the silent cases, PATH-based presence detection, and the subprocess contract that no `permissionDecision` is emitted
- Tests control `$PATH` rather than trusting the machine's, so they mean the same thing locally and in CI. `runHook` now spawns via `process.execPath` — a bare `"node"` stopped resolving once `$PATH` was overridable
- `pnpm run check` clean; full unit suite 2124 passed
- Hook exercised end-to-end with a real payload

Hook and tests only; no runtime code touched, nothing to screenshot.

### DIFF
--- a/.claude/hooks/block-heavy-under-pressure.cjs
+++ b/.claude/hooks/block-heavy-under-pressure.cjs
@@ -150,7 +150,89 @@ function isHeavyCommand(command) {
   return segments.some(isHeavySegment);
 }
 
-module.exports = { isHeavyCommand };
+// --- Offload reminder (PP-9ka9) ----------------------------------------------
+// When an agent overrides the gate with FORCE_MEM_PRECHECK=skip, the block
+// message it already read is behind it — it is re-running the same command with
+// a prefix. The override is the moment a reminder is worth delivering, so this
+// attaches one to the tool call. It never denies: the override is a legitimate
+// escape hatch and stays one.
+//
+// A reminder is only honest when a job actually exists. `pnpm run build` is in
+// the heavy set but crabbox has no build job, so it maps to null and stays
+// silent. Anything ambiguous (bare `vitest run`, a raw `playwright test` whose
+// config we cannot infer) also maps to null rather than naming a job that might
+// run the wrong suite.
+const CRABBOX_JOB_BY_SCRIPT = new Map([
+  ["test:integration", "integration"],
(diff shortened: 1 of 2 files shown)


timothyfroehlich:
Closing — wrong tier. This encodes a fact about one machine (crabbox, which only the Mac can reach) in the repo, which is the rule PP-ljx7 just established. Rebuilding it as a machine-level hook in Tim's dotfiles instead. No repo change is needed for it to work. —Claude

PR #1854, opened 2026-08-10
docs(agy): redesign the agy reviewer around sliced fan-out and refutation (PP-c6xz)

timothyfroehlich:
Replaces the approach on #1811, which is not being continued. Nothing here runs in CI or ships to users — it is two agent skills (both outside the repo, in `~/.gemini` and `~/.claude`), the version-controlled record of agy's permission profile, and the measurement log that justifies the design.

## The problem

The single-pass agy reviewer scored **0 of 4** against PR #1825's known findings — on `gemini-3.6-flash-high` **and** on `gemini-3.1-pro-high`, so it was not a model-tier problem. Pro didn't merely miss the high-severity one, it asserted the negation at 21k output tokens of deliberation.

A direct probe settled the cause. The same model, same commit, same permissions, asked two **pointed** questions — "trace this exact sequence and state the end state", "does this client comply with what §7 requires" — answered both correctly in 23 seconds using two commands. It can reason about those bugs. It does not go looking for them when handed "review this PR".

## The redesign

- Claude picks **4–5 concern slices** — domains, never file lists, and never a suspected defect. Claude owns coverage; recall inside a slice stays agy's, so a reviewer isn't reduced to confirming hypotheses.
- agy fans one subagent out per slice. Every brief carries an obligation to **trace a concrete sequence end to end and state the end state** before concluding anything, including "clean".
- A **second wave of refuters**, dispatched as a separate run so a refuter cannot see how the finding it judges was arrived at.
- The calibration section no longer licenses an empty result, and a **valid** nit is now reportable at `low` while an invented one still isn't.

## Result

**2 of 4** on the same commit, plus one finding outside ground truth that survived refutation and looks real, and one false positive the refuters killed. The two it found are the two expensive ones: the drift bug pro had called flawless, and a connection leak.

The `prepare: false` miss has a candidate fix already in — #1834's `.claude/rules/` tier puts that rule in a short subject-scoped file, and the skill now sends the rules slice there first instead of at the 20-rule catalogue. Untested; the next calibration run should re-check it.

## Measured facts, all in the plan doc

- Subagents work under the locked-down profile, and `invoke_subagent` takes an **array** — one call fans out N in parallel.
- A subagent that hits an unconfigured command is **canceled, not fatal**, and **its blocked command never reaches the run log**. It only surfaces if the parent messages the idle subagent and relays it.
- **`status: ERROR` with an empty `.result.response` does not mean the run failed.** The complete review sits in the `text_delta` stream. Reconstruct with `jq -j`; `-r` inserts a newline per delta and corrupts the JSON.
- The parent will **fabricate a clean pass** if you let it — after one run died mid-fan-out, the resume made zero tool calls, contacted no subagent, and returned `approve` / `findings: []` / `blocked_slices: []` with a confident summary. agy's skill now carries that incident as a worked example.

## Also

Strips the huddle wiring from `.agents/hooks/antigravity-bootstrap.cjs`. A review run followed it into `bash scripts/hooks/huddle-bootstrap.sh` mid-review — correct instruction-following, but a read-only reviewer has no business running shell scripts.

Closes PP-c6xz.

https://claude.ai/code/session_01LcSvtVRP93799GgNLzArqi


timothyfroehlich:
Closing for now — the agy reviewer work has stalled out and these two PRs both target PP-c6xz with conflicting branches. Reopen or start fresh from main when the work resumes. The bead (PP-c6xz) stays open. —Claude

PR #1855, opened 2026-08-10
docs(agents): audit the 15 skill descriptions against their post-sweep bodies (PP-22e4)

timothyfroehlich:
docs(agents): audit the 15 skill descriptions against their post-sweep bodies (PP-22e4)

Task 9 of the deletion sweep, and the last one. The sweep rewrote bodies
across nine PRs; descriptions are the routing signal and were never re-derived
against what the bodies ended up containing.

Eleven descriptions changed, four left alone. `pinpoint-deployment`,
`pinpoint-security`, `pinpoint-superpowers-bridge` and `pinpoint-typescript`
audited clean — every clause maps to a body heading and no substantial section
goes unnamed.

**The dominant failure was under-claiming, not over-claiming.** Every skill's
description was accurate about what it named; the gaps were real body sections
with no routing term, so the skill would never load for the thing it answers:

- `pinpoint-briefing` named only the audit-reading caveats. Three of its five
  parallel groups — main-branch CI, new GitHub issues, security-review beads —
  had no term at all.
- `pinpoint-chores` omitted two of ten checklist items outright (prod backup
  validation, the memory review) and folded the pnpm corepack pin, which has
  its own cooldown and its own sites, into "Supabase CLI pin".
- `pinpoint-pr-workflow` omitted the GitHub MCP gotchas and the merge escape
  hatches. An agent whose label write had just replaced the whole label set had
  nothing routing it here.
- `pinpoint-orchestrator` omitted worktree reaping and cleanup entirely.
- `pinpoint-e2e` omitted the content `pinpoint-testing` explicitly defers to it
  — pre-flight bug-class routing, selector strategy, the auth roles and
  `loginAs` scaffolding.
- `pinpoint-testing` omitted the PGlite `~/server/db` forwarding pattern.
- `pinpoint-ui` omitted the color system: semantic tokens only, dark-only.
- `pinpoint-design-bible` omitted the shell contract and the two-reds AA rule.
- `pinpoint-huddle` omitted the work digest, the volume knobs and the
  multi-machine Dolt semantics.

Two are safety-shaped rather than convenience-shaped:

- `pinpoint-prototype-mode` described everything the mode relaxes and nothing
  it does not. An agent routing on the description alone could not tell that
  auth, permissions and data-privacy rules still apply in full and that
  committing is banned while the marker exists.
- `pinpoint-memory-review` never surfaced that auditing skill contents is out
  of scope by decision — the exclusion most likely to misroute a task like
  this one.

**Collisions (step 2): none found.** briefing / chores / memory-review each
own their maintenance topic outright, with two hand-offs already recorded in
the bodies (Sentry+Supabase advisors moved briefing→chores; memory pruning
delegated chores→memory-review). ui↔design-bible and testing↔e2e each had one
side stating the split; the other side now states it too.

**Gaps (step 3): nothing unrouted.** The sweep's deletions are all still
reachable — logging is self-hosted in `pinpoint-ui` and never depended on the
deleted `docs/LOGGING.md`; DEVELOPMENT.md's survivors went to README, AGENTS.md
§5 and the testing skill; general TypeScript technique is uncovered on purpose
and `pinpoint-typescript` says so.

Also fixes drift from #1834: AGENTS.md §2.1 still called `always.md` "the four
rules", which became six when CORE-ARCH-008 and CORE-ARCH-011 moved there.

---

Task 9 of `docs/superpowers/plans/2026-08-01-context-corpus-deletion-sweep.md` — the final task. Closes the sweep.

**Method.** Three read-only auditors, five skills each, given a fixed brief: read the full body, report what it covers, which description terms the body does not answer, which body topics the description never names, and any ownership claim. Lead did the collision and gap analysis across all fifteen and wrote the descriptions.

**One auditor finding was rejected rather than applied.** The brief said to read `references/` filenames only, so one auditor counted "this content lives in the skill's own references/ file" as unsupported and reported `pinpoint-e2e` as the batch outlier — three headline clauses supposedly absent. Grepping the references directly found all of them in `references/e2e-best-practices.md`. A skill is one unit: the body is an index and the references are one hop away by design, so intra-skill delegation is not a routing bug. That whole column was an artifact of my brief.

Docs-only, 12 files, one line changed in each.


.agents/skills/pinpoint-ui/SKILL.md
```diff
@@ -1,6 +1,6 @@
...
...nts, Client Components, form handling, Tailwind CSS v4, accessibility. The color system is here: semantic tokens only, never a raw palette class or hex, and the app is dark-only so `dark:` variants are wrong. Also the Radix Select form-reset footgun (useActionState dispatch, PP-0fvr/PP-1ajq), CREATE form reset (return-redirect + dual-pass reset), the native `<select>` stale-option-ID silent fallback (PP-lql), authoring config-driven enums with rich metadata (labels/icons/styles), discriminated-union component props, Server Action conventions (`Action` suffix, `checkPermission`, `Result`, `serverActionError`), colocated data access with React `cache()` and `revalidatePath`, derived machine status, and why `console.*` is correct inside client components. Use when building UI, forms, components, badges, server actions, or data fetching, or when user mentions UI/styling/components/forms/enums/dropdowns/logging. Design-system depth — page archetypes, spacing rhythm, severity vocabulary, the form-correctness conventions themselves, and whether a web platform feature clears the Baseline floor — is owned by `pinpoint-design-bible`; this skill states the rules and how to build against them.
```
timothyfroehlich, line 3: The new clause "The color system is here: semantic tokens only" claims ownership of a body section that contradicts the authority it defers to. Body line 32 ("Use `bg-primary`, `text-destructive`, etc.") and line 53 (`text-destructive` in the list of correct semantic tokens) both hold up `text-destructive` as the right token — but `pinpoint-design-bible` §18, whose description this same PR rewrites to say "reaching for `text-destructive` ships a contrast failure", measures it at 3.96:1 on the dark background and requires `text-destructive-text` for destructive text. Concrete failure: an agent routed here by the new clause for error copy copies line 32 and ships an AA failure — the exact class PP-mjms is sweeping. Either fix the two examples to `text-destructive-text` (fill vs text) or drop the ownership claim. —Claude

.agents/skills/pinpoint-memory-review/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: pinpoint-memory-review
-description: Weekly curated pass over everything recorded about how we work — beads memories, Claude auto-memories on both the Mac and Bazzite, and the canonical context files. Prunes stale facts, promotes and demotes between tiers, dedupes across tiers, and hands Tim a short veto list. Use when the weekly chores checklist reaches the memory item, or when Tim says "memory review", "review memories", or asks what has been recorded.
+description: Weekly curated pass over everything recorded about how we work — beads memories, Claude auto-memories on both the Mac and Bazzite, and the canonical context files. Prunes stale facts, promotes and demotes between tiers, dedupes across tiers, hands Tim a short veto list, then applies the verdicts on both machines. Use when the weekly chores checklist reaches the memory item, or when Tim says "memory review", "review memories", or asks what has been recorded. **Auditing the contents of skills is explicitly out of scope** — this pass covers recorded facts, not the skill corpus.
```
timothyfroehlich, line 3: Stale conditional left in the body this audit covered: line 38 still reads "`rules` (whose `exists` is `false` until PP-22e4 PR 8 lands)". PR 8 landed — `.claude/rules/always.md` and `README.md` exist on main, which is what this PR's own AGENTS.md hunk is correcting for. Concrete effect: a memory-review run reads that parenthetical and treats an `exists: false` for rules as expected rather than as a broken collector. Since this PR is the last task of the sweep and is already fixing #1834 drift in AGENTS.md, this is the same drift one file over. —Claude

.agents/skills/pinpoint-design-bible/SKILL.md
```diff
@@ -1,6 +1,6 @@
...
...y (cosmetic / minor / major / unplayable — never low/medium/high or critical). Sole owner of the browser-support policy (§19 — the Baseline Widely available floor, deferred Newly-available features, and the per-feature opt-in), the form-correctness conventions (§20 — required input attributes, autocomplete tokens, validation-feedback timing), the picker pattern and component inventory (§12), and the modern-web-guidance lookup (§22). Also the shell contract (§3 — the fixed header and tab-bar dimensions and the z-index hierarchy every overlay has to sit inside) and the canonical token forms (§18), including the two reds: `bg-destructive` is the fill and is correct behind white, but that same red as text on the dark background measures 3.96:1 and fails AA, so destructive text takes `text-destructive-text` — reaching for `text-destructive` ships a contrast failure. Use when building any new UI, page, or component to ensure visual consistency; when naming, labelling, or writing copy for issue severity levels; when positioning anything against the app shell or picking a z-index; when deciding whether a web platform feature is safe to use; or when writing a form or a single-select picker.
```
timothyfroehlich: Length note, not a factual error: this description goes from 844 to 1357 characters, making it the second longest in the corpus. The listing that carries these to the model truncates — `pinpoint-deployment` (1758 chars) is cut mid-sentence at ~1535 characters in the live skill listing, losing the tail of its "Use when" clauses, which are the routing triggers. 1357 is still under that, but the trailing triggers added here ("when positioning anything against the app shell or picking a z-index") sit in the part that disappears first, and the Agent Skills spec documents a 1024-character description limit that other tools in this repo's cross-tool set may enforce more strictly than Claude Code does. Worth trimming the `text-destructive` explanation to its conclusion and leaving the 3.96:1 derivation to §18. —Claude

.agents/skills/pinpoint-ui/SKILL.md
timothyfroehlich: Fixed in 21a99971. Both sites now use `text-destructive-text` — line 32 in the Quick Reference and line 53 in the Color System. You are right that the description change made this worse rather than merely coexisting with it: routing more agents to this skill for color while the skill hands them the failing token is the bug.

—Claude

.agents/skills/pinpoint-memory-review/SKILL.md
timothyfroehlich: Fixed in 21a99971. Line 38 now reads plain `rules`. Line 110 carried the same class of rot and is fixed too — it claimed "PP-22e4 reduces AGENTS.md to a ≤10-line stub with a CI gate", which is dead twice over: the stub was superseded by the 2026-08-01 replan, and the byte-budget gate was dropped for good today. Replaced with the reason that does not expire.

—Claude

.agents/skills/pinpoint-design-bible/SKILL.md
timothyfroehlich: Fixed in 21a99971, and this was the most useful of the three. `pinpoint-deployment` is 1774 → 1291 and `pinpoint-design-bible` 1371 → 1137, no routing term lost — the cuts are clauses restating a topic already named earlier in the same description.

I confirmed the truncation independently before cutting: `pinpoint-deployment` arrives in this session own skill listing cut off mid-sentence at "when a PR audit job…", so its whole trailing trigger list never reaches the model.

This reverses the call I made opening the PR. I left deployment alone because an auditor found every clause mapped to a body heading. Wrong test — a load-bearing clause past the truncation point is not load-bearing, it is absent.

—Claude

timothyfroehlich:
Claude review of head 21a9997 — `/code-review medium` — 3 low findings: pinpoint-ui body taught text-destructive (the token that fails AA at 3.96:1) as its semantic-token example; stale PP-22e4 conditionals in pinpoint-memory-review; two descriptions past the listing truncation point. All fixed.

PR #1856, opened 2026-08-10
fix(pinballmap): record the listing a retitle leaves on the public map (PP-l81u)

timothyfroehlich:
## Summary

Retitling a machine that is listed on Pinball Map used to clear `pinballmap_listed` and discard `pinballmap_lmx_id`. That is correct about the **new** title — but the entry for the **old** title is still live on APC's public lineup, and the lmx was the only handle PinPoint had for ever finding or removing it. The desync check couldn't see it either: it keys on a machine's *current* title, and nothing pointed at the old one any more.

This records the abandoned entry instead of discarding it, surfaces it on the machine's Info tab, and clears the record automatically once the entry leaves the lineup.

It also closes the quieter half of the bead: `resolvePbmLinkColumns` took the listing flags as optional arguments defaulting to off, so **any** new write path that forgot to thread them silently unlisted a machine. Exactly one caller got it right, by hand.

## What changed

- **New table** `pinballmap_abandoned_listings` (migration `0063`) — one row per abandoned entry: which machine abandoned it, the lmx, the old catalog title, when. A machine can abandon several over time, so this cannot be columns on `machines` (and the `lmx_requires_listed` CHECK forbids reusing `pinballmap_lmx_id` anyway).
- **The resolver splits by intent** — a create entry point that cannot express listing state at all, and an update entry point that takes the stored row and decides carry-over itself. No caller computes `linkUnchanged`; nobody can pick the behaviour by omitting an argument. It returns the abandonment *alongside* the columns so a future caller can't apply one without the other.
- **Recorded in the same transaction** as the machine columns, in both of `updateMachineAction`'s transaction blocks. Local writes only — nothing added inside a transaction that CORE-ARCH-011 forbids.
- **Self-clearing** — `reconcileAfterSync` drops rows whose lmx is gone from a freshly synced lineup. Reads use `PINBALLMAP_API_TOKEN`, not the operator write credentials, so this works for an operator who never provisioned outbound writes. **There is deliberately no dismiss action** — the record verifies itself.
- **Surfaced** on the machine PinballMap card, gated on `canLink` like the existing desync alert.

## Design decisions worth knowing

Full spec: `docs/superpowers/specs/2026-08-09-pbm-orphaned-listing-design.md`.

**The unsynced state is legal.** PBM write credentials are optional configuration, and `applyAutoLinkWrite` marks machines listed with no credentials at all — so an operator who never set up outbound writes still accumulates listed machines. A retitle is therefore **never blocked**; blocking would hold an unrelated core workflow hostage to an optional integration, and would strand exactly the operator who cannot unlist through PinPoint.

**No cleanup control here.** Removal is manual on pinballmap.com. Split to **PP-o355.32**, which is cheaper than it looks — `client.removeMachine` already shipped in #1818; what's missing is a variant working from an abandonment record rather than a machine's current listing columns.

**Cascade on machine delete.** That leaves a live entry with no PinPoint record; that class belongs to the fleet dashboard (PP-o355.7.2) and PP-o355.31.

## Test plan

- [x] Unit **2109/2109**, integration **591/591** (main is 573 — the 18 new tests are all present)
- [x] `pnpm run check` clean
- [x] Full E2E on the Bazzite runner: **264 passed / 10 failed**. Those 10 are **pre-existing** — I baselined unmodified main on the same runner and got the identical 10 failures (main: 260 passed / 10 failed). Filed as **PP-168u**. The delta is exactly the 4 tests this branch adds.
- [x] The new E2E proves itself: with `|| showAbandoned` removed the spec **fails**; restored, it **passes**. See below.

## The bug that nearly shipped

The card renders only when `machine.pinballmapListed || showDesync`. A machine that just abandoned an entry is **not listed**, and `derivePbmMachineStatus` correctly reports it `ok` — so both were false *exactly* when the notice was needed. The feature would have been invisible to every user while every test stayed green.

Caught by inspection, not by a failing test — which is why `e2e/full/machine-pinballmap-abandoned-notice.spec.ts` exists: it goes through the real page, and it is the only layer that can see this line.

## Review notes

Three findings from an adversarial review are fixed here rather than deferred:

1. **`onConflictDoNothing` mis-attributed orphans.** If machine A abandons an lmx and machine B is later auto-linked onto that same live entry, B's insert silently no-opped while still writing B a timeline event — the orphan showed on A's page, not B's.
2. **A reclaimed lmx left a notice telling someone to destroy a live listing.** Nothing deleted a row when a machine *took the lmx back*, and auto-link exists to do precisely that. Retitle by mistake, retitle back, and the machine's own card told its owner to remove its own live listing. Fixed in the **clear query**, not the four capture sites, so the next capture path can't reintroduce it.
3. **A parseable-but-degenerate PBM payload wiped every record.** `parseLocation` tolerates a missing `location_machine_xrefs` key, so a 200 response with a renamed field yielded an empty lineup — and the empty-lineup branch then deleted everything and logged it as cleaned up. Now cross-checked against PBM's own `machine_count`.

Deferred, deliberately: the notice still says "clears itself once it is gone" when the integration is disabled, and the catalog-miss fallback reads "Pinball Map entry 4471 is still on Pinball Map." Both are copy, and copy belongs to PP-o355.21.

Also filed **PP-o355.33** — pre-existing: `captureAutoLink` updates by machine id without re-checking the title, so a retitle landing between the auto-link decision and its write leaves a machine listed on its old title's lmx. Neither CHECK nor the one-lister index catches it.

*(Branch is named `docs/…` because it started as the spec commit and grew; the content is a fix.)*

Closes PP-l81u

https://claude.ai/code/session_01Y39wAcWjeH7BKkB8p96eqo


timothyfroehlich:
## UI Screenshots — head `877a5ca2`

### Issues List

| --- | --- |
| desktop | mobile |

| --- | --- |
| desktop | mobile |

### Report an Issue

| --- | --- |
| desktop | mobile |

### Dashboard

| --- | --- |
| desktop | mobile |

| --- | --- |
| desktop | mobile |

### Collections

| --- | --- |
| desktop | mobile |

| --- | --- |
| desktop | mobile |

### New Machine

| --- | --- |
| desktop | mobile |

| --- | --- |
| desktop | mobile |

timothyfroehlich:
## Abandoned-listing notice — captures

The standard manifest comment above cannot show this one: it shoots `/m/TAF` as a **member**, and `machines.pinballmap.link` is owner-only for members, so the card's render gate is false for that viewer. These were shot separately as a maintainer, against a dev DB hand-seeded into the state the bead exists for — TAF pointed at catalog title 6222 ("Godzilla (Premium)"), unlisted, with an abandonment record for lmx 4471 under title 6221 ("Godzilla (Pro)").

The notice reads: **"Godzilla (Pro) is still on Pinball Map. Remove it there — this notice clears itself once it is gone."**

| desktop (1440×900) | mobile (390×844) |
| --- | --- |
| desktop card | mobile card |

Full pages for placement in the rail: [desktop](https://raw.githubusercontent.com/timothyfroehlich/PinPoint/pr-screenshots/pr-1856/21ec03c4/notice-desktop.png) · [mobile](https://raw.githubusercontent.com/timothyfroehlich/PinPoint/pr-screenshots/pr-1856/21ec03c4/notice-mobile.png)

—Claude-PbmListUnlist

timothyfroehlich:
Claude review of head 877a5ca — `/code-review high` — 5 findings. Fixed 3: (1) the reclaim delete existed only in applyAutoLinkWrite — the manual link, outbound list and verify/heal all set pinballmap_listed alongside an lmx without it, and two of those genuinely reclaim (PBM returns the existing lmx on an add; a heal claims a re-minted id), so a card told its owner to remove an entry another machine held live; now one retireAbandonmentForLmx helper at all four sites, with an integration test through the real list action. (2) getCatalogEntry ran on every Info view including anonymous landings that render no card; now gated on the card. (3) the status line claimed 'not linked' for a linked machine whose catalog row is missing; now names the title id. Declined 2, both replied on-thread: re-pointing lmx_id after a PBM reissue (we cannot tell which live entry IS the orphan, and the new predicate already clears once someone claims it), and machine-delete handling (spec decision 7 — cascade, owned by PP-o355.7.2/PP-o355.31).

timothyfroehlich:
## Abandoned-listing notice — reworked copy (`382aca33`)

Supersedes the captures earlier in this thread. The card now states the machine's own standing before naming the leftover title, because the notice necessarily names a title that is **not** this machine's — on The Addams Family's page that read "Godzilla (Pro)" and looked like a bug rather than a task.

```
◆ PINBALL MAP
Linked as Godzilla (Premium) · not listed.
⚠  Previous listing still live: “Godzilla (Pro)”
   ▸ Why is this here?
```

The quotes are load-bearing — they mark the odd name as a reference to something named elsewhere. The disclosure is a native `<details>`, so no client component, and it works on touch where a tooltip would not.

Prior-art research ruled out the obvious labels: every shipped product using "dangling" / "orphaned" / "drifted" is ops-facing (Cloudflare DNS, Terraform, Kubernetes); AWS and Terraform keep that vocabulary backstage as a filter, and Dropbox/Google skip the state noun entirely and lead with the verb.

| desktop (1440×900) | mobile (390×844) |
| --- | --- |
| desktop card | mobile card |

Full pages: [desktop](https://raw.githubusercontent.com/timothyfroehlich/PinPoint/pr-screenshots/pr-1856/382aca33/notice-desktop.png) · [mobile](https://raw.githubusercontent.com/timothyfroehlich/PinPoint/pr-screenshots/pr-1856/382aca33/notice-mobile.png)

Shot as a maintainer against a hand-seeded dev DB; the manifest run cannot reach this state because it shoots `/m/TAF` as a member and `machines.pinballmap.link` is owner-only for members.

—Claude-PbmListUnlist

src/lib/pinballmap/sync.ts
```diff
@@ -79,6 +85,15 @@ async function applyAutoLinkWrite(
...
     .where(eq(machines.id, write.machineId));
+  // Capturing this lmx IS the reclaim `clearResolvedAbandonments` looks for, so
+  // retire any abandonment record for it here rather than leaving it to the next
+  // hourly pass. The window matters: `captureAutoLink` also runs post-commit
+  // from `updateMachineAction`, so retitling a listed machine away and back
+  // would otherwise leave its own card telling the owner to go remove an entry
+  // the same save just re-listed it on (CORE-ARCH-012).
+  await tx
+    .delete(pinballmapAbandonedListings)
+    .where(eq(pinballmapAbandonedListings.lmxId, write.lmxId));
```
timothyfroehlich: **Medium — the reclaim delete is applied at only one of four sites that claim an lmx.**

This inline delete is right, and the reasoning above it ("the window matters... CORE-ARCH-012") applies verbatim to three other writes that set `pinballmap_listed = true` alongside an lmx, none of which retire the record:

- `src/app/(app)/m/pinballmap-actions.ts:266` -- `linkPinballmapEntryAction` takes `lmx.id` straight off the stored snapshot, so it can claim an lmx a machine abandoned.
- `src/app/(app)/m/pinballmap-actions.ts:546` -- `listMachineOnPinballMapAction`; PBM returns the **existing** lmx when the entry is already on the lineup, so an "add" reclaims the orphan.
- `src/app/(app)/m/pinballmap-actions.ts:870` -- `verifyPinballmapLinkAction`'s heal, the same `reconnected` write `applyAutoLinkWrite` handles here.

Scenario: machine A retitles off lmx 4471 (record written). A maintainer then connects machine B to that title via `linkPinballmapEntryAction`, reclaiming lmx 4471. Until the next hourly `clearResolvedAbandonments`, A's Info card says "Previous listing still live: Godzilla (Pro)" with a disclosure telling the reader to remove it on pinballmap.com -- and following that instruction deletes B's live listing.

Fixed in the working tree: the same delete added inside each of those three transactions.

---
*Claude Code review*

src/app/(app)/m/[initials]/(tabs)/page.tsx
```diff
@@ -102,25 +104,59 @@ export default async function MachineInfoTab({
...
+    })
+  );
+
+  // The machine's own current standing, stated above any alert. The abandoned
+  // notice necessarily names a DIFFERENT title than this machine's, so without
+  // this line the reader's first thought is "wrong machine", not "task to do".
+  const linkedTitleEntry =
+    machine.pinballmapMachineId === null
+      ? null
+      : await getCatalogEntry(machine.pinballmapMachineId);
```
Comment (timothyfroehlich) [inline src/app/(app)/m/[initials]/(tabs)/page.tsx]: **Low -- unconditional catalog read on every Info page view, including ones that render no card.**

`getCatalogEntry` is a plain `db.select(...).limit(1)` with no React `cache()` wrapper, so this is a real round trip per render. It runs for any machine with a `pinballmapMachineId`, including an anonymous QR-scan landing on an unlisted, non-desynced machine -- where `pinballmapCard` is `null` and the value is discarded. That is the exact cost the `pbmState` read 50 lines up is deliberately gated to avoid ("so the public QR-scan landing never pays for (or shows) it").

Fixed in the working tree: `showPinballmapCard` is hoisted above this and the read is gated on it.

---
*Claude Code review*

src/app/(app)/m/[initials]/(tabs)/page.tsx
```diff
@@ -102,25 +104,59 @@ export default async function MachineInfoTab({
...
+  // would never render for the one case this bead exists to surface.
+  const showAbandoned = canLink && abandoned.length > 0;
   const pinballmapCard =
-    machine.pinballmapListed || showDesync ? (
+    machine.pinballmapListed || showDesync || showAbandoned ? (
       <MachinePinballmapCard
         locationUrl={pinballmapLocationUrl()}
         desynced={showDesync}
         desyncReason={pbmStatus.reason}
+        linkedTitle={linkedTitleEntry?.name ?? null}
```
Comment (timothyfroehlich) [inline src/app/(app)/m/[initials]/(tabs)/page.tsx]: **Low -- a missing catalog row makes the status line contradict the card it sits in.**

`linkedTitle` collapses two different states into one: "no PBM link" and "linked, but the catalog mirror has no row for that title". The card renders `Not linked to a Pinball Map title.` for both.

Scenario: `machine.pinballmapListed` is true (which is *why* the card renders) but `pinballmap_catalog` has no row for `machine.pinballmapMachineId` -- a DB seeded from a prod dump before the weekly `refresh-catalog` cron has run, or a title PBM retired. The user sees "View on Pinball Map" plus "Not linked to a Pinball Map title." `resolvePbmLinkColumnsForUpdate` treats this state as reachable (it rejects the save with "no longer in the catalog"), and the abandoned notice right below has an explicit `Pinball Map entry #<lmxId>` fallback for the same reason -- the status line has none.

Fixed in the working tree: falls back to `Pinball Map title #<id>` when the machine is linked but the catalog row is missing, matching the notice's idiom.

---
*Claude Code review*

src/lib/pinballmap/abandoned-listings.ts
```diff
@@ -0,0 +1,204 @@
...
+        ? sql`true`
+        : notInArray(pinballmapAbandonedListings.lmxId, liveLmxIds);
+
+    conditions.push(
+      unclaimedTitleIds.length === 0
+        ? lmxIsGone
+        : sql`${lmxIsGone} AND ${notInArray(
+            pinballmapAbandonedListings.pinballmapMachineId,
+            unclaimedTitleIds
+          )}`
```
timothyfroehlich, line 220: **Low -- the record's `lmxId` is never re-pointed when PBM reissues the row id, so the notice can outlive the entry it names.**

The reissue branch deliberately holds a record open when its own lmx is gone but its title still has an unclaimed entry. It never updates `lmx_id` to the new one, and two consequences follow:

1. When the catalog row for the title is absent, the card falls back to `Pinball Map entry #4471` -- an id that no longer exists on pinballmap.com. A maintainer searching for it finds nothing and has no way to act on the notice.
2. "Old lmx gone, new unclaimed lmx under the same title" is indistinguishable from "someone removed the orphan, then someone else added a fresh, wanted entry for that title." In the second case the notice stays up indefinitely (there is deliberately no dismiss control) telling a maintainer to remove a listing that is not the abandoned one -- the CORE-ARCH-012 hazard the reclaim clear elsewhere in this file exists to prevent.

Not fixed -- re-pointing `lmx_id` at the surviving entry under the title is a design call, not a mechanical one. Flagging so it is a decision rather than an oversight.

---
*Claude Code review*

src/server/db/schema.ts
```diff
@@ -268,6 +268,54 @@ export const pinballmapCatalog = pgTable(
...
+ * removes the entry by hand on pinballmap.com. There is deliberately no dismiss
+ * action — the record verifies itself.
+ */
+export const pinballmapAbandonedListings = pgTable(
+  "pinballmap_abandoned_listings",
+  {
+    id: uuid("id").primaryKey().defaultRandom(),
+    machineId: uuid("machine_id")
+      .notNull()
+      .references(() => machines.id, { onDelete: "cascade" }),
```
timothyfroehlich, line 298: **Low -- deleting a machine is a third exit path that abandons a live entry, and the cascade also discards records already written.**

`deleteMachineAction` (`src/app/(app)/m/actions.ts`) issues a plain `.delete(machines)` with no PBM handling: it neither unlists on pinballmap.com nor records an abandonment, so deleting a listed machine leaves the entry live with nothing tracking it -- the same defect class this table exists to close for the retitle path. `ON DELETE cascade` then also drops any record the machine had already accumulated, so an orphan that *was* tracked becomes untracked.

The cascade itself is the right call given `machine_id NOT NULL`; the gap is that machine deletion never gets a chance to write a record first. Adjacent to this diff rather than in it, so not fixed here -- worth a follow-up bead if delete-while-listed is reachable in the UI.

---
*Claude Code review*

src/lib/pinballmap/sync.ts
timothyfroehlich: Fixed in 09bae1ab, and extracted rather than copied a fourth time — `retireAbandonmentForLmx` in `abandoned-listings.ts` now carries the reasoning once and all four sites call it. Covered by an integration test through the real list action against a lineup that already carries the abandoned lmx. Good catch: two of those three sites can genuinely reclaim, since PBM hands back the existing lmx on an add and a heal claims a re-minted id. —Claude-PbmListUnlist

src/lib/pinballmap/abandoned-listings.ts
timothyfroehlich: Declining, and I think the second half is already handled by the predicate as written. Clearing requires the lmx to be off the lineup AND every live entry under that title to be CLAIMED by a listed machine — so "orphan removed, fresh entry added under the same title" clears as soon as some machine claims it. It only persists while that new entry is unclaimed, which is itself a state worth surfacing rather than hiding. On re-pointing `lmx_id`: the record's job is to name a thing on the public map, and after a reissue we cannot tell which live entry under that title IS the orphan without a fingerprint PBM does not give us. Writing our guess into the record would make a wrong id look authoritative. The stale id only ever surfaces when the catalog also lost the title, which is the rarer branch of an already-rare state. —Claude-PbmListUnlist

src/server/db/schema.ts
timothyfroehlich: Declining — this is spec decision 7, made deliberately when the design was approved: cascade delete, with the resulting desync class (a live entry with no PinPoint record at all) owned by the fleet dashboard (PP-o355.7.2) and PP-o355.31 rather than by this bead. See `docs/superpowers/specs/2026-08-09-pbm-orphaned-listing-design.md`. The finding is a real gap in the feature, just one that is already routed elsewhere; the table's `machine_id NOT NULL` is a consequence of that decision, not an oversight. —Claude-PbmListUnlist

src/app/(app)/m/[initials]/(tabs)/page.tsx
Comment (timothyfroehlich) [reply src/app/(app)/m/[initials]/(tabs)/page.tsx]: Fixed in 09bae1ab — the read is now gated on `showPinballmapCard`, matching the `pbmState` gate above it. —Claude-PbmListUnlist

src/app/(app)/m/[initials]/(tabs)/page.tsx
Comment (timothyfroehlich) [reply src/app/(app)/m/[initials]/(tabs)/page.tsx]: Fixed in 09bae1ab — a linked machine whose catalog row is missing now reads "Pinball Map title #<id>" instead of claiming it is not linked, matching the abandoned notice's own fallback. —Claude-PbmListUnlist

timothyfroehlich:
## Abandoned-listing notice — re-captured at head (`877a5ca2`)

Same UI as the `382aca33` captures above; re-shot so the screenshots cover the current head after the review round that extracted `retireAbandonmentForLmx` and widened the card's status line.

```
◆ PINBALL MAP
Linked as Godzilla (Premium) · not listed.
⚠  Previous listing still live: “Godzilla (Pro)”
   ▸ Why is this here?
```

| desktop (1440×900) | mobile (390×844) |
| --- | --- |
| desktop card | mobile card |

Full pages: [desktop](https://raw.githubusercontent.com/timothyfroehlich/PinPoint/pr-screenshots/pr-1856/877a5ca2/notice-desktop.png) · [mobile](https://raw.githubusercontent.com/timothyfroehlich/PinPoint/pr-screenshots/pr-1856/877a5ca2/notice-mobile.png)

Shot as a maintainer against a hand-seeded dev DB; the manifest run cannot reach this state because it shoots `/m/TAF` as a member and `machines.pinballmap.link` is owner-only for members.

—Claude-PbmListUnlist

PR #1857, opened 2026-08-10
chore: move the memory-pressure gate out of the repo (PP-p9cy)

timothyfroehlich:
## What

**PinPoint no longer says anything about anyone's RAM.** The memory-pressure gate existed because one developer runs 4–5 parallel agent sessions on a 16 GB laptop — a fact about that machine, not about this project. For every agent on the 60 GB desktop and in CI it was noise to read past.

It keeps working, from that machine's own Claude hooks.

## Removed

| Path | Was |
| --- | --- |
| `scripts/guard/mem-precheck.sh` | the banded gate (macOS `vm_stat`/`sysctl`; already failed open elsewhere) |
| `.claude/hooks/block-heavy-under-pressure.cjs` | the PreToolUse hook that ran it, + its registration and unit test |
| `.claude/hooks/verify-guard-stack.cjs` | two entries expecting that hook |
| `scripts/diagnostics/mem-budget.sh` | the measurement harness |
| `docs/memory-budget-investigation-2026-06.md` | titled "16 GB M5 Air" |
| `AGENTS.md` | the `FORCE_MEM_PRECHECK=skip` row |
| `scripts/workflow/heavy-run.sh` | the gate call before the sem slot |

## Kept, deliberately

**`heavy-run.sh`'s GNU-parallel concurrency cap.** Capping concurrent heavy runs is true of any host — it isn't a hardware fact the way a free-RAM threshold is. Only the gate call was removed; the cap and its `preflight:unlocked` escape hatch are untouched.

**`resolve-command.cjs`.** `block-direct-merge` and `block-main-worktree-branch-switch` still use it. The machine-level hook carries its own copy — duplication across two repos is the price of the split, and the alternative is a machine hook reaching into a checkout that may not exist.

## Known behaviour change

The gate covered two paths: agent-issued Bash calls (via the hook) and `pnpm run test:integration` typed directly (via `heavy-run.sh`). **Only the agent-issued path survives.** A human typing a heavy command in a terminal is no longer gated. Nearly all heavy runs on that machine are agent-issued, so this was accepted rather than worked around.

## Verification

- `rg --hidden` for `mem-precheck` / `FORCE_MEM_PRECHECK` / `block-heavy` / `mem-budget` returns only dated `docs/superpowers/plans/` records, which are historical artifacts and left alone
- `pnpm run check` clean; full unit suite **2058 passed** (223 files)
- `verify-guard-stack.cjs` run live against the edited settings: silent, which is its healthy state
- On the machine side, an 18-case behaviour matrix covers the gate and the offload reminder — all passing

No runtime code touched; nothing to screenshot.


timothyfroehlich:
Claude review of head 0401160 — `/code-review medium` — No correctness bugs. Three stale references fixed (resolve-command header, isHeavyCommand test fixtures, memory-doc pointer); gate moved from the all/ to mac/ stow package after review flagged it was synced one machine too wide.

PR #1858, opened 2026-08-10
fix(e2e): stop three specs from leaving seeded fixtures mutated (PP-168u)

timothyfroehlich:
Closes PP-168u.

## The bug

The full E2E suite ran **ten failures on unmodified main** — five specs, each failing in firefox and Mobile Chrome and passing in chromium.

It is not a browser difference. The three browser projects share one database, chromium is scheduled first and consumes the pristine seed, and three specs were mutating seeded fixtures without restoring them. The same four files pass in isolation because nothing runs before them.

## What each spec was leaking

| Spec | Leak | How it presented |
| :--- | :--- | :--- |
| `machine-timeline.spec.ts:144` | Moved seeded **TAF-01 to EBD permanently** | `/m/TAF/i/1` no longer exists → next project times out on a kebab that page never renders |
| `machine-info.spec.ts:20` | *(victim, not a mutator)* | Its hero assertion expects TAF-01 in the known-issues peek |
| `machine-owner-picker.spec.ts:82`, `:132` | Promotes seeded `guest@test.com` to **member** | The file's `testEmails` cleanup set was **never added to** — next project finds a member where it needs a guest, so no promote dialog |
| `machine-settings.spec.ts:73` | Creates a settings set on shared TAF | Assertion reads `.first()`, but the route orders `asc(createdAt)`, so after reload `.first()` is the *previous* project's leftover |

## The fixes

- **Timeline** files its own throwaway issue and cleans up by title prefix. Moving the seeded issue back would not have worked — a returning issue gets a fresh per-machine number, so TAF-01 stays gone either way.
- **Owner picker** demotes the guest in `afterEach`, unconditionally (the confirm test can fail at any assertion *after* the promotion commits). Dead `testEmails` set removed.
- **Settings** editor journey seeds its own machine, matching the sibling `section ops persist` block in the same file.
- **machine-info is untouched** — it was collateral, and it passes once the mutator stops.

## Verification

On the Bazzite runner across chromium + firefox + Mobile Chrome:

- **Before:** 10 failed / 2 flaky / 260 passed
- **After:** **0 failed** / 270 passed / 3 skipped

The reassign journey then flaked once on chromium — arriving via the report form's redirect races the ProseMirror editor's focus, which is precisely what `openDropdownMenu` was written for. Routed through it; **33/33** over a `--repeat-each=3` stress run on chromium + firefox.

The two remaining firefox flakes (`invite-signup.spec.ts:128`, `status-overhaul.spec.ts:25`) were in the baseline too and are untouched.

## The AGENTS.md correction

#1849 added a line saying cross-browser coverage belongs to CI, "where each browser gets its own job and its own database."

That is true of the three **required** PR E2E jobs — and it is exactly why none of them could see any of this. It is **not** true of the post-merge `E2E Comprehensive Tests` job, which runs chromium + Mobile Chrome + Mobile Safari in one Playwright process against one database.

The same line also told agents to read these reds as "red runs that are not bugs." That is how ten real bugs stayed on main. Rewritten to state the CI topology accurately and to say the opposite.

**Filed PP-jxhy** for the underlying gap: the only job that can catch this class is not a required check, and it reported `skipped` or `cancelled` on every main run between 2026-08-03 and 2026-08-10.


e2e/full/machine-owner-picker.spec.ts
```diff
@@ -69,14 +80,17 @@ test.describe("Machine Owner Picker — promote-dialog journeys (PP-6oi)", () =>
...
-        userEmails: Array.from(testEmails),
       });
       testMachines.clear();
-      testEmails.clear();
     }
+    // Unconditional, and after the machines are gone (the promoted user owns
+    // the machine the confirm test creates). Only the confirm test promotes,
+    // but it can fail at any assertion AFTER the promotion has committed, so
+    // "demote when the test passed" would be exactly backwards. One UPDATE.
+    await updateUserRole(await getProfileIdByEmail(GUEST_EMAIL), "guest");
```
timothyfroehlich: **The unconditional demote can throw and fail a passing test.** Migration `0027_machine_owner_member_invariant.sql` installs `user_profiles_no_demote_owner` — a `BEFORE UPDATE OF role` trigger that raises `check_violation` ("Cannot demote user to guest while they own machines") whenever `OLD.role <> 'guest' AND NEW.role = 'guest'` and any row in `machines` still has `owner_id = NEW.id`. `updateUserRole` does `if (error) throw error`, so the raise fails the `afterEach`.

Deleting *this* test's machines first only covers this worker's machines. Concrete failure: `test-e2e-comprehensive` runs `playwright.config.full.ts` with `--project=chromium --project="Mobile Chrome" --project="Mobile Safari" --workers=3` against **one** database. `fullyParallel: false` makes the file the parallel unit, so the same file runs concurrently in all three projects. Chromium's confirm test promotes `guest@test.com` and creates `OPC…` owned by them; a moment later Mobile Chrome's *first* test finishes, deletes only its own `OPK…` machine, and runs this UPDATE — the guest still owns chromium's `OPC…`, the trigger fires, and Mobile Chrome's test is reported failed in teardown even though its body passed. Two of the three projects can hit this on any run.

Same trigger also means: if any past run leaked a guest-owned machine, this line throws on *every* subsequent run rather than silently no-op'ing.

Worth making the restore non-fatal and ownership-aware — e.g. clear `owner_id` on machines owned by the guest before the UPDATE, or swallow a `check_violation` — or sidestep the shared fixture entirely by promoting a throwaway user instead of the seeded guest.

e2e/full/machine-owner-picker.spec.ts
```diff
@@ -69,14 +80,17 @@ test.describe("Machine Owner Picker — promote-dialog journeys (PP-6oi)", () =>
   test.use({ storageState: STORAGE_STATE.admin });
 
   test.afterEach(async ({ request }) => {
-    if (testMachines.size > 0 || testEmails.size > 0) {
+    if (testMachines.size > 0) {
       await cleanupTestEntities(request, {
```
timothyfroehlich: **A cleanup failure skips the demote, reintroducing the leak this PR fixes.** `cleanupTestEntities` throws on any non-2xx from `/api/test-data/cleanup` (`cleanup.ts:47-51`), and that `await` is on the statement *before* the restore. So the one case where the restore matters most — teardown already going wrong — is the case where it never runs, and the guest stays a member for every later run in that database.

The comment on line 89 argues the restore must be unconditional because the confirm test can fail after the promotion commits; the same argument applies to the cleanup call failing. A `try { … } finally { … }` around the machine cleanup would make the restore actually unconditional.

e2e/full/machine-timeline.spec.ts
```diff
@@ -138,17 +146,47 @@ test.describe("Machine Timeline (PP-0x98)", () => {
...
+  // rendered — and machine-info.spec.ts, which asserts TAF-01 in the hero's
+  // known-issues peek, went red as collateral. Moving it back is not a repair:
+  // a returning issue gets a fresh per-machine number, so TAF-01 would stay
+  // gone. (PP-168u.)
   test.describe("admin reassign journey", () => {
     test.use({ storageState: STORAGE_STATE.admin });
 
+    test.afterEach(async ({ request }) => {
+      await cleanupTestEntities(request, {
+        issueTitlePrefix: REASSIGN_PREFIX,
```
timothyfroehlich: **Cleaning by a file-constant prefix deletes other concurrent runs' in-flight issues.** `REASSIGN_PREFIX` is the same string in every worker and every project, and the route turns it into `ilike(issues.title, 'E2E PP-0x98 Reassign%')` — an unbounded delete over *all* matching rows, not just this test's.

In `test-e2e-comprehensive` (`playwright.config.full.ts`, `--workers=3`, three projects, one database) this file runs concurrently in chromium, Mobile Chrome and Mobile Safari. Chromium finishes first and its `afterEach` deletes Mobile Safari's just-filed issue while that test is between the report redirect and `reassign-confirm`. The reassign then acts on a row that no longer exists, `waitForURL(/m/EBD/i/[0-9]+$/)` never resolves, and the test times out — a failure with no relation to the code under test.

The title already carries `Date.now()`, so capturing it in a variable and passing that exact title as `issueTitlePrefix` makes the delete match only this test's row. (`issues-reassign-machine.spec.ts` has the same shape and the same exposure — worth fixing there too rather than copying it.)

e2e/full/machine-settings.spec.ts
```diff
@@ -70,6 +82,21 @@ test.describe("Machine Settings (PP-43q3)", () => {
...
+    test.beforeEach(async () => {
+      const adminId = await getProfileIdByEmail("admin@test.com");
+      const created = await createTestMachine(adminId);
+      editorMachineId = created.id;
+      editorMachine = created.initials;
+    });
+
+    test.afterEach(async () => {
+      // ON DELETE CASCADE drops any set this journey created with the machine.
+      await deleteTestMachine(editorMachineId);
```
timothyfroehlich, line 102: **If `beforeEach` throws, this masks the real error.** Playwright still runs `afterEach` when a `beforeEach` fails, so a `createTestMachine` failure (it can throw on the promote update, the insert — `initials` is a random `TM####` that can collide — or the watcher insert) leaves `editorMachineId` `undefined` and this line issues `delete from machines where id = 'undefined'`, which Postgres rejects as an invalid uuid. The reported failure is then the teardown error, not the setup error that caused it.

A `if (!editorMachineId) return;` guard (or initialising to `""` and checking) keeps the original failure visible. The `section ops` block below has the same pattern, so this is a pre-existing shape being duplicated rather than something new — but it is cheap to guard in both.

e2e/full/machine-owner-picker.spec.ts
timothyfroehlich: Fixed in 0a05814b — by removing the restore rather than guarding it. Each test now creates its own throwaway guest (`createTestUser`; auth signup defaults to `guest` via `handle_new_user`) and deletes it in teardown, so `guest@test.com` is never touched and the demote that could hit the trigger no longer exists. Confirmed the trigger reads as you described in `drizzle/0027_machine_owner_member_invariant.sql`. —Claude

e2e/full/machine-owner-picker.spec.ts
timothyfroehlich: Fixed in 0a05814b. Moot in the new shape (no restore to skip), but the per-test user delete is in a `finally` so a failing `cleanupTestEntities` cannot leak the account. Machines still go first — `machines.owner_id` references `user_profiles` with no `ON DELETE`, so a still-owned machine blocks the delete. —Claude

e2e/full/machine-timeline.spec.ts
timothyfroehlich: Fixed in 0a05814b. Cleanup now matches the exact title the run filed, and the title carries `project-workerIndex-timestamp` so two projects filing in the same millisecond still differ. Also fixed the same shape in `issues-reassign-machine.spec.ts` — pre-existing, but leaving the twin of the bug this PR fixes would be worse than the scope creep. —Claude

e2e/full/machine-settings.spec.ts
timothyfroehlich: Fixed in 0a05814b. `editorMachineId` is `string | null`, reset in `beforeEach`, and the teardown returns early when it is null. —Claude

timothyfroehlich:
Claude review of head 7a3378d — `/code-review high` — 6 findings: unscoped reassign assertions could pass with no event emitted; throwaway auth user at an unsweepable domain; teardown masking a cleanup error with its FK symptom; 5-char initials colliding into a cross-test machine delete; unguarded section-ops teardown; 10k-value initials draw in createTestMachine. All fixed.

PR #1859, opened 2026-08-11
fix(mcp): single-source the issue enums, share the UUIDv8 key derivation (PP-brbu)

timothyfroehlich:
Hardening follow-ups on the MCP issue tools from #1851. **No correctness bugs were found in #1851** — the honest-reporting concerns there (`deduped`, `changed`, `partial`, `total` vs `count`, total-order pagination) are all handled and documented. These four changes make latent failure modes impossible rather than merely absent today.

## Where this came from

A `/code-review medium --fix` pass was launched against PR #1852 but resolved its target to `main`'s tip instead — commit `05e56e7d` (#1851), already merged. The findings are real, so rather than discard them they get their own branch. **#1852 is still unreviewed.**

## Changes

**`NamedProfile.role`: `string` → `UserRole`.** The declared return type of `findProfilesByFullName` was widening what Drizzle inferred from the column enum. A misspelled comparison in `resolveOwner` / `resolveAssignee` — `m.role !== "guests"` for `"guest"` — would have compiled into a guest filter matching nobody, letting a guest be assigned an issue. Now it fails to compile.

**Extracted `contentAddressedUuid()` and `retryWindowPart()` into `shared.ts`.** The UUIDv8 derivation — NUL join, version/variant bit stamping, 8-4-4-4-12 rendering — plus `RETRY_WINDOW_MS` and `KEY_SEPARATOR` were duplicated verbatim between `create-issue.ts` and `add-issue-comment.ts`. Two copies of hand-rolled bit stamping is two chances to get a nibble wrong, and it is load-bearing for idempotency. Both key builders now sit on the shared helpers (~50 lines of duplication removed).

**Added `ISSUE_SEVERITY_VALUES` / `ISSUE_PRIORITY_VALUES` / `ISSUE_FREQUENCY_VALUES`.** The zod schemas in `list-issues.ts` and `update-issue.ts` re-typed their literals rather than deriving from `IssueSeverity` / `IssuePriority` / `IssueFrequency`. Adding a value to those unions would leave every MCP schema silently rejecting it, with nothing failing to compile. The arrays are now the source and the types derive from them — the pattern `ISSUE_STATUS_VALUES` already used.

**Added `desc(c.id)` to the `get-issue.ts` comment window.** It ordered on `createdAt` alone under a `LIMIT`, which is not a total order. Two comments sharing a timestamp at the limit boundary sit in an order Postgres may vary, so the window can drop one and a repeat call surface a different one. This is the same reasoning `list-issues.ts:563` already applies to its own ordering, not carried over.

Plus one jsdoc fix in `get-issue.ts` — it said "when the caller doesn't ask for a count"; it's a limit, not a count.

## Deliberately not fixed

Both are behavior decisions rather than bug fixes, so they are left for a call rather than folded in here:

- **`update-issue.ts:1188`** — the `assignee` entry in `applied` reports `from`/`to` as raw UUIDs while every other field reports a human-readable value, so a caller that passed a name gets back an opaque id it cannot map to a person. Fixing means an extra name lookup and a response-payload change.
- **`get-issue.ts:329`** — when `commentsWithheld` is set, `commentCount` reports `0`, contradicting its documented meaning ("full thread length"). `commentsWithheld` saves the caller, but the count field itself lies. The branch is unreachable today because the permission matrix grants `comments.view` at every level, and fixing it means running a count query on a path that currently runs none.

## Testing

`pnpm run check` clean, `pnpm run typecheck` clean, 2072 unit tests pass — including the idempotency-key tests, which is what confirms the `contentAddressedUuid()` extraction is behavior-identical rather than merely compiling.

Closes PP-brbu.

### DIFF
--- a/src/lib/mcp/tools/add-issue-comment.ts
+++ b/src/lib/mcp/tools/add-issue-comment.ts
@@ -1,7 +1,5 @@
 import "server-only";
 
-import { createHash } from "node:crypto";
-
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { after } from "next/server";
 import { z } from "zod";
@@ -12,26 +10,15 @@ import { plainTextToDoc } from "~/lib/tiptap/types";
 import { addIssueComment } from "~/services/issues";
 
 import {
+  contentAddressedUuid,
   issueUrl,
   McpToolError,
   resolveIssue,
+  retryWindowPart,
   runTool,
   type ToolOutcome,
 } from "./shared";
 import type { McpAuthContext } from "~/lib/mcp/verify-token";
-
-/** Retries inside this window resolve to the comment already posted. */
-const RETRY_WINDOW_MS = 10 * 60 * 1000;
-
-/**
- * The joiner for the idempotency key's field parts: a NUL, written as the
- * `\u0000` ESCAPE and never as a literal byte. A literal NUL in the source makes
- * the whole file binary to the toolchain — `git diff` reports "Binary files
- * differ" so the file is unreviewable, and `rg` skips it in directory searches,
- * which silently drops it out of every future codebase sweep. `create_issue`'s
- * key uses the same escape.
- */
-const KEY_SEPARATOR = "\u0000";
 
 const addIssueCommentSchema = z.object({
   machine: z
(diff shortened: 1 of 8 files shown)


timothyfroehlich:
Claude review of head e59c772 — `/code-review medium` — No findings. Verified contentAddressedUuid extraction byte-identical to both deleted copies (part order, NUL joiner, both bit stamps), no orphaned refs to the removed constants, zod 4 readonly-tuple enums typecheck, database.ts becoming a runtime module is bundle-safe (all imports are import type), NamedProfile.role narrowing matches the column enum, desc(c.id) tiebreaker correctly placed.

PR #1860, opened 2026-08-11
fix(ui): stop painting a raw ISO instant in the relative-time slot (PP-h490)

timothyfroehlich:
## What

`RelativeTime` defaulted its pre-hydration fallback to `value.toISOString()`. Every call site that omitted `fallback` painted a 24-character instant — `2026-08-11T00:41:44.489Z` — into the timestamp slot until the shared ticker mounted. The three timeline row components and the collection timeline page all omit it.

Default to the empty string instead.

## Why this was the overflow bug

The timeline rows pin that timestamp into an `ml-auto shrink-0 whitespace-nowrap` slot sized for a label like "2 minutes ago" — 127px at a 375px viewport, with 4px to spare. The ISO instant measures **166px**, overruns the row by 35px, and is then clipped by the row's own `overflow-hidden`. That is content a user cannot read or reach.

Measured directly, instrumenting the assertion helper:

```
before hydration: text="2026-08-11T00:41:44.489Z"  w=166  right=410  vw=375   ← clipped
after  hydration: text="less than a minute ago"    w=127  right=371  vw=375   ← fits
```

Mobile Safari hydrates late enough that the state is visible to a user and to `responsive-overflow.spec.ts`; Chrome settles before the assertion runs. That is the whole reason only the non-gating Mobile Safari project went red — the bug itself is not WebKit-specific, and every browser paints the raw instant for some window.

## Why empty rather than a formatted date

The ISO default was chosen so SSR and hydration could not diverge across locales and time zones (the original comment says so). An empty string satisfies that constraint just as well and additionally cannot overflow a slot. Call sites that want a date in the pre-hydration paint keep passing one explicitly — `NotificationList.tsx:173`, `IssueList.tsx:574` and `m/[initials]/i/[issueNumber]/page.tsx:295` already pass `formatDateTime(...)`, which is stable because the string is built once on the server and handed down as a prop.

Empty-slot-then-swap was Tim's call over a short absolute date.

## Test coverage

The default fallback had **no** test — every existing case passed an explicit one. That gap is how this shipped. Added an SSR test asserting the no-`fallback` render emits no ISO instant.

## Verification

| Gate | Result |
| :--- | :--- |
| `responsive-overflow.spec.ts`, Mobile Safari | 22/22 (was 4 failed) |
| Same spec, chromium + Mobile Chrome + firefox + Mobile Safari | 60/60 |
| `pnpm run test` | 2073 passed |
| `pnpm run check` | clean |

Failing routes before the fix: `/m`, `/m/TAF`, `/m/TAF/maintenance`, `/c/owner/[member]/timeline`.

## Not in scope

PP-h490 also asks for a decision on the watch gap — Mobile Safari smoke runs only in the post-merge `E2E Comprehensive Tests` job, which is not a required check and which explicitly excludes Safari from its pass/fail decision (`ci.yml:909` counts Safari failures and echoes them to a log line). Tim's call is that Safari stays low-priority and non-gating, fixed reactively as it comes up. Recorded on the bead; no CI change here.

Closes PP-h490.

### DIFF
--- a/src/components/issues/IssueTimeline.tsx
+++ b/src/components/issues/IssueTimeline.tsx
@@ -2,6 +2,7 @@
 
 import React, { useTransition } from "react";
 import { formatDateTime } from "~/lib/dates";
+import { useRelativeNow } from "~/components/issues/RelativeTimeProvider";
 import { Avatar, AvatarFallback } from "~/components/ui/avatar";
 import { AddCommentForm } from "~/components/issues/AddCommentForm";
 import { OwnerBadge } from "~/components/issues/OwnerBadge";
@@ -202,6 +203,17 @@ function TimelineItem({
   const [isEditing, setIsEditing] = React.useState(false);
   const [isDeleteDialogOpen, setIsDeleteDialogOpen] = React.useState(false);
 
+  // Subscribing here — not just inside the `<RelativeTime>` children — is what
+  // makes the timestamp buttons' `aria-label` safe. `formatDateTime` resolves
+  // the runtime's own locale and zone, so evaluating it during SSR and again in
+  // the browser yields different strings. React does not patch a mismatched
+  // *attribute* the way it patches text, and without this subscription nothing
+  // would ever re-render the button, so a UTC-hosted render would leave the
+  // server's zone announced to screen readers forever. Rendering a static label
+  // until the first tick keeps SSR and the client's first render in agreement,
+  // then swaps in the real timestamp once we are demonstrably client-side.
+  const now = useRelativeNow();
+
   const { currentUserId, currentUserRole } = userContext;
 
   const isSystem = event.type === "system";
@@ -263,14 +275,20 @@ function TimelineItem({
                   {/* The label names the absolute instant the tooltip reveals,
                       rather than leaving the relative label as the accessible
                       name. `<RelativeTime>` renders nothing until the client
-                      ticker mounts, so without this the button is nameless on
-                      the pre-hydration paint (axe `button-name`, PP-h490) — and
-                      even hydrated, "3 minutes ago" is a poor name for a
+                      ticker mounts, so without a label the button is nameless
+                      on the pre-hydration paint (axe `button-name`, PP-h490) —
+                      and even hydrated, "3 minutes ago" is a poor name for a
                       control whose whole purpose is to disclose the exact
-                      time. */}
+                      time. The pre-tick label is a static string because
+                      `formatDateTime` is zone-dependent; see the
+                      `useRelativeNow` comment above. */}
                   <button
                     type="button"
-                    aria-label={formatDateTime(event.createdAt)}
+                    aria-label={
+                      now === null
+                        ? "Show exact time"
+                        : formatDateTime(event.createdAt)
+                    }
                     className={cn(
                       "cursor-help bg-transparent p-0 text-[11px] text-muted-foreground",
                       timestampTooltipTriggerClassName
@@ -316,7 +334,11 @@ function TimelineItem({
                     <TooltipTrigger asChild>
                       <button
                         type="button"
-                        aria-label={formatDateTime(event.createdAt)}
+                        aria-label={
+                          now === null
+                            ? "Show exact time"
+                            : formatDateTime(event.createdAt)
+                        }
                         className={cn(
                           "cursor-help bg-transparent p-0 text-xs text-muted-foreground",
                           timestampTooltipTriggerClassName
@@ -332,9 +354,15 @@ function TimelineItem({
                   {isEdited && !isIssue && (
                     <Tooltip>
                       <TooltipTrigger asChild>
+                        {/* No `aria-label` here: unlike the two triggers above,
+                            this one carries visible text ("• edited"), so it is
+                            never nameless even while `<RelativeTime>` is empty.
+                            Leaving the content to name it also keeps the
+                            accessible name a superset of the visible label
+                            (WCAG 2.5.3), and avoids a zone-dependent
+                            attribute. */}
                         <button
                           type="button"
-                          aria-label={`Edited ${formatDateTime(event.updatedAt)}`}
                           className={cn(
                             "cursor-help bg-transparent p-0 text-xs text-muted-foreground",
                             timestampTooltipTriggerClassName
--- a/src/components/issues/IssueTimeline.tsx
+++ b/src/components/issues/IssueTimeline.tsx
@@ -260,8 +260,17 @@ function TimelineItem({
               )}
               <Tooltip>
                 <TooltipTrigger asChild>
+                  {/* The label names the absolute instant the tooltip reveals,
+                      rather than leaving the relative label as the accessible
+                      name. `<RelativeTime>` renders nothing until the client
+                      ticker mounts, so without this the button is nameless on
+                      the pre-hydration paint (axe `button-name`, PP-h490) — and
+                      even hydrated, "3 minutes ago" is a poor name for a
+                      control whose whole purpose is to disclose the exact
+                      time. */}
                   <button
                     type="button"
+                    aria-label={formatDateTime(event.createdAt)}
                     className={cn(
                       "cursor-help bg-transparent p-0 text-[11px] text-muted-foreground",
                       timestampTooltipTriggerClassName
@@ -307,6 +316,7 @@ function TimelineItem({
                     <TooltipTrigger asChild>
                       <button
                         type="button"
+                        aria-label={formatDateTime(event.createdAt)}
                         className={cn(
                           "cursor-help bg-transparent p-0 text-xs text-muted-foreground",
                           timestampTooltipTriggerClassName
@@ -324,6 +334,7 @@ function TimelineItem({
                       <TooltipTrigger asChild>
                         <button
                           type="button"
+                          aria-label={`Edited ${formatDateTime(event.updatedAt)}`}
                           className={cn(
                             "cursor-help bg-transparent p-0 text-xs text-muted-foreground",
                             timestampTooltipTriggerClassName
--- a/src/components/issues/RelativeTime.tsx
+++ b/src/components/issues/RelativeTime.tsx
@@ -17,10 +17,17 @@ interface RelativeTimeProps {
    * so the raw instant overflowed the row and was clipped until hydration
    * (PP-h490). Empty is both zone-independent and unable to overflow.
    *
-   * Pass a server-computed absolute label where the pre-hydration paint should
-   * still carry a date — `NotificationList`, `IssueList` and the issue detail
-   * page all pass `formatDateTime(...)`, which is stable because the string is
-   * built once on the server and handed down as a prop.
+   * Pass an absolute label where the pre-hydration paint should still carry a
+   * date. The pattern to copy is `IssueUpdatedTimestamp`, whose `fallback` is
+   * computed by the issue-detail *server* page and handed down as a prop — the
+   * string is then built once, in one zone, and cannot diverge.
+   *
+   * `NotificationList` and `IssueList` also pass `formatDateTime(...)`, but
+   * both are `"use client"` and call it inline, so the value is computed twice
+   * in two possibly-different zones. That is tolerable *here* only because a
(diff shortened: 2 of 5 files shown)


timothyfroehlich:
## UI Screenshots — head `da689142`

### Issues List

| --- | --- |
| desktop | mobile |

| --- | --- |
| desktop | mobile |

### Report an Issue

| --- | --- |
| desktop | mobile |

### Dashboard

| --- | --- |
| desktop | mobile |

| --- | --- |
| desktop | mobile |

### Collections

| --- | --- |
| desktop | mobile |

| --- | --- |
| desktop | mobile |

### New Machine

| --- | --- |
| desktop | mobile |

| --- | --- |
| desktop | mobile |

timothyfroehlich:
## Why the screenshots above look unchanged

They should. This bug only exists **before hydration**, so every settled-state
screenshot is identical either side of the fix.

The evidence is the measurement, taken by instrumenting the real
`assertNoHorizontalOverflow` helper so it reports at exactly the moment it
asserts, on Mobile Safari at 375px:

```
before hydration: text="2026-08-11T00:41:44.489Z"  w=166  right=410  vw=375   <- clipped
after  hydration: text="less than a minute ago"    w=127  right=371  vw=375   <- fits
```

`RelativeTime` defaulted its pre-hydration `fallback` to `value.toISOString()`
— 24 characters, 166px — painted into a `shrink-0` slot sized for a relative
label like "2 minutes ago" (127px). The row is `overflow-hidden`, so it clipped
until the client ticker mounted.

**Not WebKit-specific.** Every browser paints that instant for some window;
WebKit just hydrates late enough that a user, and the assertion, can see it.
The offending rows sat at `top=840/914/987` — below the 812px fold — which is
why the original failure screenshot appeared to show a correctly-fitting
timestamp.

## Two commits, two defects

1. **`35080f3`** — default the fallback to `""`. A default derived from `value`
   is wrong twice over: locale/zone formatting diverges between SSR and
   hydration, and the ISO form cannot fit the slot. Call sites that want a
   pre-hydration date pass a server-computed `formatDateTime(...)` string, and
   three already do.

2. **`69f983d`** — the empty default exposed that two `IssueTimeline` tooltip
   triggers had `<RelativeTime>` as their *entire* content, so they became
   nameless buttons (`axe` `button-name`, critical). Fixed with explicit
   `aria-label`s. This was a latent weakness either way: "3 minutes ago" is a
   poor accessible name for a control whose purpose is to disclose the exact
   time.

Every pre-existing `RelativeTime` test passed an explicit `fallback` — the
default was uncovered, which is how this shipped. The regression test added
here covers it.

## Verification

- CI green on head, including the a11y specs that caught defect 2.
- Mobile Safari is **not** run by any PR job, so it was verified locally:
  `responsive-overflow.spec.ts` 22/22 and the full local Mobile Safari subset
  32/32.

—Claude

src/components/issues/IssueTimeline.tsx
```diff
@@ -260,8 +260,17 @@ function TimelineItem({
...
+                      rather than leaving the relative label as the accessible
+                      name. `<RelativeTime>` renders nothing until the client
+                      ticker mounts, so without this the button is nameless on
+                      the pre-hydration paint (axe `button-name`, PP-h490) — and
+                      even hydrated, "3 minutes ago" is a poor name for a
+                      control whose whole purpose is to disclose the exact
+                      time. */}
                   <button
                     type="button"
+                    aria-label={formatDateTime(event.createdAt)}
```
timothyfroehlich: **Hydration mismatch: `formatDateTime` here is time-zone dependent and runs on both sides of hydration.**

`formatDateTime` goes through `Intl.DateTimeFormat(undefined, …)` (`src/lib/dates.ts:51`), which resolves to the *runtime's* locale and time zone. `IssueTimeline` is `"use client"`, but Next still SSRs it (rendered from the server page `src/app/(app)/m/[initials]/i/[issueNumber]/page.tsx`), so this `aria-label` is baked into the server HTML using the server's zone — UTC on Vercel — and then recomputed during hydration using the viewer's zone.

React 19's own message for this case is explicit (react-dom-client, `emitPendingHydrationWarnings`): *"A tree hydrated but some attributes of the server rendered HTML didn't match the client properties. **This won't be patched up.** … Date formatting in a user's locale which doesn't match the server."*

Concrete failure: an Austin member (UTC−5) opens an issue commented on at 15:45 local. The server emits `aria-label="Aug 11, 2026, 8:45 PM"`; the client computes `"Aug 11, 2026, 3:45 PM"`. React does not patch the attribute, and nothing re-renders `TimelineItem` afterwards (the 60s ticker only re-renders `RelativeTime`, the child), so a screen-reader user hears **8:45 PM** for the rest of the page's life while the tooltip — rendered client-side on open — says **3:45 PM**. Plus a hydration error in dev/Sentry. It does not reproduce locally because the dev server and the browser share a time zone.

The codebase already has the fix for exactly this: `IssueUpdatedTimestamp` takes the formatted string as a prop, computed once in the server page (`page.tsx:295`), and uses that same string for both `aria-label` and `fallback`. Same treatment applies to lines 319 and 337.

src/components/issues/RelativeTime.tsx
```diff
@@ -6,8 +6,22 @@ import { useRelativeNow } from "./RelativeTimeProvider";
...
+   * SSR and hydration, and the ISO instant this once defaulted to is 24
+   * characters — far wider than any relative label it stands in for. The
+   * timeline rows pin this into a `shrink-0` slot sized for "2 minutes ago",
+   * so the raw instant overflowed the row and was clipped until hydration
+   * (PP-h490). Empty is both zone-independent and unable to overflow.
+   *
+   * Pass a server-computed absolute label where the pre-hydration paint should
+   * still carry a date — `NotificationList`, `IssueList` and the issue detail
+   * page all pass `formatDateTime(...)`, which is stable because the string is
+   * built once on the server and handed down as a prop.
```
timothyfroehlich: **This paragraph's rationale is wrong for two of the three call sites it cites, and will steer future ones into the bug it warns about.**

> `NotificationList`, `IssueList` and the issue detail page all pass `formatDateTime(...)`, which is stable because the string is built once on the server and handed down as a prop.

Only the issue detail page does that: `src/app/(app)/m/[initials]/i/[issueNumber]/page.tsx:295` computes it in a Server Component and passes it into `IssueUpdatedTimestamp`.

- `src/components/notifications/NotificationList.tsx:173` — `"use client"`, calls `formatDateTime(notification.createdAt)` inline.
- `src/components/issues/IssueList.tsx:574` — `"use client"`, calls `formatDateTime(issue.updatedAt)` inline.

Both are SSR'd, so the string is built **twice** — once in the server's zone, once in the viewer's — and both are already producing hydration text mismatches in production for any non-UTC viewer (invisible in dev, where the two zones agree). Worth either fixing them or, at minimum, rewording so the doc names the actual invariant: *the string must be computed in a Server Component and passed as a prop* — not merely "pass `formatDateTime(...)`".

src/components/issues/RelativeTime.tsx
```diff
@@ -19,11 +33,7 @@ export function RelativeTime({
   // After mount the shared ticker emits a number every 60s, causing a re-render.
   const now = useRelativeNow();
 
-  const resolvedFallback = (() => {
-    if (fallback !== undefined) return fallback;
-    const date = typeof value === "string" ? new Date(value) : value;
-    return Number.isNaN(date.getTime()) ? "" : date.toISOString();
-  })();
+  const resolvedFallback = fallback ?? "";
```
timothyfroehlich, line 43: **Four existing call sites pass no `fallback`, so this flips them from "wrong-looking timestamp" to "no timestamp", and one of them now paints a dangling separator.**

The doc above says to pass a server-computed absolute label wherever the pre-hydration paint should still carry a date, but no call site was updated in this PR. The sites that omit `fallback` are:

- `src/components/collections/CollectionOverviewTable.tsx:385`
- `src/components/machines/timeline/MachineTimelineIssueRow.tsx:106`
- `src/components/machines/timeline/MachineTimelineSystemRow.tsx:77`
- `src/components/machines/timeline/MachineTimelineCommentRow.tsx:109`
- plus the three in `IssueTimeline.tsx`

Concrete: the collection overview's "Last activity" cell renders `<RelativeTime …>{" — "}{getTagLabel(…)}`, so the SSR paint is a dangling `" — Comment"` with nothing before the em dash until hydration. The machine timeline's right-pinned slot paints empty. Empty is the right *default* (it matches `CompactAge`), but these five sites are the ones the doc says should opt in to `formatDateTime(...)` — passed down from their server pages so it doesn't reintroduce the hydration problem.

src/components/issues/IssueTimeline.tsx
timothyfroehlich: Confirmed and fixed in `da68914`. Verified the whole chain before fixing: `formatDateTime` is `Intl.DateTimeFormat(undefined, …)` (locale *and* zone from the runtime), `IssueTimeline` is imported directly by the issue-detail server page with no `ssr:false`, and nothing re-renders `TimelineItem` — the ticker subscription lives in the `RelativeTime` child.

`TimelineItem` now subscribes to the ticker itself and renders a static `"Show exact time"` until the first tick, so SSR and the client's first render agree and the real timestamp only appears once we are demonstrably client-side. I did not thread a server-computed string down the way `IssueUpdatedTimestamp` does: the timeline events are built inside the client component from the `issue` prop, so per-event labels would mean the server page passing a parallel map — more surgery than this PR should carry.

The `• edited` trigger lost its `aria-label` entirely rather than gaining a gated one: it has visible text, so it was never nameless, and letting the content name it keeps the accessible name a superset of the visible label.

Regression test added in `issue-detail-permissions.test.tsx` — renders a system event, an edited comment and the issue row so all three trigger branches are covered, then asserts no `aria-label` carries a formatted clock time. I checked it fails against the reintroduced bug, not just that it passes now.

—Claude

src/components/issues/RelativeTime.tsx
timothyfroehlich: Correct — verified both: `NotificationList.tsx:173` and `IssueList.tsx:574` are each `"use client"` and call `formatDateTime` inline. Only the issue-detail page computes it server-side. Comment rewritten in `da68914` to point at `IssueUpdatedTimestamp` as the pattern to copy.

One refinement to the "same latent mismatch" framing: for those two it is milder than for an attribute, and the corrected comment now says why. A `fallback` is *text*, and the ticker replaces it on mount regardless — so a zone divergence there is transient and self-healing. The same inline call feeding an `aria-label` is the real bug, because React never patches the attribute and nothing re-renders it. That distinction is the thing the next call site actually needs to know, so it is stated rather than left as "both are wrong".

—Claude

src/components/issues/RelativeTime.tsx
timothyfroehlich: The `CollectionOverviewTable` dangling separator is real and fixed in `da68914`. Extracted a `LastActivityCell` that subscribes to the ticker, so the em dash is gated on the same tick as the time it separates. Kept the subscription in the cell rather than the row map so the 60s re-render stays scoped exactly where it already was.

I swept the other call sites rather than assuming that was the only one. Non-test sites with no `fallback`: `MachineTimelineIssueRow:106`, `MachineTimelineSystemRow:77`, `MachineTimelineCommentRow:109`, the three `IssueTimeline` triggers, and this one. The three machine-timeline rows all render the timestamp alone as `rightMeta` with no adjacent separator or punctuation, so an empty slot there is exactly the intended state — that was Tim's explicit choice over a short or full date. The `IssueTimeline` triggers are covered by the `aria-label` fix. So `CollectionOverviewTable` was the only site with an artifact.

On not opting any site into an absolute label: that stays deliberate. The empty default matches `CompactAge`, which already carries the same convention and rationale, so this aligns `RelativeTime` with the existing house pattern rather than inventing one.

—Claude

timothyfroehlich:
Claude review of head da68914 — `/code-review high` — 3 findings, all fixed: zone-dependent aria-label evaluated across SSR/hydration (high), doc comment overstating which call sites pass a server-computed fallback (medium), dangling em-dash separator in the collections activity column (low)

PR #1861, opened 2026-08-11
fix(ci): make the comprehensive E2E gate work, and fix what it then catches (PP-jxhy)

timothyfroehlich:
Three commits, separable if you'd rather review them apart.

## 1. The gate never worked

The post-merge `E2E Comprehensive Tests` job is the only one that can see cross-spec seed contention — three browser projects, **one** Playwright process, **one** database. Its Playwright steps carry `continue-on-error: true` so non-gating Mobile Safari red doesn't fail the job, which makes the evaluate step the only thing that *can* fail it. It was broken three ways, all silent.

**The jq never reached a browser test.** `.suites[].specs[]` assumes specs hang off the file suite; Playwright nests them under the `describe` suite. A/B against run 31430839981's real 71 MB report:

| expression | specs found |
| :--- | ---: |
| `.suites[].specs[]` (old) | **3** |
| recursive (new) | **210** |

The three are `auth.setup.ts`'s — the only specs with no `describe` block. Inject a failure into a real nested chromium spec and the old expression still finds 0. Every "Gating browsers green" this job ever printed covered zero browser tests.

**A dead run inherited the previous run's verdict.** Both steps wrote `playwright-report/results.json`, and the JSON reporter writes once, at the end. That run's artifact proves it: the uploaded `results.json` has `configFile: playwright.config.smoke.ts`. Each step now writes its own file and clears it first; missing is a hard failure, as is a zero-spec report.

**The job timeout cancelled the verdict.** 45 min against a measured setup 2.5m + smoke 13m + full ~24m. A job-level timeout cancels outright, so the run reported `cancelled` — which reads as "superseded by a newer push". Timeouts moved onto the steps, where `continue-on-error` makes them survivable; job timeout is now a 75-min backstop.

The duplicated block became `scripts/workflow/evaluate-e2e-results.sh` with 14 tests driving the real bash.

## 2. What the working gate then caught

`full × Mobile Chrome` has **never** run in a gating job — the three required E2E jobs are smoke×chromium, smoke×Mobile Chrome, full×chromium. It fails ~1 spec per run, with a rotating victim.

One mechanism, not four bugs: **Playwright's actionability checks (visible, stable, enabled, receives-events) say nothing about whether React has attached a handler.** A click in the paint-before-hydrate window is silently discarded — no error, nothing pending — and the spec fails wherever it next asserts the consequence: a Radix portal that never opened (`status-overhaul:25`), a Sheet that never appeared (`form-resets:190`, `rich-text:105`), a navigation that never happened (`technician-role:89`).

Evidence: serial `--workers=1` passes 92/92; `--workers=3` fails ~1 per run; every victim passes in isolation; the victim rotates. Only Mobile Chrome fails because the webServer runs `pnpm run dev` and mobile renders trees chromium never compiles.

Fix is one wait, not fifty guards: a `HydrationBeacon` in `ClientProviders` sets `data-hydrated`, and a project-local `test` (`e2e/support/fixtures.ts`) makes `page.goto` wait for it. All 53 spec files import from there.

**Verified: 9 consecutive Mobile Chrome full runs green** (from ~1 failure/run), chromium unchanged at 94/94.

Also fixes a race this exposed in `report-stale-machine.spec.ts` — it seeded the draft after navigating, but the draft store's persist effect rewrites that key with empty state once it restores, so the write only survived if it landed after that effect. Now seeded via `addInitScript` pre-navigation.

## 3. A correction

An earlier comment claimed "20s did not rescue `status-overhaul:25`". It was never 20s — **the crabbox runner leaves `$CI` unset**, so the constant took its local 5s branch and `playwright.config.ts` applied `actionTimeout` 5s / `expect` 10s rather than CI's 30s. The conclusion held but was reached for the wrong reason. Commit 3 corrects the comment.

The upside: every verification above ran under **tighter** budgets than CI applies.

## Known gap — not fixed here

`PP-2b3r` (P1, filed): the same dropped-click mechanism on **client-side** navigations. `collections.spec.ts:34` clicks a modal trigger, `:35` fills its field and dies with `locator.fill: Timeout 5000ms` — the modal never opened. The beacon hangs off `page.goto`, and these specs click through to the page, so there's no new document to wait on. It helped but didn't close it: branch smoke 2 of 4 runs clean vs `main` 0 of 3.

**So the comprehensive job's smoke step will still be red.** That is pre-existing on `main`, not caused here, but it means this PR does not make the job green on its own.

## Verification

- `pnpm run check`, `check:python` (544), `typecheck:e2e` — green
- Gate script vs the real 71 MB CI report: 210 specs, 2.3s; with an injected nested failure, correctly red while the old expression stays blind
- Mobile Chrome full ×9 green; chromium full ×2 green (94/94); `report-stale-machine` isolated green


timothyfroehlich:
Claude review of head 83ba123 — `/code-review xhigh` — 14 findings addressed; head since changed only by REMOVING the hydration work to #1873 — every remaining line is byte-identical to what was reviewed

PR #1862, opened 2026-08-11
fix(perms): allow the review-marker script so the attestation step is deterministic (PP-yx97)

timothyfroehlich:
## What

Adds the repo's first `permissions.allow` entry to `.claude/settings.json`, scoped to one script:

```json
"allow": ["Bash(bash scripts/workflow/mark-claude-review.sh *)"]
```

## Why

`mark-claude-review.sh` is the only thing that satisfies `merge-pr.sh`'s `reviewed` gate. It was **intermittently** denied by the auto-mode classifier:

| Date | PR | Outcome |
| :--- | :--- | :--- |
| 2026-08-03 | #1815 | denied — `Blocked by classifier` |
| 2026-08-09 | #1832, #1828, #1829 | succeeded |
| 2026-08-10 | #1848 | succeeded |

Contextual, not a standing rule — the worst shape for a required step, because it fails only sometimes and leaves the PR at `unreviewed` with no path forward. A background subagent has no human to hand the command to at all.

Permission rules evaluate **deny → ask → allow**, so an explicit allow resolves the call before the classifier is consulted. No existing deny or ask rule matches this command.

## The tradeoff, stated plainly

The allow rule removes the harness's opinion about whether the attestation was *earned*. One of the datapoints suggests the classifier may have been reasoning about exactly that — the 2026-08-10 success came in a session where Tim had just run `/code-review` and the review was in the agent's context.

So this deliberately trades a possibly-correct safeguard for determinism, and the docs say so rather than pretending the rule fixes the honesty problem. **Merging stays human-only (PP-wi85)**, so a marker posted in bad faith misleads Tim rather than merging anything by itself.

Tim approved this permission explicitly on 2026-08-11.

## Docs

`pinpoint-pr-workflow` Phase 3.4 gains a subsection covering why the rule exists, what it does not do, and two limits: chained subcommands don't inherit the allow, and an unbalanced quote in the summary makes the command unresolvable to `block-direct-merge.cjs` (which then fails closed).

## Verification

- `pnpm run check` green.
- Guard-stack + merge-guard unit tests: 112 passed.
- `verify-guard-stack.cjs` canary reports healthy — `deny` (11) and `ask` (41) intact.
- Probed `block-direct-merge.cjs` directly: it allows the marker command even when the free-text summary contains `merge-pr.sh` or `pr merge`, because a statically resolvable command returns before the raw indicator scan.

Not a CORE-SEC-010 surface — that rule's ban on `allow` governs prod-mutating Supabase tools.

Closes PP-yx97.

### DIFF
--- a/.agents/skills/pinpoint-pr-workflow/SKILL.md
+++ b/.agents/skills/pinpoint-pr-workflow/SKILL.md
@@ -130,6 +130,23 @@ This is a narrow exception and it is self-policing. "It's only a small change" i
 
 **The marker attests that a review actually happened.** Posting it otherwise is a false attestation, not a shortcut — the same honesty model as `merge-pr.sh --force`.
 
+#### Why the marker command carries a permission allow rule
+
+`.claude/settings.json` has one `permissions.allow` entry, and it is this script:
+
+```json
+"allow": ["Bash(bash scripts/workflow/mark-claude-review.sh *)"]
+```
+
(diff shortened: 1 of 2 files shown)


timothyfroehlich:
Claude review of head e1bc206 — `/code-review medium` — One medium finding: the note falsely claimed normalize-workspace-paths.cjs rewrites absolute paths (regex is hardcoded to /home/froeht, never fires on this Mac). Corrected in e1bc206c; hook staleness filed as PP-yk8b.

PR #1823, opened 2026-08-04
docs: delete LOGGING.md and DEVELOPMENT.md, both fully cached (PP-22e4)

timothyfroehlich:
Sweep task 7 of PP-22e4. **−464 lines**, two whole docs.

## `docs/LOGGING.md` (233 lines) — test 2

`src/lib/logger.ts` is 139 lines and comments its own decisions. That includes the one the plan flagged as a possible survivor — the read-only-platform stdout fallback — which the code states twice (`logger.ts:22-24`, `:33-34`, both naming Vercel serverless explicitly). `LOG_LEVEL` and `PINPOINT_LOG_DIR` are already in `docs/ENV_VARS.md:187` with their defaults.

Two sections were worse than redundant:

- **The `/parse-logs` section documented a command that no longer exists.** `.claude/commands/` is gone; `rg --hidden parse-logs` returned exactly one file — LOGGING.md, describing it.
- **The logging examples taught the wrong context key.** They show `error:` / `stack:`, while `logger.ts:77-78` records that every call site was standardised on `err` because `pino.stdSerializers.err` fires for that key alone. The doc was teaching the shape the code moved away from.

## `docs/DEVELOPMENT.md` (231 lines) — the second front door

Its own opening line conceded the problem: *"For full project rules and constraints, always start with `AGENTS.md`. This file is a shorter, hands-on reference."* Section by section:

| DEVELOPMENT.md section | Already stated by |
| --- | --- |
| Quickstart, install, dev server | `README.md` § Local Setup |
| Worktree ports / post-checkout | `AGENTS.md` §4 |
| check / test / preflight | `AGENTS.md` §5 key-commands table + README |
| Testing overview | `pinpoint-testing` skill (+ README pointer) |
| `shadcn@latest add` | `pinpoint-ui/references/key-files.md:10` |
| db:generate / migrate / reset | `README.md` § Database Workflow |
| Vercel automated migrations, `mark-migration-applied` | `AGENTS.md` §7 |
| TS7 dual-install + oxlint mirror | `AGENTS.md` §5 — down to the same 14.9s/3.2 GB vs 3.8s/1.2 GB measurements |
| Troubleshooting (connection, type errors, PGlite isolation) | Generic, or CORE-TEST-001 |

Where it diverged from those, it was **stale**: it said preview deploys run migrations (§5 says they skip them) and that `preflight` runs smoke (it does not).

## What moved instead of being deleted

Test 2 in the other direction — the thing the last two reviews caught:

[...]

PR #1866, opened 2026-08-11
fix(security): widen the logger redact list beyond one nesting level (PP-tg9y)

timothyfroehlich:
Closes PP-tg9y.

## The gap

`src/lib/logger.ts` redacted with `["reporterEmail", "*.reporterEmail", "email", "*.email"]`. Pino's `*` matches **exactly one** level, so an email nested two deep —

```ts
log.info({ ctx: { user: { email } } }, "…");
```

— was written raw to `logs/<session>/app.log` and, on Vercel, to production stdout. `.claude/rules/always.md` scopes CORE-SEC-007 to display surfaces and names this redact list as the backstop for logs, so it is the only thing standing between a nested email and prod stdout.

Surfaced by the `/code-review` on PR #1823, which was docs-only — the gap is pre-existing.

## What changed

The path list is now generated from two explicit inputs rather than hand-written:

- **Keys** (5): `email`, `reporterEmail`, `userEmail`, `contactEmail`, `submittedEmail`. Only the first two reach `log.*` today. The other three are defensive: `submittedEmail` is the login-form value in `src/app/(auth)/actions.ts`, but it is returned to the *client* in a `Result` error payload and is never logged; `userEmail` and `contactEmail` are the two the bead's acceptance criteria named and do not exist at all. `to` is deliberately excluded despite being a live email log key in `src/lib/email/client.ts` — the name is too generic (timeline events log `{ from, to }` for status, title and presence changes), and that call site already masks with `maskEmail`.
- **Wildcard depth**: 0–3, i.e. `email`, `*.email`, `*.*.email`, `*.*.*.email`. 20 paths total.

`baseLoggerOptions` is hoisted out of `createLogger()` and exported, so the new test drives the **real** options object through pino rather than a copy of the path list.

## Depth chosen, and why

**Depth 3** — a key is caught from the top level of the payload down to four levels in (`{ a: { b: { c: { email } } } }`). An array index consumes a level the same as an object key, so `{ members: [{ email }] }` is covered too.

Not "as deep as possible", for two reasons:

1. **Pino has no recursive wildcard.** Unbounded is not on the menu; the only lever is more paths.
2. **Wildcard paths are the expensive part of redaction**, at roughly linear cost in keys × levels. Benchmarked on pino 10.3.1 / node 24 against a representative payload, ns per `log.info`:

   | config | paths | ns/log |
   | --- | ---: | ---: |
   | no redaction | 0 | 785 |
   | old list | 4 | 2,375 |
   | 5 keys, depth 0 | 5 | 1,284 |
   | 5 keys, depth ≤1 | 10 | 7,917 |
   | 5 keys, depth ≤2 | 15 | 12,268 |
   | **5 keys, depth ≤3 (this PR)** | **20** | **17,853** |

   ~18µs against a serverless invocation measured in tens of milliseconds is noise, but it is why this is an explicit bounded set rather than an open-ended one.

Depth 3 is two levels of headroom over the deepest shape any current call site produces: `reportError` does `log.error({ ...context, err })`, so a domain object passed as context puts its keys one level down.

## Test

New `src/lib/logger.test.ts` (unit, class G — pure logic; node environment via pragma since the logger is server-side). 12 cases, the load-bearing one being the exact PP-tg9y regression: a two-level-nested `email` must not appear in the output line.

Also covers depths 0–3, an array index, each of the five key names, an email-*named property* on a serialized `err`, and that non-email siblings survive. The helper asserts the log line was actually produced, so a silenced logger cannot make `not.toContain(email)` pass vacuously.

Verified non-vacuous by temporarily setting the depth back to 1: 9 of the 12 fail.

## One doc correction

`.agents/skills/pinpoint-ui/SKILL.md` § Logging quoted the old four-path list verbatim and named `contactEmail`, `userEmail`, and two-deep nesting as things logged raw. All three are now covered, so that bullet became a false claim about this exact control. Restated against `EMAIL_LOG_KEYS × depth 0–3`, keeping the honest caveat that pino has no recursive wildcard and call-site `maskEmail` is still the primary defence (AGENTS.md §8 — fold divergence into canonical text).

## What this does not fix

A key list matches key *names*. It cannot see an address embedded in a value — and the highest-probability such vector is still open: a mail-provider send failure carries the recipient in its `message`, `stack`, `rejected` and `envelope.to`, and reaches the log raw through `reportError`'s `err` serializer (`src/lib/email/transport.ts:82,128` → `src/lib/observability/report-error.ts:44`). Reproduced during the `/code-review` on this PR. **`ResendTransport` is the production path**, so this is live, not dev-only. Closing it needs a scoped serializer, not more key names — filed as **PP-45qx**, and `EMAIL_LOG_KEYS`' doc comment points there.

## Scope

Deliberately not touched: CORE-SEC-007's wording in `docs/NON_NEGOTIABLES.md`, and every call site that logs an email (call-site masking via `maskEmail` remains the house precedent — this list is the backstop for the sites that forget).

## Gates

`pnpm run check` clean; `pnpm run test` 2100/2100 green (AGENTS.md §5 tier 2).

## Review

`/code-review medium` (2026-08-13) returned three non-blocking findings and verified the implementation empirically against pino 10.3.1 — depth 0–3 redaction, array traversal, no mutation of the caller's object, getters and frozen objects surviving. Two findings are addressed above (the `submittedEmail` claim, and naming the serializer gap). The third proposed adding `to` to the key list; declined, with the reasoning recorded in the code — see the Keys bullet.


timothyfroehlich:
Claude review of head bbf6fa3 — `/code-review medium` — 3 non-blocking findings: 2 fixed (submittedEmail coverage claim corrected; serialized-error test renamed to what it proves), 1 declined with reasoning in-code (adding 'to' — too generic, timeline events log from/to). Follow-up err-serializer leak filed as PP-45qx.

PR #1868, opened 2026-08-12
fix(routing): resolve /m/<initials> case-insensitively (PP-ofvl)

timothyfroehlich:
## What

`https://pinpoint.austinpinballcollective.org/m/rush` 404s while `/m/RUSH` works. Machine initials are stored uppercase (the `machines` table carries a `^[A-Z0-9]{2,6}$` check constraint) and every `/m/[initials]` route looks them up with an exact `eq()`, so any hand-typed or lowercased link dies.

Middleware now redirects the lowercase form to the canonical uppercase one (`/m/afm` → 307 → `/m/AFM`) rather than making the lookups case-insensitive, so there stays exactly one URL per machine and every `href` built from `machine.initials` keeps matching.

- Only the initials segment is rewritten — `/m/afm/i/12`, `/m/afm/maintenance` and the query string survive.
- `/m/new` (the create route) is excluded by name; uppercasing it would send it to a machine lookup for "NEW".
- Segments outside the initials shape (`^[A-Za-z0-9]{2,6}$`) are left alone.
- `/report?machine=<initials>` gets the same treatment inline, since that param is hand-typeable for the same reasons.

## Testing

- `src/lib/machines/canonical-path.test.ts` — the path rule (casing, sub-paths, reserved `/m/new`, non-machine paths).
- `src/test/unit/middleware-machine-canonical.test.ts` — the middleware wiring: 307, `Location`, query preserved, session refresh skipped.
- `src/app/(app)/report/default-machine.test.ts` — case-insensitive initials match.
- `pnpm run check` green, `pnpm run test` green (2097 tests).

Manually verified against a local dev server with a seeded `AFM` machine:

| request | result |
| --- | --- |
| `/m/afm` | 307 → `/m/AFM`, 200 after following |
| `/m/afm/i/1?from=qr` | 307 → `/m/AFM/i/1?from=qr` |
| `/m/AFM` | 200, no redirect |
| `/m/new` | untouched (auth gate, not uppercased) |

That verification needed a workaround, which is a separate bug: **`next dev` ignores the root `middleware.ts`** — no CSP or `x-nonce` header on any response, no redirects, reproduced against unmodified `HEAD`. Copying the same file to `src/middleware.ts` makes it run immediately. Production is unaffected (prod CSP carries a per-request nonce, which can only come from middleware). Filed as **PP-tld9** with the evidence and a proposed fix; it is not touched here.

No E2E spec: until PP-tld9 lands, the dev server Playwright starts has no middleware either, so the spec would fail in CI for environment reasons rather than code ones.

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -27,7 +27,11 @@ export async function middleware(request: NextRequest): Promise<NextResponse> {
   if (canonicalPath) {
     const url = request.nextUrl.clone();
     url.pathname = canonicalPath;
-    return NextResponse.redirect(url);
+    // 308, not the 307 default: the rule is structural, not situational.
+    // Initials are uppercase by check constraint and permanent once assigned,
+    // so the lowercase form will never become the right URL. A permanent
+    // redirect lets browsers and crawlers stop paying for the hop.
+    return NextResponse.redirect(url, 308);
   }
 
   // 1. Run Supabase middleware first to handle session
--- a/middleware.ts
+++ b/middleware.ts
@@ -1,5 +1,6 @@
-import { type NextRequest, type NextResponse } from "next/server";
+import { NextResponse, type NextRequest } from "next/server";
 import { updateSession } from "~/lib/supabase/middleware";
+import { canonicalMachinePath } from "~/lib/machines/canonical-path";
 
 /**
  * Next.js middleware for Supabase SSR authentication and security headers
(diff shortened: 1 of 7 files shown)


timothyfroehlich:
Claude review of head 3282c1f — `/code-review medium` — 5 findings: repeated-query-param crash on /report (fixed), reserved-segment drift now guarded by a route-directory test, /M/afm prefix casing (fixed), 307->308 (fixed), CSP headers on the bodyless redirect (declined)

PR #1870, opened 2026-08-12
feat(ci): fail when an e2e spec is matched by zero suite config (PP-2g7m)

timothyfroehlich:
Closes PP-2g7m.

## The problem

A spec file that no Playwright suite config collects still exists, still type-checks, and still lints. Every signal a developer sees stays green while the file runs nowhere. PP-stut found `e2e/profiles/profile-edit.spec.ts` had been dead that way for about five weeks.

## The gate

`scripts/check_e2e_suite_coverage.py` enumerates every `e2e/**/*.spec.ts` on disk and asserts each one is collected by `playwright.config.full.ts` or `playwright.config.smoke.ts`. It runs as `check:e2e-suites` inside `pnpm run check` and as a step in the required **Fast Linters** CI job.

Today's coverage, as the gate reads it:

```
playwright.config.full.ts:  testDir=e2e        testMatch=**/full/**/*.spec.ts
playwright.config.smoke.ts: testDir=e2e/smoke  testMatch=(default)
```

## Two design calls worth reviewing

**Glob, not `playwright test --list`.** `--list` would remove any need to reimplement `testMatch` semantics, but it costs a Node process plus a TypeScript transform of the whole e2e tree per config and boots the config's `globalSetup`, which for this repo resets the local Supabase database. This gate lives in `pnpm run check`, whose entire value is finishing in ~9s without touching a database — the local memory-pressure gate blocks a bare `playwright test --list` outright, which is a fair proxy for its weight. The bead explicitly permits the glob implementation.

**The configs are parsed, not hardcoded.** Hardcoding "e2e/full and e2e/smoke are covered" would encode a snapshot of the configs and go stale the moment one changed — the same class of rot the gate exists to catch. Instead the two configs' top-level `testDir`/`testMatch` are read out of their source. The parse is deliberately strict: anything it does not recognise (a `testMatch` that is a constant, a regex, or a ternary; a `testDir` inherited only via the `...baseConfig` spread) is a hard **exit 3** telling you to update the gate, never a silent assumption of coverage.

Known limitation, stated in the module docstring rather than papered over: only the **top-level** `testDir`/`testMatch` are read. Playwright also lets an individual `projects` entry narrow `testMatch` further, which `playwright.config.smoke.ts` does for firefox and Mobile Safari when running locally. Coverage is the union over projects, so a per-project narrowing only matters if *every* project in a config narrows away from a spec — no config does that today, and modelling per-project matchers would be a much larger parser with more ways to be wrong.

Exit codes: `0` clean · `1` uncovered spec(s) · `2` a suite config or the spec root is missing · `3` a config's shape is not recognised.

## Proof it works — by construction, not assertion

Created a temporary spec at a path no config collects, ran the gate, then moved it around:

| Spec location | Gate |
| :--- | :--- |
| `e2e/profiles/orphan-probe.spec.ts` | **exit 1** |
| `e2e/full/orphan-probe.spec.ts` | exit 0 |
| `e2e/orphan-probe.spec.ts` (e2e root) | **exit 1** |
| removed | exit 0 |

The red output:

```
check:e2e-suites: e2e spec(s) matched by no suite config -- nothing runs them

  e2e/profiles/orphan-probe.spec.ts

Suite configs considered:
  playwright.config.full.ts: testDir=e2e testMatch=**/full/**/*.spec.ts
  playwright.config.smoke.ts: testDir=e2e/smoke testMatch=(default)

Move the spec under a directory a suite already collects, or widen a suite
config to collect it. A spec that no suite collects type-checks and lints
green while running nowhere (PP-stut).
```

The probe was removed; the tree is clean. Green on current `main` — **no unmatched specs found**, all 53 spec files are collected.

## Tests

`scripts/tests/test_check_e2e_suite_coverage.py`, 37 cases. The two that matter most: the gate goes red for a spec no config collects, and it fails **loudly** rather than assuming coverage when a config's shape is unreadable. Also covered: glob semantics (`**/` spans segments, `*` does not cross `/`, bare patterns get Playwright's `**/` prefix, match anchored at end), comment stripping that leaves `//` inside a string alone, project-level keys not leaking up to the top level, and an absent `e2e/` root exiting 2 rather than passing vacuously.

Writing the tests found a real bug before this shipped: a commented-out `testDir` was being glued onto the front of the next real key, which made the parse return the wrong value. Fixed by stripping comments in one pass up front.

## Deliberately out of scope

The bead's second-order defect — `pnpm run e2e:mobile` is `playwright test --project='Mobile Chrome'` with no spec path and no `--config`, the bare whole-tree invocation AGENTS.md §5 forbids — is filed as **PP-sj3k** rather than folded in. It is a `package.json` script change with its own blast radius, and it is also the only thing that would have collected PP-stut's orphaned spec, so it silently "covers" specs no suite config claims. This PR is the gate alone.

## Review round 1 — `/code-review high` (3 medium/low findings + 1 doc)

All three code findings were this gate's own stated promise ("anything it does not recognise is a hard failure, never a silent assumption of coverage") being violated in ways that made it go **green** while a spec ran nowhere. Fixed in `54720d4`.

**1. `testIgnore` was parsed and then dropped.** `top_level_entries` read it; `read_suite_config` ignored it. That is the worst of the three, because it never even reached the unknown-shape error path — it silently widened coverage. Now subtracted in `collects()`. Implemented rather than raised: the semantics are unambiguous (matching files are excluded) and raising would have made a legitimate config unusable.

**2. An absent `testMatch` meant "match everything".** The asymmetry was backwards — the value whose *absence widens coverage* was the permissive one while an absent `testDir` was a hard error. Raising on absence was not available: `playwright.config.smoke.ts` ships no `testMatch`, so that would hard-fail the repo. Fixed by modelling Playwright's documented default, `**/*.@(spec|test).?(c|m)[jt]s?(x)`. Absence is a known shape with a known meaning; it is the *spread* that is the unknown. This is why `glob_to_regex` now understands extglob groups (`@()`, `?()`, `*()`, `+()`) and bracket expressions — the default pattern uses both. Negated extglob `!(` raises rather than guessing.

**3. Spreads are now resolved, not skipped.** Both suite configs are `{...baseConfig, ...}`, so with spreads skipped the moment `playwright.config.ts` grew a `testMatch` this gate would keep using the wider default and call dead specs covered. A spread of a plain identifier default-imported from a relative module is followed and merged in **source order** (later members win, exactly as JavaScript does it — there is a test for a spread placed *after* a key overriding it). Any other spread — conditional, inline object, bare-package import — raises, as the reviewer asked.

**4. (doc) Known-limitation paragraph now covers both directions.** It described per-project `testMatch` *narrowing* (risk: false GREEN). Per-project `testDir` **widening** is the opposite risk and produces a false **RED** on working specs — `playwright.config.smoke.ts` already has a project overriding `testDir` (`auth-setup`), harmless today only because its `testMatch` is the single file `auth.setup.ts`. Both are now documented, with the escape hatch named.

**Also corrected: the `*.test.ts` exclusion rationale was false.** It claimed including `*.test.ts` would flag a Vitest unit test for an e2e helper. But all three Vitest projects root their `include` at `src/**`, so Vitest never collects anything under `e2e/` — a file at `e2e/support/foo.test.ts` is run by *nothing*. Behavior changed rather than just the comment: `*.test.ts` is now enumerated.

### Verified against the real configs, not just fixtures

| Temporary edit | Result |
| :--- | :--- |
| `testIgnore: "**/machine-*.spec.ts"` on the full config | **exit 1**, all 7 `e2e/full/machine-*.spec.ts` reported uncovered |
| `testMatch` added to the base `playwright.config.ts` | **exit 1**, all 17 smoke specs reported uncovered — the spread resolves |
| both reverted | exit 0 |

The second is the load-bearing one: `playwright.config.smoke.ts`'s effective `testMatch` changed from the Playwright default to the base config's pattern, which is only visible if the spread is followed. Confirmed the real base config's 12 top-level keys are read through it. Modelling the default kept the repo green — the gate now reports `testMatch=**/*.@(spec|test).?(c|m)[jt]s?(x)  (Playwright default; config sets none)` for smoke.

Tests: **64 cases**, up from 37, with a named regression test per finding (`test_regression_test_ignore_must_not_be_read_then_dropped`, `test_regression_test_ignore_on_a_default_match_config`, `test_regression_base_config_test_match_reaches_the_suite_that_spreads_it`).

**Noted, no change:** the CI step is partly redundant with `test_real_repo_is_clean`, which runs `main([])` against the real tree in the same job. Keeping both — the dedicated step gives a better failure message, and the test additionally proves the real `playwright.config.ts` parses through the spread.

## Coordination

No `e2e/**/*.spec.ts` files were touched, so there is no collision with #1861's 53-file fixtures sweep. The gate matches only `*.spec.ts`, so #1861's new `e2e/support/fixtures.ts` is not enumerated — there is an explicit test for that (`test_support_modules_never_trigger_a_failure`). The `ci.yml` edit is a step in the `linters` job (~line 300); #1861's `ci.yml` edits are in the E2E jobs (~line 830+).

## Gates run

- `pnpm run check` — green (includes the new `check:e2e-suites`)
- `pnpm run check:python` — green, 566 pytest cases
- Self-reviewed by hand against `REVIEW.md` and the bead's acceptance criteria

## Review

**Needs Tim's `/code-review`.** A new CI gate is not a trivial change, so this is not self-attested.


timothyfroehlich:
Claude review of head 54720d4 — `/code-review high` — testIgnore parsed then dropped, absent testMatch read as match-everything, and skipped config spreads all let the gate go green on a dead spec; fixed by subtracting testIgnore, modelling Playwright's default testMatch, and resolving spreads in source order

timothyfroehlich:
Closing unmerged — scope call, not a quality problem.

The gate worked and Tim's `/code-review high` findings were all fixed. But addressing them grew `scripts/check_e2e_suite_coverage.py` from 428 to 638 lines of hand-rolled TypeScript config parsing — spread resolution, extglob translation, module-path resolution — to catch a bug that has occurred **once** (PP-stut, fixed in #1849). All 52 e2e specs live in exactly two directories, so the same failure is catchable by a ~15-line directory-convention check if we want one at all.

Tim's call: shelve it and put the effort into code coverage instead. Branch `feat/e2e-suite-coverage-gate-PP-2g7m` stays on origin if we want to revive it.

Bead PP-2g7m reopened with the reasoning; PP-sj3k (bare `pnpm run e2e:mobile` invocation) is unaffected and stays open.

PR #1871, opened 2026-08-12
feat(huddle): derive session names from the herdr workspace label (PP-355h)

timothyfroehlich:
## What

A session carried three unrelated names:

| Name | Was set by | Example |
| --- | --- | --- |
| Huddle name | model invents it from the task | `Claude-HuddleNaming` |
| Display name | `<dir>-<hash>` fallback | `pinpoint-97` |
| Terminal title | drifting conversation summary | `Messaging system and name selection…` |

Nothing tied them together. This derives the huddle name from the **herdr workspace label** — which Tim already types when he creates the workspace — and then asks him to run `/rename <name>` so the display name matches.

Naming order in the registration notice:

1. Workspace label, CamelCased (`Agent naming` → `Claude-AgentNaming`)
2. The assigned bead, when the label is generic (`PinPoint`, `Orchestrating`, `Busywork`, `main`, a bare number)
3. The task, when there is neither

## Why `$HERDR_WORKSPACE_ID`

Hooks inherit the environment herdr launched the agent in, so the variable names the workspace directly — no pane matching. Both alternatives misidentify the session:

- `herdr pane list`'s `agent_session.value` has been measured pointing at a **different live session** (documented in the `herdr` skill). That is PP-788v's failure mode — signing as someone else, silently.
- `terminal_title` carries Claude Code's animating activity glyph, and `terminal_title_stripped` only partly removes it: `✳` and the braille frames yes, `◐`/`◑` no.

A forked session inherits a stale `$HERDR_WORKSPACE_ID`. Tolerable here and nowhere else in the huddle, because the output is a *suggestion* both the agent and Tim read before it is registered.

## Note for the reviewer

`/rename` exists and I initially reported it did not — I searched the compiled binary's strings and concluded from a null result. It is user-typed only, so an agent still cannot self-name, but it does mean the name can be set *after* the task is known, which is why this lands as a notice change rather than a launcher wrapper.

CamelCasing goes through `jq`, not `sed`: BSD sed has no `\u`, so `s/(^| )([a-z])/\1\u\2/g` emits a literal "u" on this Mac — `Agent naming` became `Agentunaming`.

## Testing

`pnpm run check` and `pnpm run check:python` (530 passed) both green.

Hook output verified in three states:

- **label present** → `1. Your herdr workspace label, which is "Agent naming"… Claude-AgentNaming`
- **`$HERDR_WORKSPACE_ID` unset** → drops to bead-then-task, prints why
- **unknown workspace id** → same fallback, no error

Fails open at every step: no herdr, a bare terminal, Bazzite, another harness.

### DIFF
--- a/.agents/skills/pinpoint-huddle/SKILL.md
+++ b/.agents/skills/pinpoint-huddle/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: pinpoint-huddle
-description: The conventions behind the inter-session coordination channel that the huddle hooks inject but do not state — which events are auto-posted versus the judgment calls only you can post (added scope is the most-skipped and most-needed), peer-response etiquette, the em-dash self-filter signature, why a session_id belongs to whoever registered it first, why a dispatched subagent is refused registration outright, and the rotation subagent dispatch the session-start notice routes here for. Also how to read the work digest, the env knobs that quiet the nudge and the poll (`HUDDLE_NUDGE_SECONDS`, `HUDDLE_THROTTLE_SECONDS`), and the multi-machine setup — one shared Dolt server, `config.json` as a rebuildable cache rather than a source of truth, `today_bead.id` as a hint that gets verified. Use when you see a rotation-needed notice, when deciding whether something is worth posting to the daily bead, when your own posts start re-injecting, when a peer's update scrolls by, when the huddle is too noisy, or when a huddle lookup returns null on a machine.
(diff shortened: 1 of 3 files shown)


timothyfroehlich:
Claude review of head e8ef283 — `/code-review medium` — 6 findings: bounded the unbounded herdr socket read with timeout 1 (it sat in a 5000ms hook already running ~4s; over-budget discards ALL stdout and the session never registers — PP-2m3l, verified 30.7s vs ~1s); made the notice harness-neutral (<Harness>- not Claude-, conditional rename step) since Antigravity/Codex run this hook; guarded empty camelization from offering a bare <Harness>-; documented the moved-pane staleness case; popped HERDR_WORKSPACE_ID in the test harness so it stops reaching the real herdr. Declined one: /rename is type:local-jsx requires:{ink:true}, not supportsNonInteractive, so user-typed-only stands. +5 tests.

PR #1872, opened 2026-08-12
fix(hooks): anchor node hook commands to CLAUDE_PROJECT_DIR (PP-iapy)

timothyfroehlich:
## Summary

Every **node** hook command in `.claude/settings.json` was invoked by a bare relative path (`node .claude/hooks/x.cjs`), while every **bash** hook was already anchored to `${CLAUDE_PROJECT_DIR:-.}`. When the session cwd moved off a directory that happens to contain `.claude/hooks` (a removed worktree, a mid-session cwd change), node exited `MODULE_NOT_FOUND`, Claude Code reported a non-blocking hook failure, and **the tool call proceeded anyway** — taking the guard hooks (`block-direct-merge`, `block-main-worktree-branch-switch`, `block-worktree-dispatch-from-linked`) silently inert with them. Full repro and incident notes: PP-iapy.

- Anchors all seven node hook commands (`normalize-workspace-paths`, `inject-beads-actor`, `block-direct-merge`, `block-main-worktree-branch-switch`, `block-worktree-dispatch-from-linked`, `ui-screenshot-reminder`, `verify-guard-stack`) to `${CLAUDE_PROJECT_DIR:-.}`, matching the exact quoting the bash hooks in the same file already use.
- `verify-guard-stack.cjs`'s `extractScriptPaths()` already special-cased the `${CLAUDE_PROJECT_DIR:-.}` substitution pattern, so its own reverse dead-registration check needed no code change — it was written anticipating this fix.
- Grepped the repo (`rg --hidden 'node \.claude/hooks'`) for the same bare-relative-path shape elsewhere. Found only historical/inert references, left untouched: a comment illustrating the *old* form in `verify-guard-stack.cjs` (still valid as documentation), a dated planning doc under `docs/plans/` and `docs/superpowers/plans/`, and test fixtures in `src/test/unit/hooks/verify-guard-stack.test.ts` that construct synthetic settings objects (not the real settings.json, so not affected by this bug). No `.claude/settings.local.json` exists in this repo. Nothing outside `.claude/settings.json` needed a functional fix.

## Out of scope

The bead's DESIGN section also floats making the guard hooks fail **closed** when a hook can't load (rather than warn/proceed). That is an open design decision left for Tim — not implemented here.

## Verification

`pnpm run check` (static gate) — passed:
```
$ npm-run-all --silent --parallel typecheck typecheck:tests typecheck:e2e lint:local format:fix check:linters check:rule-ids audit:role-checks audit:sm-structural check:behind-main
All checks passed!
37 files already formatted
No findings to report. Good job! (1 ignored, 10 suppressed)
```

`pnpm run check:python` (ruff + pytest scripts/tests/) — passed, 547 tests:
```
======================= 547 passed in 111.26s (0:01:51) ========================
```

Manual reproduction — anchored form resolves from an unrelated cwd with `CLAUDE_PROJECT_DIR` set (as Claude Code sets it); the old bare form still fails the same way described in the bead:
```
$ cd /tmp/unrelated-cwd
$ CLAUDE_PROJECT_DIR=<worktree-root> bash -c 'echo "{}" | node "${CLAUDE_PROJECT_DIR:-.}"/.claude/hooks/verify-guard-stack.cjs'
exit=0

$ CLAUDE_PROJECT_DIR=<worktree-root> bash -c 'echo "{}" | node .claude/hooks/verify-guard-stack.cjs'
node:internal/modules/cjs/loader:1503
  throw err;
Error: Cannot find module '/tmp/unrelated-cwd/.claude/hooks/verify-guard-stack.cjs'
  code: 'MODULE_NOT_FOUND'
exit=1
```

## New test coverage

`scripts/tests/test_node_hook_paths.py` (17 tests):
- Reads the real `.claude/settings.json` (not a fixture copy) and asserts every wired node-hook command is `${CLAUDE_PROJECT_DIR:-.}`-anchored, one test per hook.
- For each of the seven hooks, actually runs the exact command string from `settings.json` with cwd set to an unrelated tmp dir and `CLAUDE_PROJECT_DIR` set to the repo root — asserts no `MODULE_NOT_FOUND` / "Cannot find module" in stderr. This is the direct regression catch: reintroducing a bare relative path fails this test.
- A control test proves the bare-relative-path form genuinely does fail under the same conditions, so the regression assertions aren't vacuous.
- A sanity test asserts all seven expected hook basenames are still wired at all.

## Test plan

- [x] `pnpm run check`
- [x] `pnpm run check:python`
- [x] Manual repro: anchored form resolves from unrelated cwd, bare form fails the same way as the bead's reported banner
- [ ] CI green (watching)

https://claude.ai/code/session_01Bj5S3d6aaKfxqCtcBqeVwn

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -66,12 +66,12 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node .claude/hooks/normalize-workspace-paths.cjs",
+            "command": "node \"${CLAUDE_PROJECT_DIR:-.}\"/.claude/hooks/normalize-workspace-paths.cjs",
             "timeout": 5000
           },
           {
             "type": "command",
-            "command": "node .claude/hooks/inject-beads-actor.cjs",
+            "command": "node \"${CLAUDE_PROJECT_DIR:-.}\"/.claude/hooks/inject-beads-actor.cjs",
             "timeout": 5000
           }
         ]
@@ -81,7 +81,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node .claude/hooks/block-direct-merge.cjs",
+            "command": "node \"${CLAUDE_PROJECT_DIR:-.}\"/.claude/hooks/block-direct-merge.cjs",
             "timeout": 5000
           }
         ]
@@ -91,7 +91,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node .claude/hooks/block-main-worktree-branch-switch.cjs",
+            "command": "node \"${CLAUDE_PROJECT_DIR:-.}\"/.claude/hooks/block-main-worktree-branch-switch.cjs",
             "timeout": 5000
           }
         ]
@@ -101,7 +101,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node .claude/hooks/block-worktree-dispatch-from-linked.cjs",
+            "command": "node \"${CLAUDE_PROJECT_DIR:-.}\"/.claude/hooks/block-worktree-dispatch-from-linked.cjs",
             "timeout": 5000
           }
         ]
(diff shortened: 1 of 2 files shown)


timothyfroehlich:
Claude review of head c728307 — `/code-review low` — No findings — 7 node hook commands re-anchored to "${CLAUDE_PROJECT_DIR:-.}" matching the existing bash-hook form; new pytest file skipped at this level.

PR #1873, opened 2026-08-12
fix(e2e): wait for React hydration after navigation (PP-kz47)

timothyfroehlich:
Split out of #1861 at Tim's direction, so that PR stays a self-contained CI-gate fix. This is the other half — the reason the Mobile Chrome full suite was failing.

## The mechanism

Playwright's actionability checks are visible, stable, enabled and receives-events. None of them know whether React has attached a handler, so "clickable" and "does something" are different facts and only the first is checked. A click in that window lands on a live-looking control and is silently discarded — nothing errors, nothing is pending afterwards — and the spec fails wherever it next asserts the consequence.

That is why one mechanism looked like four unrelated bugs: a Radix Select portal that never opened (`status-overhaul:25`), a comment Sheet that never appeared (`form-resets:190`, `rich-text:105`), a tab navigation that never happened (`technician-role:89`). It is also why raising timeouts did nothing — a slow mount is rescued by a longer wait, a dropped click is not.

**Evidence it is one mechanism and not four bugs**, from the Mobile Chrome full suite on the Bazzite runner: serial `--workers=1` passes 92/92, `--workers=3` fails about one spec per run, every victim passes in isolation, and the victim rotates.

Mobile Chrome is the only project affected because the E2E `webServer` runs `pnpm run dev` — routes compile on first request — and mobile renders component trees chromium never compiles (`MetadataDrawer`, `StickyCommentComposer`, `RowEditSheet`), so it pays extra compiles while two other workers keep the server busy.

## What this does

- `HydrationBeacon` in `ClientProviders` sets `data-hydrated` on `<html>`. Ships in production deliberately: an attribute that exists only under test is one whose absence nobody notices when it breaks.
- `e2e/support/fixtures.ts` exports a project-local `test` whose `page` waits for that attribute after `goto` **and** `reload`. 52 specs swept onto it.
- `attachHydrationWait` is exported for the 11 pages built via `browser.newContext().newPage()`, which the fixture never sees.
- `/api/*` is excluded rather than waited for — those render outside the root layout and have no React root. `/api/unsubscribe` returns hand-rendered `text/html`, so content-type could not be the discriminator.
- The wait does not swallow its own timeout. That would hide a real hydration break behind 30s per navigation and then fail at whatever got clicked next, pointing at the wrong thing.

## Latent defects fixed along the way — none of which caused this

- `openDropdownMenu` clicks are bounded (20s first, 10s retry) so the worst path is 50s and its authored diagnostic can print inside the 60s CI test timeout, instead of dying as a bare "Test timeout exceeded".
- `openIssueCommentForm` decides by viewport width rather than sampling with `isVisible()` (which never retries, and whose `timeout` option is deprecated and ignored), and retries only when the dialog is absent from the DOM — a blind second click gets eaten by Radix's overlay.
- Two more `isVisible()` samples became real waits.

## Known incomplete

Stated rather than papered over: the beacon fires from an effect in the **root** layout, so it reports the root commit — not that every interactive descendant has handlers. Suspense-boundary children (any segment with a `loading.tsx`) and `ssr: false` imports are still dehydrated when the wait returns. The client-side-navigation half of the same class is **PP-2b3r**, still open.

This narrows the window. It does not close it.

## Verification

Bazzite runner with `CI=1` (crabbox leaves `$CI` unset, which is how a timeout conclusion got written wrong earlier in this work): 33/33, 50/50 and 26/26 across the touched specs on chromium + Mobile Chrome.

Bead: PP-kz47. Sequencing: independent of #1861, but #1861 should land first.

## Review round (`/code-review high`, c3d79eb4)

Six findings, all addressed. Two were substantive:

- **`submitFormAndWaitForRedirect` returned before hydration.** It resolves on `waitForURL(…, { waitUntil: "commit" })` — a full document navigation the fixture cannot see. Eight specs reach the issue page through it, including `form-resets` and `rich-text`, so for the two victims this PR names, the fixture never actually covered the navigation that produced the failure. `waitForHydration` is now exported and awaited once at the end of that helper.
- **A hydration failure was swallowed as a lost race.** Branches B/C catch any `goto` throw as "a sibling branch won", release the claim and park on a never-resolving promise — turning the authored diagnostic into a bare 60s timeout. It is now a named `HydrationTimeoutError` that both branches re-throw; Branch C had no reject path and got one.

Four smaller: `dialog.count()` couldn't distinguish "never opened" from "mounted but not yet in the a11y tree" (`includeHidden: true`); the comment blaming a missing `aria-expanded` was wrong (Radix sets it — the cause is `hideOthers()` aria-hiding the trigger); `attachHydrationWait` wasn't idempotent; `technician-role`'s 5s wait wasn't CI-aware.

Re-validated on the Bazzite runner after the fixes: smoke 201 passed / 1 flaky (`collections`, unrelated), e2e-full 276 passed / 0 flaky.

### DIFF
--- a/e2e/full/technician-role.spec.ts
+++ b/e2e/full/technician-role.spec.ts
@@ -7,7 +7,7 @@
  * - CANNOT access admin panel
  */
 
-import { test, expect } from "@playwright/test";
+import { test, expect } from "../support/fixtures.js";
 import { STORAGE_STATE } from "../support/auth-state.js";
 import { seededMachines } from "../support/constants.js";
 
@@ -20,7 +20,15 @@ test.describe("Technician Role Permissions", () => {
     const heading = page.getByRole("heading", {
       name: seededMachines.addamsFamily.name,
     });
-    if (await heading.isVisible()) {
+    // Waited for, not sampled: `isVisible()` never retries, so a heading that
+    // had merely not painted yet read as "name was changed" and sent this
+    // afterEach down the restore branch — an unnecessary DB write on every
+    // slow render.
+    const nameIsCorrect = await heading
+      .waitFor({ state: "visible", timeout: 5000 })
+      .then(() => true)
+      .catch(() => false);
+    if (nameIsCorrect) {
       // Name is already correct — nothing to do
       return;
     }
--- a/e2e/full/technician-role.spec.ts
+++ b/e2e/full/technician-role.spec.ts
@@ -23,9 +23,11 @@ test.describe("Technician Role Permissions", () => {
     // Waited for, not sampled: `isVisible()` never retries, so a heading that
     // had merely not painted yet read as "name was changed" and sent this
     // afterEach down the restore branch — an unnecessary DB write on every
-    // slow render.
+    // slow render. CI-aware because a flat 5s reopens the same hole under the
+    // conditions this suite actually fails in (Mobile Chrome, three workers, a
+    // dev server compiling routes for the other two).
     const nameIsCorrect = await heading
-      .waitFor({ state: "visible", timeout: 5000 })
+      .waitFor({ state: "visible", timeout: process.env["CI"] ? 15_000 : 5000 })
       .then(() => true)
       .catch(() => false);
     if (nameIsCorrect) {
--- a/e2e/support/fixtures.ts
+++ b/e2e/support/fixtures.ts
@@ -35,7 +35,7 @@
  *
  * ## What it does not cover
  *
- * Two real gaps, both narrower than "it's handled":
+ * Four real gaps, all narrower than "it's handled":
  *
  * 1. **Pages this fixture never sees.** A page from
  *    `browser.newContext().newPage()` is constructed by the spec, not by the
@@ -47,7 +47,14 @@
  * 2. **Route handlers.** `/api/*` renders outside the root layout and has no
  *    React root, so it is excluded rather than waited for — see `isAppPage`.
  *
- * 3. **Subtrees that hydrate after the root.** The beacon fires from an effect
+ * 3. **Navigations that are not `goto`/`reload`.** A form submit that redirects
+ *    is a real document navigation this fixture never sees, and
+ *    `submitFormAndWaitForRedirect` resolves at `waitUntil: "commit"` —
+ *    earlier still. That helper therefore calls {@link waitForHydration}
+ *    itself. Any future helper that resolves on a navigation must do the same;
+ *    a link click followed by `waitForURL` has the identical hole.
+ *
+ * 4. **Subtrees that hydrate after the root.** The beacon fires from an effect
  *    in the root layout, so it reports that the root commit landed — not that
  *    every interactive descendant has handlers. Anything inside a Suspense
  *    boundary (any segment with a `loading.tsx`, plus streamed segments) is
@@ -101,7 +108,27 @@ function isAppPage(page: Page): boolean {
   }
 }
 
-async function waitForHydration(page: Page, target: string): Promise<void> {
+/**
+ * Thrown when the beacon never appears. Named so that callers which catch
+ * broadly — `submitFormAndWaitForRedirect`'s branch races treat a `goto` throw
+ * as "a sibling branch got there first" — can re-throw this one instead of
+ * swallowing it into a bare test timeout.
+ */
+export class HydrationTimeoutError extends Error {
+  override readonly name = "HydrationTimeoutError";
+}
+
+/**
+ * Wait until React has hydrated the current document.
+ *
+ * Exported because navigation does not only happen through `goto`:
+ * `submitFormAndWaitForRedirect` resolves on a URL commit, which this fixture
+ * cannot see, so that helper calls this itself.
+ */
+export async function waitForHydration(
+  page: Page,
+  target: string
+): Promise<void> {
   if (!isAppPage(page)) return;
 
   try {
@@ -115,7 +142,7 @@ async function waitForHydration(page: Page, target: string): Promise<void> {
     // during mount, say. Swallowing it would spend HYDRATION_TIMEOUT on every
     // navigation and then fail later at whatever got clicked next, pointing at
     // the wrong thing.
-    throw new Error(
+    throw new HydrationTimeoutError(
       `React never hydrated within ${String(HYDRATION_TIMEOUT)}ms after navigating to ${target}. ` +
         `Expected \`${HYDRATED_SELECTOR}\` (set by HydrationBeacon in ClientProviders). ` +
         `A client component most likely threw during mount — check the browser console in the trace.`
@@ -124,13 +151,24 @@ async function waitForHydration(page: Page, target: string): Promise<void> {
 }
 
 /**
+ * Pages already wrapped. Wrapping twice nests the wrappers, so a single `goto`
+ * would wait for hydration twice — and on a genuinely broken page would burn
+ * `2 × HYDRATION_TIMEOUT` (60s in CI) against a 60s test timeout, replacing the
+ * authored diagnostic with a bare "Test timeout exceeded".
+ */
+const wrapped = new WeakSet<Page>();
+
+/**
  * Make `goto` and `reload` on this page wait for hydration before returning.
  *
  * The `page` fixture below applies this for you. Call it directly on pages you
  * construct yourself — `browser.newContext()` then `newPage()` — which the
- * fixture never sees.
+ * fixture never sees. Idempotent, so passing an already-wrapped page is safe.
  */
 export function attachHydrationWait(page: Page): Page {
+  if (wrapped.has(page)) return page;
+  wrapped.add(page);
+
   const originalGoto = page.goto.bind(page);
   const originalReload = page.reload.bind(page);
 
--- a/e2e/support/fixtures.ts
+++ b/e2e/support/fixtures.ts
@@ -1,0 +1,159 @@
+/**
+ * Project-local `test` — identical to Playwright's, except that a navigation
+ * does not return until React has hydrated.
+ *
+ * ## Why this exists
+ *
+ * Playwright's actionability checks are visible, stable, enabled, and receives-
+ * events. None of them know whether React has attached a handler yet, so
+ * "clickable" and "does something" are different facts and only the first is
+ * checked. A click in that window lands on a live-looking control and is
+ * silently discarded — no error, nothing pending afterwards.
+ *
+ * The spec then fails wherever it next asserts the consequence, which is why
+ * the symptom looked like unrelated bugs: a Radix Select portal that never
+ * opened (`status-overhaul:25`), a comment Sheet that never appeared
+ * (`form-resets:190`, `rich-text:105`), a tab-strip navigation that never
+ * happened (`technician-role:89`). It is also why raising timeouts did nothing:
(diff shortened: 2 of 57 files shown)


e2e/support/page-helpers.ts
```diff
@@ -327,3 +336,90 @@ export async function fillReportForm(
...
+  // "false", and the retry get eaten by the overlay. Verified on Mobile Chrome:
+  // `form-resets:190` and `rich-text:105` failed exactly that way.
+  await sheetTrigger.click({ timeout: SHEET_FIRST_CLICK_TIMEOUT });
+  const openedFirstTry = await dialog
+    .waitFor({ state: "visible", timeout: SHEET_OPEN_TIMEOUT })
+    .then(() => true)
+    .catch(() => false);
+
+  if (!openedFirstTry) {
+    if ((await dialog.count()) > 0) {
```
timothyfroehlich: **`dialog.count()` cannot make the distinction this branch exists to make.**

`dialog` is `getByRole("dialog", …)`, and `getByRole` defaults to `includeHidden: false` — it only matches elements that are in the accessibility tree. So `count()` returns 0 both when the Sheet was never opened *and* when it is mounted but not yet ARIA-visible, which is precisely the "mounted but slow — the click registered" case the comment says this guard protects. Radix unmounts closed Sheet content outright, so the branch is near-dead as written and the blind retry runs in the case it was meant to prevent.

Concrete failure: the click registers, the Sheet commits just after `SHEET_OPEN_TIMEOUT` (10s in CI), `count()` reads 0, the retry fires — and because opening a Radix dialog calls `hideOthers()` on everything outside the content, `sheetTrigger` (a `getByRole("button")` locator) no longer matches, so the retry click dies after `SHEET_RETRY_CLICK_TIMEOUT` with "locator resolved to no elements" while the Sheet was open the whole time.

```suggestion
    if ((await mountedDialog.count()) > 0) {
```
where `mountedDialog` is the same locator with `includeHidden: true`. (Fixed in the working tree.) —Claude

e2e/support/page-helpers.ts
```diff
@@ -327,3 +336,90 @@ export async function fillReportForm(
...
+  //
+  // But a BLIND second click is actively harmful: if the first one worked and
+  // the Sheet is merely slow, Radix has already marked the outside content
+  // `aria-hidden` and laid an overlay over it, so re-clicking either toggles
+  // the Sheet shut or is intercepted — turning a recoverable slow mount into a
+  // hard failure. So the retry is conditional on evidence the click did NOT
+  // register: the dialog is absent from the DOM entirely. If it is mounted but
+  // not yet visible, the click landed and the right move is to keep waiting.
+  //
+  // Note this deliberately does NOT go through `openDropdownMenu`. That helper
```
timothyfroehlich, line 438: **The stated reason for skipping `openDropdownMenu` is not the real one, and the wrong version is the kind that invites a bad "fix".**

Radix's dialog `Trigger` *does* set `aria-expanded` — `node_modules/@radix-ui/react-dialog/dist/index.mjs:77`, `"aria-expanded": context.open`. Vaul re-exports that same Trigger, which is why the `MetadataDrawer` `-trigger` path in `selectOption` works fine through `openDropdownMenu`.

What actually breaks here is line 145 of that same file: opening the dialog calls `hideOthers(content)`, marking everything outside the content `aria-hidden="true"`. `sheetTrigger` is a `getByRole("button")` locator, and `getByRole` skips ARIA-hidden elements, so the moment the Sheet opens the locator matches nothing — the `toHaveAttribute` assertion fails as "no elements" (reading, in the trace, like the attribute never flipped) and the retry click has nothing to click.

Worth correcting, because as written the note reads "this component is missing `aria-expanded`" and the obvious follow-up is to add it to the button — which is already there. The transferable rule is narrower: never assert on a *role* locator for a control that an open modal aria-hides. (Rewritten in the working tree.) —Claude

e2e/support/fixtures.ts
```diff
@@ -0,0 +1,159 @@
...
+}
+
+/**
+ * Make `goto` and `reload` on this page wait for hydration before returning.
+ *
+ * The `page` fixture below applies this for you. Call it directly on pages you
+ * construct yourself — `browser.newContext()` then `newPage()` — which the
+ * fixture never sees.
+ */
+export function attachHydrationWait(page: Page): Page {
```
timothyfroehlich, line 168: **`attachHydrationWait` is not idempotent, and it is exported precisely so spec authors will call it by hand.**

A second call on the same page captures the already-wrapped `goto`/`reload` as `originalGoto`/`originalReload` and nests, so one navigation waits for hydration twice. Harmless on a healthy page; on a broken one it burns `2 × HYDRATION_TIMEOUT` = 60s in CI against the 60s test timeout, so the authored "React never hydrated" diagnostic never prints and you get a bare "Test timeout exceeded" — the exact failure mode the `openDropdownMenu` budgets in this PR were bounded to avoid.

Nothing double-wraps today, but the doc block invites hand-calls on pages a helper might also have wrapped. A `WeakSet<Page>` guard at the top is enough. (Added in the working tree.) —Claude

e2e/full/technician-role.spec.ts
```diff
@@ -20,7 +20,15 @@ test.describe("Technician Role Permissions", () => {
     const heading = page.getByRole("heading", {
       name: seededMachines.addamsFamily.name,
     });
-    if (await heading.isVisible()) {
+    // Waited for, not sampled: `isVisible()` never retries, so a heading that
+    // had merely not painted yet read as "name was changed" and sent this
+    // afterEach down the restore branch — an unnecessary DB write on every
+    // slow render.
+    const nameIsCorrect = await heading
+      .waitFor({ state: "visible", timeout: 5000 })
```
timothyfroehlich: 5s here is not CI-aware, unlike every other budget this PR introduces (`PORTAL_MOUNT_TIMEOUT`, `DROPDOWN_*`, `SHEET_*`, `HYDRATION_TIMEOUT` all branch on `$CI`).

Under the conditions the PR says the suite actually fails in — Mobile Chrome, three workers, the dev server compiling routes for the other two — a heading that has simply not rendered in 5s still reads as "name was changed", which sends this `afterEach` down the restore branch: a needless DB write, and a hard failure inside `afterEach` if `machine-tab-edit` is not there for this role/page state. That is the misbehaviour the change was written to remove, just with a smaller window.

```suggestion
      .waitFor({ state: "visible", timeout: process.env["CI"] ? 15_000 : 5000 })
```
(Applied in the working tree.) —Claude

timothyfroehlich:
## Code review — two findings that land outside the diff hunks

Both are in `e2e/support/page-helpers.ts`, on unchanged lines that the new fixture nonetheless changes the behaviour of, so GitHub will not take them inline.

### 1. `submitFormAndWaitForRedirect` swallows a genuine hydration failure (lines 168 and 210)

Branches B and C call `page.goto(target, { waitUntil: "domcontentloaded" })`. Those calls now go through the fixture wrapper, so they can throw the authored `React never hydrated within …ms` error — and both are inside a `catch` that reads any `goto` throw as "raced a sibling branch": it calls `releaseClaim()` and then `await new Promise(() => undefined)`, i.e. never resolves.

So on a real hydration break during a Server Action redirect the outcome is a bare 60s `Test timeout of 60000ms exceeded`, pointing at nothing — which is precisely what `fixtures.ts:111-117` says it refuses to allow ("Swallowing it would … fail later at whatever got clicked next, pointing at the wrong thing"). Re-throwing when the error is the hydration one (or checking the message) would keep the diagnostic.

### 2. The helper returns before hydration, and it is the seam the two named victims navigate through (line 250)

`Promise.race` resolves as soon as Branch A's `waitForURL(…, { waitUntil: "commit" })` fires — the moment the new document commits, before it hydrates. That is a **full document navigation** the fixture cannot see, because it is neither `goto` nor `reload`.

`form-resets.spec.ts:222` and `rich-text.spec.ts:125` — the two specs the PR description names — reach the issue detail page exactly this way and then click "Add a comment". So for those two the fixture does not in fact cover the navigation that produced the failure; the only thing standing between them and a dropped click is the new `openIssueCommentForm` retry. Same for `machine-timeline`, `issues-reassign-machine`, `report-form-clear`, `image-upload`, `issues-crud`, `email-and-notifications`.

Exporting `waitForHydration` from `fixtures.ts` and awaiting it before this helper returns would close the highest-traffic remaining gap. I did **not** apply this one — it changes shared behaviour for eight specs and I cannot validate it without a full E2E run.

---

Four smaller findings are posted inline; the fixes for all four are in the working tree (uncommitted), not pushed. —Claude

e2e/support/page-helpers.ts
timothyfroehlich: Fixed in c3d79eb4 — matched with a second locator using `includeHidden: true`, so a non-zero count really does mean the click registered. You are right that the old form could not separate "never opened" from "mounted but not yet in the a11y tree", which is the only distinction the branch exists to make. —Claude

e2e/support/page-helpers.ts
timothyfroehlich: Fixed in c3d79eb4 — comment rewritten. You are right that Radix's dialog Trigger does set `aria-expanded`; the real cause is `hideOthers()` aria-hiding the trigger once the Sheet opens, so the role locator stops matching at all. The old wording would have sent someone to add an attribute that was already there. —Claude

e2e/support/fixtures.ts
timothyfroehlich: Fixed in c3d79eb4 — WeakSet guard, and the doc now says it is idempotent. It is exported specifically for hand use, so a double wrap was reachable. —Claude

e2e/full/technician-role.spec.ts
timothyfroehlich: Fixed in c3d79eb4 — now `CI ? 15_000 : 5000`, matching the other budgets in this PR. —Claude

timothyfroehlich:
Both fixed in c3d79eb4, and both were right.

**1. Swallowed hydration failure.** `HydrationTimeoutError` is now a named class exported from `fixtures.ts`, and Branches B and C re-throw it rather than treating it as a lost race. Branch C had no reject path at all — its executor only took `resolve` — so it got one.

**2. The helper returned before hydration.** `waitForHydration` is exported and awaited once at the end of `submitFormAndWaitForRedirect`, for whichever branch won the race. This was the more important of the two: the PR claimed the fixture covered `form-resets` and `rich-text`, and it did not — Branch A resolves at `waitUntil: "commit"`, which the fixture cannot observe. The fixture doc now lists this as gap 3 of 4, with the general rule that any future helper resolving on a navigation has the identical hole.

You declined this one because it changes shared behavior for eight specs and needs an E2E run. Ran it on the Bazzite runner against this tree:

- **smoke** — 201 passed, 8 skipped, 1 flaky (`collections.spec.ts` on Mobile Chrome, a dropped click on a dialog trigger that retried green; that spec does not touch this helper, and the failure is the PP-2b3r class this PR says it does not close)
- **e2e-full** — 276 passed, 3 skipped, 0 flaky

—Claude

timothyfroehlich:
Claude review of head c3d79eb — `/code-review high` — 6 findings, all fixed in c3d79eb4 — 2 substantive (submitFormAndWaitForRedirect returned pre-hydration; a hydration failure was swallowed as a lost race), 4 smaller; re-validated on the runner, e2e-full 276 passed / 0 flaky

PR #1874, opened 2026-08-12
fix(notifications): build Discord links from initials, not resource ids (PP-gzq2)

timothyfroehlich:
Discord notification DMs linked to `${siteUrl}/issues/<uuid>` and `${siteUrl}/machines/<uuid>`. Neither is a route — PinPoint addresses issues as `/m/<INITIALS>/i/<number>` and machines as `/m/<INITIALS>`, and `/machines` is not a path segment at all. **Every Discord notification link since the channel shipped has 404'd.**

Reported by Tim from a live DM:

```
New comment on RUSH-01 — Right Pop Bumper need rebuild
https://pinpoint.austinpinballcollective.org/issues/8da49c82-…
```

Email was never affected: `email-channel.ts` parses `RUSH-01` into initials + number and builds the canonical URL itself. Discord simply never got that logic.

## Approach

**One URL builder, two channels.** Copying email's parse into `messages.ts` would have made two places to get this wrong. It moves to `buildResourceUrl()` in `src/lib/notifications/resource-url.ts`; email now calls it too, so its inline version is deleted rather than duplicated.

**The dispatcher resolves what the link needs.** `machineInitials` is new on `ChannelContext` — it is the only handle that addresses a machine in a URL. `planNotification` resolves it (and backstops `formattedIssueId`) from the row when a caller doesn't pass it, so a future caller that forgets a field degrades to a list page instead of emitting another dead deep link. Both lookups are guarded and sit after the empty-recipient early return, so they cost nothing in the common path.

**Redirect stubs rescue the links already sent.** Discord messages can't be edited after delivery, so every DM in members' history keeps its dead URL forever unless the URL starts resolving. `/issues/[id]` and `/machines/[id]` look up the row and redirect to the canonical page. They render nothing and gate nothing — the redirect target runs the real `checkPermission()`.

## Judgment calls worth a look

- **Fallback is a list page, not a 404.** Unparseable initials → `/issues` or `/m` rather than a hard failure. Debatable; a 404 would be louder but strictly worse for the recipient.
- **The 7 machine-notification call sites still don't pass `machineInitials`.** The dispatcher's fallback covers them uniformly with one indexed PK lookup on a rare event. Passing it at each site would save that query at the cost of 7 chances for the `machine` select to lack the column.

## Verification

| Check | Result |
| :--- | :--- |
| `pnpm run check` | clean |
| `pnpm run test` | 2128 passed (230 files) |
| Smoke E2E (Bazzite runner) | 210 passed, 8 skipped, 1 flaky (`collections.spec.ts`, unrelated) |
| `legacy-resource-redirects.spec.ts` targeted | 3/3 green against a real server |
| Integration (Bazzite runner) | see comment below |

New coverage: 15 unit cases on the URL helper (both resource types, every malformed-input path, and an explicit assertion that the id-addressed shape is never emitted), updated Discord/email channel suites, and a smoke spec asserting both redirects land and an unknown id 404s.

No migration. No UI surface, so no screenshots.

Closes PP-gzq2.

https://claude.ai/code/session_01XXnCRVPp4xCxXL6FFeaVg9

### DIFF
--- a/e2e/smoke/legacy-resource-redirects.spec.ts
+++ b/e2e/smoke/legacy-resource-redirects.spec.ts
@@ -44,4 +44,20 @@ test.describe("Legacy resource-id redirects", () => {
 
     expect(response?.status()).toBe(404);
   });
+
+  // A non-UUID segment reaches a `uuid` column, where Postgres raises 22P02
+  // instead of returning no rows — so without a shape check these 500 through
+  // the error boundary and report to Sentry. Crawlers and truncated pastes hit
+  // this far more often than a well-formed but unknown id.
+  for (const path of [
+    "/issues/not-a-uuid",
+    "/issues/robots.txt",
+    "/machines/not-a-uuid",
+  ]) {
+    test(`${path} 404s instead of erroring`, async ({ page }) => {
+      const response = await page.goto(path);
+
+      expect(response?.status()).toBe(404);
+    });
+  }
 });
--- a/e2e/smoke/legacy-resource-redirects.spec.ts
+++ b/e2e/smoke/legacy-resource-redirects.spec.ts
@@ -1,0 +1,47 @@
+/**
+ * Smoke: legacy id-addressed links redirect to the canonical routes.
+ *
+ * Coverage goal: D-class — `/issues/<uuid>` and `/machines/<uuid>` were the URL
+ * shapes Discord notifications shipped until PP-gzq2, and neither was ever a
+ * route. Discord DMs can't be edited, so those links live forever and the app
+ * has to resolve them. E2E is the cheapest honest layer here: the thing under
+ * test is Next's routing + redirect, not a function's return value (the URL
+ * builder itself is unit-tested in src/lib/notifications/resource-url.test.ts).
+ */
+import { test, expect } from "@playwright/test";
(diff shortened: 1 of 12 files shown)


timothyfroehlich:
**Integration (Bazzite runner): 635 passed, 57 files, 0 failed.**

Includes `doodle-bug-regression.test.ts`, which is the test that would catch the two DB reads this PR adds inside the `planNotification` transaction window (CORE-ARCH-011). Both are reads, both are guarded, and both sit after the empty-recipient early return — but the regression test is the thing that proves it rather than the reasoning.

—Claude

timothyfroehlich:
Claude review of head d501d92 — `/code-review medium` — 2 findings, both fixed: (1) medium — /issues/[id] and /machines/[id] passed a raw path segment to a uuid column, so a non-UUID like /issues/robots.txt raised Postgres 22P02 and hit the error boundary + Sentry instead of 404ing; both stubs now z.uuid()-validate and notFound() first, with 3 new E2E cases. (2) low, pre-existing — machine_ownership_changed emails linked to the global /issues list under a 'View Issue' label; now /m/<INITIALS> and 'View Machine'. Review covered f280d712; since then 290a7692 (these findings) and d501d923 (merge of origin/main, CI workflow files only).

Issue #2, opened 2025-06-26
chore: Configure Dependabot for automated dependency updates


timothyfroehlich:
**Description:**
To keep our project dependencies secure and up-to-date, we should use GitHub's Dependabot. This will automatically create pull requests to update our npm dependencies.

**Acceptance Criteria:**
*   A `dependabot.yml` file is created in the `.github` directory.
*   The configuration is set up to check for `npm` dependency updates on a `daily` schedule.
*   The configuration targets the root directory (`/`).

**Technical Notes:**
*   This is a standard configuration file. The setup should be straightforward.
*   Refer to the [GitHub Dependabot documentation](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file) for configuration options.

**Depends On:** N/A

PR #1875, opened 2026-08-13
feat(pinballmap): derive the listing control, and let off-catalog games have a model (PP-o355.21, PP-3bbr)

timothyfroehlich:
## Summary

Three commits, one feature: what a machine's model is, whether Pinball Map lists it, and what to do when those two disagree.

- **Info tab.** The standalone Pinball Map card is gone. A whole card for two facts hid them — a reader scanning a machine's identity had to look in a second box to learn what game it is. Both are now rows in Details: a `Model` line, and one unlabelled line beneath it that is either "View on Pinball Map" or "Not listed on Pinball Map", always linking to our location's listing (CORE-PBM-001). A `Config issue` chip sits at the end of that line for members and up when something needs resolving, and links to Manage.
- **Manage tab.** The #1683 listing control is rebuilt on derived state. It made you press "Connect" to find out where you stood, and answered "Not listed" for a fleet that was in fact on the map; the replacement paints itself at page load from stored columns plus the stored snapshot, with no call to pinballmap.com. Connect / Verify / Reconnect are retired, `verifyPinballmapLinkAction` deleted outright.
- **Manual model entry (PP-3bbr).** Picking "Not on Pinball Map" opens a panel with Model name, Manufacturer and Year, so a homebrew or a flipperless game can have a real identity. Tim asked for this to be folded in — keeping it separate made the Model control impossible to design in one pass.

## Decisions worth a second look

**Seven derived states, not the six in the design.** `unclaimed_on_pbm` — their lineup carries the title but no PinPoint machine holds the listing — is split out of "not listed" because it is one of the two things the new Config-issue warning raises. A warning that sends you to a page saying nothing about it is a dead end. Claiming it writes nothing to Pinball Map.

**The Model row is no longer permission-gated.** The old card only rendered for `machines.pinballmap.link` holders, so an anonymous QR-scan visitor never learned what game the cabinet was. That seemed wrong for public information. Easy to put back.

**New permission `machines.pinballmap.diagnose`** (member and up), deliberately wider than `link`, which a member holds only for their own machines. A wrong entry on a public map is something any member may be the one to notice. Read-only — resolving it still needs `link` / `push`.

**`isActionableDesync`** keeps `lmx_drifted` out of the warning. The hourly reconcile heals it, so surfacing it reports a self-repairing transient to someone who can do nothing about it.

**New `machines.model_name` with a CHECK tying it to `pinballmap_excluded`**, so "linked AND hand-entered" is unrepresentable. That is what removes the precedence rule PP-3bbr worried about — there is no case where both exist, so no reader has to choose. It is also why switching to a catalog title confirms first: the values really are dropped.

## Test Plan

- [x] `pnpm run check` — green
- [x] Unit: 40 files / 301 tests across the touched areas, including a rewritten `PinballmapListingControl.test.tsx` (15), new `derivePbmListingState` cases, new `PinballMapLinkField` manual-entry cases (6), and a new `schemas.test.ts`
- [x] Integration: full suite green on the Bazzite runner; `pinballmap-link-capture.test.ts` (15) covers the new accept action and the hand-entered model round trip through `updateMachineAction`
- [ ] CI E2E — `machine-pinballmap-abandoned-notice.spec.ts` was rewritten to follow the notice to its new home on Manage and to assert the Info warning points at it
- [ ] Screenshots

## Related

PP-o355.21, PP-3bbr. Migration `0064` adds one nullable column and one CHECK.


timothyfroehlich:
## UI Screenshots — head `e1445dcd`

### Issues List

| --- | --- |
| desktop | mobile |

| --- | --- |
| desktop | mobile |

### Report an Issue

| --- | --- |
| desktop | mobile |

### Dashboard

| --- | --- |
| desktop | mobile |

| --- | --- |
| desktop | mobile |

### Collections

| --- | --- |
| desktop | mobile |

| --- | --- |
| desktop | mobile |

### New Machine

| --- | --- |
| desktop | mobile |

| --- | --- |
| desktop | mobile |

src/app/(app)/m/[initials]/(tabs)/edit/page.tsx
```diff
@@ -148,73 +244,58 @@ export default async function MachineEditPage({
...
           description={machine.description}
-          canLink={canLink}
+          canLink={canSetIntent}
           pinballmapMachineId={machine.pinballmapMachineId}
           pinballmapExcluded={machine.pinballmapExcluded}
           pinballmapExcludedReason={machine.pinballmapExcludedReason}
           pinballmapTitleName={pinballmapTitleName}
+          modelName={machine.modelName}
+          manufacturer={machine.manufacturer}
+          year={machine.year}
```
Comment (timothyfroehlich) [inline src/app/(app)/m/[initials]/(tabs)/edit/page.tsx]: **Saving an off-catalog machine silently wipes its hand-entered manufacturer and year.**

`machine` here comes from `getMachineForLayout`, which returns `{ ...machine, ...PBM_METADATA_PLACEHOLDER }` (`src/app/(app)/m/[initials]/_data.ts:79`) — and that placeholder hard-codes `manufacturer: null, year: null`. So `machine.manufacturer` / `machine.year` are **always** null on this page, regardless of what the row holds.

Scenario: FB (the seed's uncataloged machine) is stored with `manufacturer: 'homebrew', year: 1972`. Open its Manage tab — the PP-3bbr panel renders both inputs empty, because `defaultManufacturer`/`defaultYear` arrive null. Change nothing but the machine name and press Save details. `readPbmLinkFormFields` reads the two empty inputs as `undefined`, `resolveCore`'s excluded branch writes `manufacturer: input.manufacturer ?? null` and `year: input.year ?? null`, and both stored values are gone. Nothing reports it.

`modelName` is unaffected only because it happens not to be in the placeholder. Either drop `manufacturer`/`year` from `PBM_METADATA_PLACEHOLDER` (they are real columns now) or read them from a query that does not spread the placeholder over the row.

src/app/(app)/m/pinballmap-actions.ts
```diff
@@ -347,7 +337,7 @@ async function classifyRemoveNotFound(args: {
 
   const isFresh = (syncedAt: Date | null): boolean =>
     syncedAt !== null &&
-    Date.now() - syncedAt.getTime() < PBM_MANUAL_SYNC_MIN_INTERVAL_MS;
+    Date.now() - syncedAt.getTime() < PBM_REFRESH_REFILL_MS;
```
timothyfroehlich, line 347: **(about the call site at line 651)** — on the abandoned-entry path this can delete the machine's CURRENT lineup entry from Pinball Map.

`classifyRemoveNotFound` is handed `machine.pinballmapMachineId` — the machine's *current* title — but when `removeMachineFromPinballMapAction` was called with an explicit `lmxId` (the abandoned-entry alert), `deletedLmxId` belongs to a title the machine no longer carries. So the "is it still there under a re-minted id?" lookup runs against the wrong title.

Scenario: M was on title X (lmx 100), re-matched to title Y; Y is on the lineup as lmx 200 (M is On, or a sibling covers it). Nobody is matched to X any more, so the abandoned alert renders on M's Manage tab. Someone already deleted lmx 100 by hand on pinballmap.com, so `client.removeMachine(100)` returns `not_found`. This function refreshes, resolves `findLmxForMachine(snapshot, Y)` → `{id: 200}`, sees `200 !== 100`, and returns `{kind: "retry", lmxId: 200}` — line 681 then issues a real `removeMachine(200)`, taking title Y off the public lineup, and the timeline records `unlisted: 200` as if that were intended.

`e2e/full/machine-pinballmap-abandoned-notice.spec.ts` cannot catch this: its new title is a 900,000,000-range id that the seeded lineup never contains, so `live` is always null and the verdict is always `gone`.

Suggested fix: pass the *abandoned entry's* `pinballmapMachineId` (the `pinballmap_abandoned_listings` row already stores it), or `null` when `explicitLmxId !== null`, which makes the re-mint check unavailable and correctly falls through to `gone`/`refuse`.

src/app/(app)/m/pinballmap-actions.ts
```diff
@@ -719,31 +691,28 @@ export async function unlistMachineFromPinballMapAction(
...
     }
   }
   // --- transaction: local state only ---
 
-  // No 23505 catch: this write only CLEARS `pinballmapListed`, and a row
-  // leaving the partial unique index cannot violate it.
+  // Intent is deliberately untouched. Removing is how the operator's existing
+  // Off decision gets carried out; writing intent here would make the push and
+  // the toggle two ways to do one thing, which is the conflation the two-line
+  // control exists to undo (spec 4.1).
```
timothyfroehlich, line 780: **(about line 717)** — `withLmxRemoved(snapshot, deletedLmxId, machine.pinballmapMachineId)` drops the machine's current title from the stored lineup on the abandoned-entry path.

`withLmxRemoved` filters out every row whose `id` is `lmxId` **OR** whose `machineId` is `pinballmapMachineId` (`snapshot-edit.ts:81-85`). That OR is right for the Lingering path, where both predicates name the same row. It is wrong when the removal came from the abandoned-entry alert: `deletedLmxId` is the OLD title's entry, while `machine.pinballmapMachineId` is the machine's NEW title.

Scenario: M retitled X → Y; Y is on the lineup as lmx 200 (put there by M or a sibling). Remove the abandoned X entry — PBM deletes lmx 100, but the stored snapshot loses lmx 100 *and* lmx 200. Every same-title cabinet's control immediately re-derives to Missing and offers Add, and the Info tab reads "Not on Pinball Map", for up to an hour until the cron re-fetches. Pressing Add then spends a PBM write to be handed lmx 200 back.

Pass `null` for the title when the removal was driven by an explicit `lmxId`, which the function already documents as "narrows it to the `id` match alone".

src/app/(app)/m/pinballmap-actions.ts
```diff
@@ -623,30 +595,30 @@ export async function unlistMachineFromPinballMapAction(
...
-  // an hour stale and carry the same dead id as the machine row. `not_found` is
-  // where that residue is caught — see `classifyRemoveNotFound`.
+  // An explicit lmx wins: the abandoned-entry alert names an entry whose title
+  // this cabinet no longer carries, so resolving by the machine's CURRENT title
+  // would find the wrong row or none at all.
+  const explicitLmxRaw = formData.get("lmxId");
+  const explicitLmxId =
+    typeof explicitLmxRaw === "string" && /^\d+$/.test(explicitLmxRaw)
+      ? Number(explicitLmxRaw)
+      : null;
```
timothyfroehlich: **The submitted `lmxId` is never checked against this machine's abandoned listings.**

Any well-formed digit string is accepted and, a few lines later, handed straight to `client.removeMachine`. The only gate above it is `machines.pinballmap.push` **for the machine named in `machineId`** — and that permission is `member: "owner"`, so a member who owns exactly one cabinet passes it.

Scenario: a member owns MM. They post `machineId=<their MM>` + `lmxId=<the lmx of somebody else's Godzilla>` to `removeMachineFromPinballMapAction` and PinPoint deletes that entry from the location's public lineup using the shared operator account. Nothing in the timeline attributes it to the right machine either — the receipt lands on MM.

Worth constraining the explicit id to the rows `listAbandonedForMachine(machine.id)` returns, which is the only surface that legitimately sends it.

src/app/(app)/m/pinballmap-actions.ts
```diff
@@ -169,12 +176,12 @@ async function authorizeListingAction(
   });
   if (!machine)
     return { ok: false, result: err("NOT_FOUND", "Machine not found") };
-  if (machine.pinballmapMachineId === null)
+  if ((opts?.requireLink ?? true) && machine.pinballmapMachineId === null)
```
timothyfroehlich, line 182: **`requireLink` defaults to true, which dead-ends the abandoned-entry Remove button for an unlinked/uncataloged machine.**

`removeMachineFromPinballMapAction` takes the default, so it refuses any machine whose `pinballmapMachineId` is null — but the abandoned-entry alert deliberately resolves its entry from an explicit `lmxId`, not from the machine's current title, and one of the ways an abandonment gets recorded is by clearing the link entirely or switching to "Not on Pinball Map" (`resolveCore` records `abandoned` whenever `!linkUnchanged && stored.pinballmapIntent === "on"`, including when the new columns are unlinked).

Scenario: GDZ is linked to title 3416 with intent On and the entry is live. Its owner edits Details, picks "Not on Pinball Map", saves. An abandonment row is written, GDZ is now `pinballmap_machine_id = NULL, pinballmap_excluded = true`, and no other cabinet carries 3416, so `titlesStillInTheFleet` lets the alert through on the Manage tab. Clicking **Remove machine from Pinball Map** returns `VALIDATION: "Machine isn't linked to a Pinball Map title yet"` in `pbm-abandoned-error`, and the entry can only be cleaned up on pinballmap.com. The e2e spec only covers the retitle case, where the machine keeps a link.

Either pass `{ requireLink: false }` from the remove action (the `liveLmxId === null` check below already rejects the no-entry case), or gate on "has a link **or** an explicit lmxId".

src/app/(app)/m/[initials]/(tabs)/page.tsx
```diff
@@ -104,86 +88,111 @@ export default async function MachineInfoTab({
...
+    pinballmapExcluded: machine.pinballmapExcluded,
+    intent: machine.pinballmapIntent,
+    presenceStatus: machine.presenceStatus,
+    snapshot,
+    siblings: sameTitle,
+  });
+  const configIssue =
+    canDiagnose &&
+    (listingView.outOfSync ||
+      (await listAbandonedForMachine(machine.id)).length > 0);
```
Comment (timothyfroehlich) [inline src/app/(app)/m/[initials]/(tabs)/page.tsx]: **The Config-issue chip counts abandoned rows the Manage tab then filters out — the same "warning links to a page that says everything is fine" bug the PR description lists as fix #2/#4.**

The Manage tab only renders an abandonment once **no** cabinet is still matched to its title (`edit/page.tsx:206-212`, `titlesStillInTheFleet`). This chip has no such filter — a bare `listAbandonedForMachine(machine.id).length > 0`.

Scenario: GDZ and HD are both on title 3416 with intent On and the entry is live. GDZ is re-matched to a different title; `resolveCore` records the abandonment for 3416's lmx because GDZ's stored intent was On. HD is still matched to 3416, so the Manage tab correctly suppresses the alert (and `clearResolvedAbandonments` will drop the row on the next hourly pass, since HD covers the title). Meanwhile GDZ's Info tab shows ⚠ Config issue linking to `/m/GDZ/edit`, which renders a clean control and no alert. Same failure with a matched-but-Off sibling, where the row survives until someone acts.

Applying the same `titlesStillInTheFleet` filter here — or extracting it so both pages share one predicate — keeps the chip and its destination in agreement.

src/app/(app)/m/new/create-machine-form.tsx
```diff
@@ -310,7 +310,7 @@ export function CreateMachineForm({
 
         {/* Model — links the machine to its PinballMap catalog model/edition
             (bead B / PP-o355.2). */}
-        <PinballMapLinkField />
+        <PinballMapLinkField machineName={nameValue} />
```
timothyfroehlich, line 313: **(about the block at lines 355-380, just below)** — the create form's Pinball Map copy is now false, and its own comment says this PR is what should have fixed it.

The comment at line 358 still describes auto-link ("captures the listing on the next hourly sync, or on the next save from the edit form") — deleted by this PR — and line 365 says outright: *"the same sentence needs rewriting anyway when PP-o355.21 ships the six-state control it should point at ... the Manage tab's control is still a placeholder."* This is PP-o355.21 and the control is no longer a placeholder.

So a person creating a machine reads "Listing it there isn't automatic yet — machine-level listing controls are coming soon", then lands on a Manage tab that has exactly those controls. Worth updating the copy (and deleting the stale comment) in this PR rather than leaving a pointer to a shipped feature described as absent.

src/app/(app)/m/[initials]/(tabs)/edit/page.tsx
Comment (timothyfroehlich) [reply src/app/(app)/m/[initials]/(tabs)/edit/page.tsx]: Confirmed and fixed in 0e9b1bd2. The edit page now reads `modelName`/`manufacturer`/`year` from the row instead of through `getMachineForLayout`, so the panel opens populated and Save round-trips. Left the placeholder itself alone — it is still correct for the header, which has no source for `edition`/`backboxImageUrl` either, and widening it would change the header for every machine. Flagged that to Tim as its own bead. —Claude

src/app/(app)/m/pinballmap-actions.ts
timothyfroehlich: Confirmed and fixed in 0e9b1bd2. `classifyRemoveNotFound` now gets the abandoned entry's own title, which the abandonment record carries, rather than the cabinet's current one. Covered by a new test that removes an abandoned entry and asserts the cabinet's live entry is still on the lineup. —Claude

src/app/(app)/m/pinballmap-actions.ts
timothyfroehlich: Confirmed and fixed in 0e9b1bd2 — same one-line cause as the `classifyRemoveNotFound` thread, and both now take the abandoned entry's title. The new test asserts the stored snapshot still holds the cabinet's row afterwards, which is the half that would otherwise only show up as every same-title cabinet reading Missing until the next cron. —Claude

src/app/(app)/m/pinballmap-actions.ts
timothyfroehlich: Confirmed and fixed in 0e9b1bd2 — this was the most serious of the seven. The submitted lmx is now checked against this machine's abandonment records before anything else happens, and a foreign id is refused with NOT_FOUND. New test posts an lmx belonging to an unrelated entry and asserts the public lineup is untouched. —Claude

src/app/(app)/m/pinballmap-actions.ts
timothyfroehlich: Confirmed and fixed in 0e9b1bd2. `requireLink: false` is passed whenever the caller brought an explicit lmx — the requirement exists because resolving an entry BY TITLE needs a title, which does not apply when the entry names itself. New test removes an abandoned entry from a machine with no link at all. —Claude

src/app/(app)/m/[initials]/(tabs)/page.tsx
Comment (timothyfroehlich) [reply src/app/(app)/m/[initials]/(tabs)/page.tsx]: Confirmed and fixed in 0e9b1bd2, and you are right that the PR description overclaimed it. Rather than copy the filter to the second caller, it moved into `listSurfacingAbandonedForMachine` in `abandoned-listings.ts` and both the chip and the alert call it, so the two cannot diverge again. —Claude

src/app/(app)/m/new/create-machine-form.tsx
timothyfroehlich: Confirmed and fixed in 0e9b1bd2. Both halves were false — auto-link is deleted in this PR and the controls it called 'coming soon' ship in it. Rewritten to say the machine starts off the lineup and points at the Manage tab. Flagged the new wording to Tim, since user-facing copy is his call. —Claude

timothyfroehlich:
Claude review of head e1445dc — `/code-review medium` — Third review, 6 findings: MCP re-stating an exclusion nulled the hand-entered model identity (confirmed data loss, test fails without the fix); the CAS basis omitted the same 3 columns; Missing offered an Add the server always refuses (regression from round 2's guard); 2 bare anchors for internal nav; formatRelative at render in a client component. 5 fixed, 1 declined (token spend has no escalation to close — push holders are a subset of sync holders).

PR #1876, opened 2026-08-13
feat(mcp): set a machine's Pinball Map title (PP-u4ab.12)

timothyfroehlich:
Closes PP-u4ab.12 — the write half of the walk-the-floor flow, and the last thing blocking the 110-machine fleet linking pass (PP-h059).

## What this adds

**`set_machine_pinballmap(machine, pinballmapMachineId | pinballmapExcluded + pinballmapExcludedReason)`** — resolve a machine by initials or UUID, gate on `machines.pinballmap.link`, record which Pinball Map catalog title it is (or that it isn't one). Manufacturer, year, OPDB id and IPDB id are re-derived from the catalog mirror for whichever id lands; there is no argument that can write them. The response echoes `McpMachinePinballmap` — the same block `get_machine` returns — plus `previousPinballmap`, so a fleet pass can see exactly what each call changed.

Argument names match `add_machine` (`pinballmapExcluded` / `pinballmapExcludedReason`) rather than the bead's shorter `excluded` / `reason`: one model drives both tools in a linking session, and two spellings of one concept is a trap.

## What was extracted

The PinballMap half of `updateMachineAction` now lives in `~/services/machines.ts` as one seam both callers use:

- `planMachinePbmLink` — the listing carry-over, the abandoned-listing detection, and the auto-link decision. No transaction, no writes.
- `applyMachinePbmLink` — the columns UPDATE plus the abandonment record, in the **caller's** transaction, so the pairing cannot be half-adopted.
- `captureMachineAutoLink` — post-commit, best-effort (moved out of the action verbatim).
- `updateMachinePbmLink` — composes all three for the MCP tool, and reads the row back so an auto-link that lands is reported as `listed: true` rather than as the plan's `false`.

Three functions rather than one because the two callers have genuinely different transaction shapes — the edit form saves the link alongside name/owner/presence/description and must keep that atomic, the tool changes the link and nothing else. Collapsing them into one `db.transaction` would either fork the seam or make a machine save two transactions, and a half-saved edit is the regression this bead was warned about.

## Regression surface: the machine edit page

The edit page keeps its single transaction. The one behaviour change is that the link columns now travel as their own UPDATE inside that transaction instead of being spread into the details statement. `machines_owner_not_guest` is the only trigger on the table and it fires only `OF owner_id, invited_owner_id`, and `machines.updatedAt` has no `$onUpdate`, so the extra statement is inert.

That change did surface one real bug, caught by the existing `pinballmap-retitle-abandonment` integration tests: a save that carries **only** PBM fields (the picker's own submit posts no `name`) left the details statement with an empty `.set()`, which Drizzle rejects. The details UPDATE is now skipped when nothing outside the PBM block changed, and the row is read instead so the `NOT_FOUND` check still runs. Net statement count for a PBM-only save is unchanged from before.

## Tests

12 new integration cases in `src/test/integration/mcp-tools.test.ts` (PGlite, CORE-TEST-001; catalog and lineup seeded locally, nothing reaches pinballmap.com):

- links an unlinked machine, metadata derived from the catalog
- re-targeting a LISTED machine clears `pinballmapListed`/`pinballmapLmxId` and records the abandoned entry
- re-sending the SAME id preserves both
- excluded branch clears the link and its metadata without tripping `machines_pinballmap_link_exclusive`
- auto-link captures a listing already on the synced lineup, and the echo reports it
- a duplicate cabinet links without disturbing the title's incumbent lister
- permission denial (non-owner member) writes nothing; the owner is allowed
- a call naming neither a title nor exclusion is refused instead of unlinking
- both-at-once is refused; an id the mirror doesn't have is refused; unknown machine is `not_found`

## Where the bead's design was stale

- **The 23505 "one lister per title" catch is gone and should stay gone.** `@Claude-PbmListUnlist` flagged this on #1810 and it holds: the listing write moved behind `captureAutoLink`, which matches by constraint name and stands down. The carry-over re-writes a value the row already holds, so it cannot collide with itself. Re-absorbing a collision branch here would be dead code.
- **The acceptance criterion "linking to a title another cabinet already lists returns a readable 'already listed by X' error" describes behaviour that would now be wrong.** Duplicate cabinets of one title are ordinary; only one *lister* is forbidden. The new machine links quietly and stays unlisted. The test asserts that instead.
- **Auto-link is included on the MCP path** — not called out either way in the bead. Excluding it would fork the seam (UI links capture the listing, MCP links don't) for no gain: it reads the stored PBM snapshot, never a caller's claim, so it does not reopen PP-o355.29.

## Follow-up worth considering

The tool cannot clear a link back to "nothing recorded", so an accidental `pinballmapExcluded: true` can only be undone from the web edit form. Deliberate — the resolver's "neither" branch is right for a human emptying the picker and dangerous for a forgotten argument — but a future `clear: true` opt-in would close it honestly.

## Verification

- `pnpm run check` — green
- `pnpm run test` — 2107 unit tests green
- `pnpm run test:integration` (crabbox) — 647 green, 57 files
- `pnpm run build` (crabbox) — green

No migrations, no new env vars, no UI.

https://claude.ai/code/session_01ByHxCmTMweKpw9WphUnp3N

### DIFF
--- a/src/app/(app)/m/actions.ts
+++ b/src/app/(app)/m/actions.ts
@@ -11,9 +11,12 @@ import { after } from "next/server";
 import { createClient } from "~/lib/supabase/server";
 import { db } from "~/server/db";
 import {
+  applyMachinePbmLink,
+  captureMachineAutoLink,
   createMachine,
+  planMachinePbmLink,
   updateMachinePresence,
-  type MachinePbmColumns,
+  type MachinePbmLinkPlan,
 } from "~/services/machines";
 import {
   machines,
@@ -22,16 +25,7 @@ import {
   invitedUsers,
 } from "~/server/db/schema";
 import { createMachineSchema, updateMachineSchema } from "./schemas";
-import {
-  resolvePbmLinkColumnsForCreate,
-  resolvePbmLinkColumnsForUpdate,
-  type AbandonedListing,
-} from "~/lib/pinballmap/link-columns";
-import { recordAbandonedListing } from "~/lib/pinballmap/abandoned-listings";
-import {
-  captureAutoLink,
-  resolveAutoLinkForMachine,
-} from "~/lib/pinballmap/sync";
+import { resolvePbmLinkColumnsForCreate } from "~/lib/pinballmap/link-columns";
 import { type Result, ok, err } from "~/lib/result";
 import { z } from "zod";
 import { eq, and } from "drizzle-orm";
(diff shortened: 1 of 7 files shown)


timothyfroehlich:
Claude review of head 6398aac — `/code-review medium` — Reason-wipe on re-confirmed exclusion (carry-over + CAS basis), conflict mis-audited as invalid — both fixed in 6398aac7

PR #1877, opened 2026-08-13
feat(mcp): let list_machines filter on a set of presence values (PP-u4ab.13)

timothyfroehlich:
## What

`list_machines` took `presence` as a **single** value. That made one question unaskable — the exact question the PinballMap fleet linking pass (PP-h059) runs on:

> which machines still have no catalog match **and** are still plausibly in the collection?

`pinballmap: "unlinked"` on its own also returns the cabinets that are `removed` or `pending_arrival`. Nobody will ever link those, so they come back on every page of every sweep and hold `total` above zero permanently. On prod today that is **14 of the 110 unlinked machines** (10 `removed`, 4 other non-floor states) — 13% of the worklist is rows the model has to recognise and skip by hand, every page, every time.

`presence` now accepts **one status or an array of them**:

```jsonc
{ "presence": "off_the_floor" }                                     // unchanged
{ "presence": ["on_the_floor", "on_loan", "off_the_floor"] }        // new
```

## The parameter shape

A `z.union([z.enum(...), z.array(z.enum(...)).min(1)])`, mirroring the `statusFilterSchema` that `list_issues` already ships — which cites this bead by name as its reason for shipping the set form first rather than second.

Both forms normalise through `resolvePresence()` to a single `inArray(...)` condition. Not an `eq` path plus an `inArray` path: one condition is one thing to keep in step with the count query.

`.min(1)` rejects an empty array. An empty set would type-check, produce `inArray(col, [])` — a predicate matching no row — and hand back `total: 0` for the whole collection, which reads as an authoritative "there are none" rather than as the malformed filter it is (CORE-ARCH-012).

## Page and count share the condition

The `inArray` joins the **same** `conditions` array the tool already builds. `const where = and(...conditions)` is computed once and passed to both `db.query.machines.findMany` and the `count()` select, so they cannot diverge. That matters because the tool's own description tells the model to answer counting questions from `total`, never from `count`.

Verified by mutation in both directions, against the new tests:

| Mutation | Result |
| --- | --- |
| Count query drops the presence condition (`conditions.slice(1)`) | 4 tests fail — `expected 5 to be 1`, `expected 60 to be 36` |
| Page query drops it instead | the same 4 fail, on the row sets |

Both reverted; the diff contains neither.

## Description text

The tool description now recommends the narrowed worklist directly rather than relying on the model to notice un-actionable rows itself: ask for `pinballmap: "unlinked"` **together with** `presence: ["on_the_floor", "on_loan", "off_the_floor"]`, because narrowing presence drops those cabinets from `total` as well as from the page.

The existing paging/drain guidance is unchanged and now closes the loop — "you are done when a request returns EMPTY, not when total reaches 0" gains a sentence saying that narrowing the filter to rows you can actually act on is what lets `total` fall to 0 at all.

## Tests

Five cases added to `src/test/integration/mcp-tools.test.ts` (class I — DB query correctness — so integration/PGlite, alongside the PP-u4ab.9 link-state cases):

- the single-value form still means what it meant
- the array form returns exactly the listed states (set equality on the states, not a row count — an OR that leaked one state in while dropping another returns the same number of rows)
- `unlinked` alone vs `unlinked` + the presence set, with a linked on-the-floor cabinet present so the two filters can't produce the same answer by accident
- a 60-machine fleet paged at `limit: 7` with the set filter: `total` equals the matching set, no repeats, no gaps
- an empty presence array is rejected by the schema

The per-state seed helper is annotated `Record<MachinePresenceStatus, string>`, so adding a sixth presence state fails to compile here rather than quietly going unseeded.

## Scope

**MCP only.** `src/components/machines/MachineFilters.tsx` already has a multi-select presence filter defaulting to `["on_the_floor"]`; this is the tool catching up to the UI. No migration, no registration change, no service change. Read-only query change, so `pnpm run check` + `pnpm run test` per AGENTS.md §5, plus the full integration suite on the remote runner (640 passed).

Closes PP-u4ab.13.

### DIFF
--- a/src/lib/mcp/tools/list-machines.ts
+++ b/src/lib/mcp/tools/list-machines.ts
@@ -6,6 +6,7 @@ import {
   count,
   eq,
   ilike,
+  inArray,
   isNotNull,
   isNull,
   or,
@@ -14,7 +15,10 @@ import {
 import { z } from "zod";
 
 import { checkPermission } from "~/lib/permissions/helpers";
-import { VALID_MACHINE_PRESENCE_STATUSES } from "~/lib/machines/presence";
+import {
+  VALID_MACHINE_PRESENCE_STATUSES,
+  type MachinePresenceStatus,
+} from "~/lib/machines/presence";
 import { db } from "~/server/db";
 import { machines } from "~/server/db/schema";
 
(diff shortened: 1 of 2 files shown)


timothyfroehlich:
Claude review of head d9606be — `/code-review medium` — No findings — presence filter union, shared where-clause across page and count, and test isolation all verified

PR #1879, opened 2026-08-13
chore(huddle): move the huddle out of this repo into dotfiles (PP-o4sy)

timothyfroehlich:
Moves the inter-session coordination system (the huddle) out of PinPoint and into Tim's dotfiles, so editing it no longer costs a PR cycle.

## Why it moved cleanly

Nothing in the huddle was PinPoint-specific. The scripts source their siblings via `$(dirname "$0")` and resolve every piece of repo state through `git rev-parse` against the **cwd** — so they moved verbatim, with no code changes.

## What moved

Added in dotfiles (`timothyfroehlich/dotfiles` `33bb220`):

| From | To |
| --- | --- |
| `scripts/hooks/huddle-*.sh` (9) | `~/.claude/hooks/huddle/` |
| `scripts/tests/test_huddle_*.py` (7) | `~/.claude/hooks/huddle/tests/` |
| `.agents/skills/pinpoint-huddle/` | `~/.claude/skills/huddle/` (renamed `huddle`) |

## What stays: the channel

`.agents/huddle/`, the registered-name map, and the root/daily beads still resolve from the cwd's repo, so the conversation is per-project exactly as before. The four hook registrations in `.claude/settings.json` now point at `"$HOME"/.claude/hooks/huddle/`. A repo without those registrations simply never runs the hooks.

## Also repointed

- **The merge script** — its huddle-lib source. The pre-existing absent-file guard is now a real path: a machine without dotfiles stowed skips the post-merge notice instead of failing the merge.
- **`.agents/hooks/antigravity-bootstrap.cjs`** — both script paths, now `os.homedir()`-rooted.
- **`verify-guard-stack.cjs`** — it skips absolute paths by design, so those four registrations are no longer verified against disk. Documented in place. Its example strings and the matching test fixtures moved to a repo-relative hook that still exists.
- **`AGENTS.md` §3** — a pointer, since the huddle is now the one skill not under `.agents/skills/`.

## Verification

- `pnpm run check` — pass
- `pnpm run check:python` — 439 pass (the 136 huddle tests moved out)
- `verify-guard-stack.test.ts` — 31 pass
- In dotfiles: 136 huddle tests pass, shellcheck + ruff clean
- **Live**: today's huddle rotation ran `huddle-rotate.sh` from the new location against the real repo and rotated `PP-lt12.72` → `PP-lt12.73` correctly

## One consequence worth naming

The huddle tests left `check:python` and nothing runs them on a schedule now — dotfiles has no CI. The command is recorded in both the `huddle` and `dotfiles` skills.


timothyfroehlich:
Claude review of head 8ba42e1 — `/code-review medium` — Reviewed at medium (three passes). Findings were about the guard canary being blind to the out-of-repo $HOME registrations, not about the move itself; Tim's call was to keep this PR a clean move, so they are deferred to PP-z88x with the implementation preserved at fab46548/390e5526/b8072717. The 5 files those touched are now byte-identical to the reviewed pure-move state.

PR #1880, opened 2026-08-14
fix(hooks): stop normalize-workspace-paths corrupting remote /var/home paths (PP-xbfn)

timothyfroehlich:
## What

`.claude/hooks/normalize-workspace-paths.cjs` rewrites absolute workspace paths to relative so they match the `settings.json` allowlist. Its matcher hardcoded `/home/froeht/Code/PinPoint` and was unanchored. That broke crabbox from the Mac.

Bazzite's home is `/var/home/froeht` and `/home` is a symlink to it, so `/home/froeht/Code/PinPoint/` appears as a **substring in the middle** of any `/var/home/...` path. The hook stripped that span and left the orphaned `/var` glued to the tail:

```
CRABBOX_STATIC_WORK_ROOT=/var/home/.../PinPoint/.claude/worktrees/crabbox-runner
                      -> /var.claude/worktrees/crabbox-runner
```

crabbox then died on `mkdir: cannot create directory '/var.claude': Read-only file system`.

## Why it was hard to see

- The transcript records the command the **agent wrote** — the rewrite happens after — so the corruption was invisible until the remote `mkdir` failed. It read as a crabbox bug.
- An `existsSync` guard on the tail made it look intermittent: `.claude/worktrees` and `.git/config` triggered it, made-up tails did not.
- It fired twice during PP-cuf1 before being traced here.

## The fix

Anchoring the regex alone would not have been enough — a `/var/home/...` string on the Mac is a **remote** path, so relativizing it is wrong however it is spelled. The prefixes are now derived from this machine's own repo root, with three rules that keep a rewrite from changing what a command means:

- **only paths under the local repo root** — a Mac session no longer touches Bazzite paths at all, and the stale `pinpoint-worktrees/` layout drops out (a worktree is its own root)
- **only at a path boundary** — negative lookbehind, so a prefix can't match mid-path
- **only when cwd IS the repo root** — a root-relative path is equivalent to the absolute one nowhere else

A Bazzite session still gets the rewrite for both spellings of its own root; `workspacePrefixes` adds the `/home` ↔ `/var/home` sibling.

**Side effect worth naming:** the hook now does something on the Mac, where every path it could match was previously foreign.

## Testing

- 17 new unit tests (`src/test/unit/hooks/normalize-workspace-paths.test.ts`) covering both halves of the regression, the boundary cases, and worktree/main-repo separation
- End-to-end hook check on real stdin: the crabbox command passes through untouched; a local absolute script path is rewritten
- `pnpm run check` green, `pnpm run test` green (2147 passed)

Closes PP-xbfn.

https://claude.ai/code/session_01WCqraD9WEZMFCp5y5VPBhc

---

_Note: this branch briefly carried a nanoid lockfile bump to get past the red `pnpm audit` gate. Reverted — #1881 (PP-5m4t) fixes it properly by bumping the override floor in `pnpm-workspace.yaml`, which the lockfile-only change did not touch. This branch needs main merged in once #1881 lands._

### DIFF
--- a/.claude/hooks/normalize-workspace-paths.cjs
+++ b/.claude/hooks/normalize-workspace-paths.cjs
@@ -12,34 +12,88 @@
  * match, causing unnecessary permission prompts. This hook auto-fixes the
  * command and tells the agent to use relative paths next time.
  *
- * The prefixes it matches are derived from THIS machine's repo root, not
- * hardcoded. Three rules keep the rewrite from changing what a command means
- * (all three learned from PP-xbfn):
+ * ---------------------------------------------------------------------------
+ * What makes this hook dangerous, and the four guards that hold it back
+ * ---------------------------------------------------------------------------
  *
- * 1. **Only this machine's repo root.** The matcher used to hardcode
- *    `/home/froeht/Code/PinPoint`, which on the Mac names no local directory at
- *    all — it is Bazzite's path. So the hook rewrote paths that were meant for
- *    the *remote* box. A crabbox invocation carrying
- *    `CRABBOX_STATIC_WORK_ROOT=/var/home/froeht/Code/PinPoint/.claude/worktrees/...`
- *    is a remote path; relativizing it is wrong however it is spelled. Matching
- *    only the local root leaves it alone.
+ * An absolute path and a root-relative one only mean the same thing when the
+ * command runs from the root, on this machine. Every guard below exists because
+ * some case broke that assumption, and a broken rewrite is invisible: the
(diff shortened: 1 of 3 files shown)


timothyfroehlich:
Claude review of head b5e7a5a — `/code-review medium` — Round 2: 3 findings, all fixed — HIGH permissionDecision:allow bypassed permissions.ask on rm -rf/git reset --hard once the matcher fix made the hook reachable; 2 MEDIUM context-shift guard missed cd/ssh behind env-var assignments, sudo/env/time wrappers, quoted -c payloads and loop bodies

PR #1881, opened 2026-08-14
fix(deps): bump the nanoid override floor to 3.3.18 (PP-5m4t)

timothyfroehlich:
Unblocks the audit gate, which is currently red on **every open PR** (#1875–#1880) — none of which touched dependencies.

## The advisory

[GHSA-2v37-7h3g-55p8](https://github.com/advisories/GHSA-2v37-7h3g-55p8) (high), published 2026-08-13: nanoid custom generators can loop indefinitely when size is zero. Patched in 3.3.18.

We reach nanoid transitively and dev-only, through `webpack` → `postcss` → `nanoid` (21 paths).

## The fix

We already pin nanoid in `pnpm-workspace.yaml` overrides, so the floor just moves:

```diff
-  nanoid: '>=3.3.17 <4'
+  nanoid: '>=3.3.18 <4'
```

The lockfile delta is nanoid and nothing else — 6 lines, all `3.3.17` → `3.3.18`.

## Why not `/audit-override`

That hatch exists for an advisory with no available patch, where the alternative is an admin merge. This one has a patch, and we already own the pin — waving it through would leave a fixable high advisory in the tree and mask the next one behind the same override.

## Verification

- `pnpm audit --audit-level=high` — **no known vulnerabilities found** (was: 1 high)
- `pnpm run check` — pass

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -15,7 +15,7 @@ overrides:
   yaml: '>=2.8.3'
   uuid: '>=14.0.0'
   postcss: '>=8.5.23'
-  nanoid: '>=3.3.17 <4'
+  nanoid: '>=3.3.18 <4'
   shell-quote: '>=1.8.4'
   ws: '>=8.21.0'
   vite: '>=8.0.16'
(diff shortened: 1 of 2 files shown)


timothyfroehlich:
Claude review of head 353ec2d — `/code-review medium` — No findings. Verified: 2-line override bump (>=3.3.17 -> >=3.3.18, <4 ceiling preserved), lockfile delta is nanoid-only with no collateral churn, integrity hash matches the npm registry for 3.3.18 so the lockfile was pnpm-regenerated not hand-edited.

PR #1882, opened 2026-08-14
fix(ci): stop main merges from cancelling each other's CI runs (PP-tbhv)

timothyfroehlich:
The third way the post-merge comprehensive E2E job could report nothing — and the one that fires most often. Found by watching the first post-merge run after #1861 landed.

## What happens

`ci.yml` keys its concurrency group on `github.ref` alone with `cancel-in-progress: true`. On `refs/heads/main` every merge shares one group, so **each merge cancels the previous merge's in-flight run**.

`test-e2e-comprehensive` takes ~30 minutes. Main routinely gets several merges inside that window, so the job was cancelled far more often than it finished.

## The evidence

Run [31755670441](https://github.com/timothyfroehlich/PinPoint/actions/runs/31755670441), the merge of #1873:

| | |
| :--- | :--- |
| job started | 00:01:42 |
| `Run Comprehensive Smoke Tests` | success |
| `Evaluate gating browser results (Smoke)` | **success** — the #1861 gate working |
| `Run Comprehensive Full Tests` | **cancelled** at 00:31:31 |
| `Evaluate gating browser results (Full)` | skipped |
| `Gate on E2E verdicts` | skipped |

Two further merges landed at 00:30:55 and 00:31:35 — the pbm docs PR and the nanoid bump. Those cancelled it. The job cap is now 90 minutes, so this was nowhere near the timeout #1861 fixed.

Net result: no verdict, which is exactly the class PP-jxhy set out to remove, reached by a different route.

## The fix

```yaml
group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name == 'push' && github.sha || '' }}
```

Every main commit gets its own group, so main runs neither cancel nor queue behind each other. Pull-request runs keep the ref-based group and cancel on force-push exactly as before — which is the only case the setting was ever for. A PR push supersedes the one before it; a merge to main does not.

## Alternatives rejected

- **Job-level `concurrency` on `test-e2e-comprehensive`.** Job-level groups govern queueing between jobs; they do not exempt a job from a workflow-level cancel.
- **`cancel-in-progress: ${{ github.event_name == 'pull_request' }}`.** Main runs would then serialize, so one 30-minute comprehensive job delays the next main run by up to 30 minutes and they pile up.
- **Splitting the job into its own workflow.** It `needs:` nine upstream jobs; a separate workflow would have to re-derive all of that or chain on `workflow_run`.

Both rejections are recorded in the comment above the block, so the next person to look at it doesn't re-derive them.

## Verification

Regression pin added beside the other `ci.yml` assertions in `test_evaluate_e2e_results.py` — it asserts both halves, that pushes are SHA-scoped and that PR runs still cancel. `pnpm run check` and all 575 python tests pass.

The real proof is behavioral and arrives after merge: two merges to main inside 30 minutes should both reach a real conclusion.

Bead: PP-tbhv.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,9 +11,31 @@ on:
 
 permissions: read-all  # zizmor: ignore[excessive-permissions]
 
-# Cancel in-progress runs when a new run is triggered
+# Cancel superseded PR runs; never cancel a push to main.
+#
+# A PR push supersedes the one before it — same branch, same question, and the
+# older answer is worthless — so cancelling saves runners with nothing lost.
+# A push to main is the opposite: each one is a distinct commit whose result is
+# the thing we want, and `test-e2e-comprehensive` takes ~30 minutes to produce
+# it. Under a shared `${{ github.ref }}` group every merge cancelled the
+# previous merge's run, so on any evening with more than one merge per half
+# hour that job was cancelled far more often than it finished — leaving the
+# same silent non-verdict PP-jxhy set out to remove, reached by a different
+# route. (Observed on run 31755670441: Smoke evaluated clean, then the Full
+# step was cancelled 30 minutes in by two later merges, and both the Full
+# evaluate and the gating step skipped.)
+#
+# Appending the SHA on `push` gives every main commit its own group, so main
+# runs neither cancel nor queue behind each other. Pull-request runs keep the
+# ref-based group and the old behaviour exactly.
+#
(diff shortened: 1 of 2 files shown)


timothyfroehlich:
Claude review of head 0df198d — `/code-review medium` — One low finding: the regression test's cancel-in-progress assertion was a whole-file substring check, satisfiable by a job-level concurrency block; rescoped to the workflow-level block and verified by mutation. Expression semantics, trigger set, shared-state races, and ci-gate needs all verified clean.

PR #1883, opened 2026-08-14
fix(auth): derive a name for OAuth signups, and forbid a blank one (PP-if48)

timothyfroehlich:
Phil could not find Paul Muntner in the machine-owner picker to hand him The Big Lebowski. Paul had an account, but no first or last name — so no search matched him. Reported in Discord 2026-08-12.

## Root cause

`handle_new_user` read `raw_user_meta_data->>'first_name'` / `'last_name'` and defaulted both to `''`. Our own signup form supplies those keys; **no OAuth provider does.** Measured on prod:

| provider | users | blank names |
|---|---|---|
| email | 25 | 0 |
| discord + email | 6 | 0 |
| **discord only** | **4** | **3** |

Discord sends the account handle as `full_name` (`pmuntner`) and the human display name nested under `custom_claims.global_name` — which held `"Paul Muntner"` the entire time.

## Why it was invisible until it wasn't

`user_profiles.name` is `GENERATED ALWAYS AS (first_name || ' ' || last_name)`. A blank user's name was **a single space** — not NULL, not `''`. So every `?? "Anonymous"` / `?? "Unknown"` fallback in the codebase stepped straight over it, and every owner/assignee picker (all of which filter on `name`) could match him only if you typed a literal space. Nothing errored; he was just absent.

## The fix

- **One fallback chain, two implementations that must agree**: `derive_profile_name()` in SQL and `deriveName()` in TS. Explicit first/last → `custom_claims.global_name` → `full_name` → handle minus discriminator → email local-part. The floor is never empty, which is what makes the new check un-trippable.
- **Both writers of `user_profiles` use it** — the trigger and `ensureUserProfile`. `supabase/seed.sql` re-defines the trigger on every local reset, so it is updated in step; otherwise `db:reset` would quietly undo this locally while prod stayed correct.
- **`btrim(first_name) <> ''`** on `user_profiles` and `invited_users`. NOT NULL never forbade `''`, which is why this was representable.
- **Only a first name is required.** Last name is optional everywhere now — plenty of people go by one name, and every OAuth fallback yields a single token. Flagging this as the one product call in here.
- **`name` columns wrapped in `btrim`**, so a user with no surname stops carrying a trailing space.

## The sharp edge

Rebuilding a generated column means dropping it, and **`public_profiles_view` depends on it**. I initially asserted no view did, and the migration failed on the runner proving me wrong. The view is now dropped and recreated verbatim, with `security_invoker` (0034) and its SELECT grant (0018) re-applied. 0018 exists *because* 0016 recreated this same view and silently dropped the grant, taking email privacy with it — `DROP ... CASCADE` would have repeated that with no trace.

## Verification

On the crabbox runner:
- full `pnpm run db:reset` green, seeds included (users are created through the trigger)
- all **20 cases** in `scripts/sql/verify-derive-name.sql` agree with `derive-name.test.ts`
- **11/11** email-privacy RLS tests pass against the rebuilt view

Locally: `pnpm run check` clean, **2155/2155** unit tests pass.

## Prod data

Already repaired by hand before this landed, so the backfill here is a no-op on prod and not a no-op anywhere else:

| user | name | role |
|---|---|---|
| pmuntner@yahoo.com | Paul Muntner | member |
| keyyek123@gmail.com | Furgers | guest |
| nickpereira.np@gmail.com | PresidentNick | guest |

The two guests were set to their Discord handles per Tim's call — with a follow-up to prompt them for a real name on next login (not in this PR; it can recompute "is this a placeholder?" from `raw_user_meta_data`, so it needs no column and no backfill).

## Deliberately not here

Investigating this meant reading prod, which over a service-role connection makes every read one typo from a mutation — so a read-only role and a query tool got built along the way. They are their own change and were holding up a production fix, so they now live in **PP-xdvw** instead. Recover them from `a946675a`, retained on this PR.

https://claude.ai/code/session_015117bLDEbXM7EqHTkYoVFM


timothyfroehlich:
Claude review of head 72efac6 — `/code-review medium` — 7 findings, all fixed: MCP owner-by-name still matched the un-btrimmed concat and missed single-name users (the bug this PR fixes); readonly-role.sql granted schema-wide SELECT on auth (plaintext Discord OAuth tokens, OAuth client secrets, TOTP seeds) and has since been pulled out to PP-xdvw; SQL/TS discriminator strip-order divergence; deriveName could return an empty first name; ensureUserProfile ignored an invited user's name; two more un-btrimmed concatenations.

PR #1885, opened 2026-08-14
docs(skills): add pinpoint-apc-records for the APC Google Drive sources (PP-d012)

timothyfroehlich:
## What

Adds `.agents/skills/pinpoint-apc-records/SKILL.md`.

APC's source of truth for **who owns which game** is a Google Sheet, not PinPoint — PinPoint is downstream. Reconciling the two on 2026-08-13 cost a full session of trial and error, and none of what it cost was reusable without writing it down.

## What the skill carries

- **File IDs** for both sheets (Games and Maintenance, Official Membership Database) and their parent folders, so a reconciliation pass starts with a `navigate`, not a Drive hunt.
- **The same-origin `gviz` CSV fetch.** Sheets renders to canvas, so `get_page_text` returns nothing and reading the grid visually costs a screenshot per ~20 rows. The CSV export is one call — but it is same-origin-only, so fetching it from a `drive.google.com` tab fails CORS with a bare `TypeError: Failed to fetch`. Navigate to the spreadsheet first.
- **A PII rule**: filter inside the page and return only the columns needed. Both sheets carry email and phone; pulling whole rows drags member PII into the transcript for nothing.
- **Three name-matching traps**, each of which produced a confident wrong answer: the two sheets spell the same person differently (`Steve Jones` / `Stephen Jones`); a Discord report is a claim, not a record (a floor-walk post was outvoted by the sheet plus PinPoint); and APC membership is not a PinPoint account (`set_machine_owner` fails for a member with no account, and there is no MCP tool to invite one).
- **The PinPoint MCP surface** and the presence vocabulary, including that `removed` vs `on_loan` is Tim's judgment call, not a guess.

## Testing

`pnpm run check` — exit 0. Docs-only change, so per AGENTS.md §5 tier 1 that is the whole gate.


timothyfroehlich:
Moving this skill out of the repo — it's Tim's personal Google Drive access, not PinPoint project context. Lives in his dotfiles now.

PR #1886, opened 2026-08-14
chore(hooks): disable normalize-workspace-paths and see who misses it (PP-yv19)

timothyfroehlich:
## What

Unregisters `.claude/hooks/normalize-workspace-paths.cjs` from `.claude/settings.json`. The file and its 45 tests stay on disk.

## What the hook did

Rewrote absolute paths pointing inside the current repo into relative ones, **in the command text**, before the command ran:

```
bash /Users/froeht/Code/PinPoint/scripts/workflow/merge-handoff.sh 1880
                      ↓
bash scripts/workflow/merge-handoff.sh 1880
```

The point was that the allowlist is written in relative paths, so the absolute spelling matched no entry and produced a permission prompt for an already-approved command.

## Why turn it off

That benefit — fewer prompts when an agent spells a repo path absolutely — cost the following across three review rounds on #1880:

| | |
|---|---|
| Silent path corruption | `/var/home/…/PinPoint/.claude/worktrees` → `/var.claude/worktrees`; broke crabbox twice before it was traced |
| Command mis-targeting | `cd /tmp && bash <root>/x.sh` → `cd /tmp && bash scripts/x.sh`, i.e. `/tmp/scripts/x.sh` |
| Command mis-targeting | `ssh bazzite 'ls <root>/scripts'` relativized on Bazzite, where local and remote roots are spelled alike |
| Permission bypass | `permissionDecision: "allow"` skipped the `permissions.ask` prompt on `rm -rf`, `git reset --hard`, `git clean` |

Each was fixed, and each was found only because someone went looking. Rewriting arbitrary shell text to save a prompt is the wrong trade.

## The prompt doesn't need a hook

Bash permission rules match the **literal command string** with `*` wildcards and no path resolution ([permissions docs](https://code.claude.com/docs/en/permissions)). So an absolute-path allow rule covers the case directly, with no rewriting. The machine-local settings already carry one for `scripts/workflow/`, with a wildcard right after the repo root so it covers worktrees too — which is why the hook's benefit was already largely redundant on this machine.

If a prompt does start showing up, the fix is one allowlist line for the case that actually came up.

## Disabled, not deleted

Re-enabling is one settings entry plus the three expected-hook list entries removed here (`verify-guard-stack.cjs`, its unit test, `test_node_hook_paths.py`). Those lists had to lose it either way — a hook listed as expected but not registered makes the guard-stack canary warn every session. Delete the lot if nothing misses it.

The file header now states that it is disabled and why, so nobody reads it as live.

## Testing

`pnpm run check`, `pnpm run check:python` (573 passed), and `src/test/unit/hooks/` (286 passed) all green.

Closes PP-yv19.

https://claude.ai/code/session_01WCqraD9WEZMFCp5y5VPBhc

### DIFF
--- a/.claude/hooks/normalize-workspace-paths.cjs
+++ b/.claude/hooks/normalize-workspace-paths.cjs
@@ -1,5 +1,29 @@
 #!/usr/bin/env node
 /**
+ * DISABLED — not registered in .claude/settings.json since 2026-08-14 (PP-yv19).
+ *
+ * Kept on disk, with its tests, because re-enabling is one settings entry plus
+ * three expected-hook list entries (verify-guard-stack.cjs, its unit test,
+ * scripts/tests/test_node_hook_paths.py). Delete all of it if nothing misses it.
+ *
+ * Why it was turned off: what it buys is fewer permission prompts when an agent
+ * spells a repo path absolutely. What it cost, over three review rounds on
+ * PR #1880, was a silent path corruption that broke crabbox twice, two
+ * command mis-targeting bugs, and a `permissionDecision: "allow"` that skipped
+ * the `permissions.ask` prompt on `rm -rf` and `git reset --hard`. Rewriting
+ * arbitrary shell text is a poor trade for that.
+ *
+ * The prompt it dodged is also solvable without rewriting anything: Bash
+ * permission rules match the literal command string with `*` wildcards, so an
+ * absolute-path allow rule covers the case directly. The machine-local settings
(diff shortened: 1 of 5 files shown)


timothyfroehlich:
Claude review of head f633e82 — trivial change, no /code-review run — Self-attested at Tim's explicit instruction. Config-only: unregisters one PreToolUse hook and drops its name from three expected-hook lists. No behavior change beyond the hook no longer running; the hook file and its tests are untouched apart from a header note.

PR #1888, opened 2026-08-14
feat(scripts): read-only Postgres role and query tool for prod investigations (PP-xdvw)

timothyfroehlich:
Agents investigating a production bug need to read production, including `auth.users`. Over the service-role string, every one of those reads carries write authority, and the only thing between a typo and a mutation is the agent's own judgement. You asked for a mechanism instead of restraint (mid-PP-if48). This is it.

Built and reviewed during #1883, then pulled out of it so that PR stayed the user-name fix. Restored from `a946675a` — and **running it turned up two bugs that the review could not have caught, one of which made the whole thing inert.**

## The auth grant silently did nothing

The version reviewed in #1883 had:

```sql
GRANT USAGE ON SCHEMA auth TO pinpoint_readonly;
GRANT SELECT (id, email, raw_user_meta_data, …) ON auth.users TO pinpoint_readonly;
```

Schema `auth` is owned by `supabase_admin`, and the `postgres` role you run this as holds USAGE on it **without grant option** — `postgres=U/supabase_admin` in `pg_namespace.nspacl`. Postgres answers that with a **WARNING, not an error**:

```
WARNING:  no privileges were granted for "auth"
```

`psql` exits 0. The column grants beneath it land in the catalog and are unreachable. The result is a role that reports successful setup and then gets `permission denied for schema auth` on every auth table — unable to read the one thing it exists to read.

There is no version of that statement that works: Supabase does not give `postgres` superuser ([docs](https://supabase.com/docs/guides/database/postgres/roles-superuser)), and `supabase_admin` is not connectable on a hosted project.

**Auth is now reached through owner-rights views** in a `readonly_auth` schema. A view runs with its owner's privileges unless created `security_invoker`, and `postgres` does hold USAGE on `auth` — so the view reads what its caller cannot. Not `security_invoker` here is the mechanism, not an oversight, and there is a check asserting it stays that way.

This is also a **tighter** allowlist than the 27 column grants it replaces: the view names its columns and nothing re-derives them, so a future GoTrue migration adding another secret column cannot widen it. `auth.flow_state` (plaintext Discord OAuth access + refresh tokens), `auth.custom_oauth_providers.client_secret`, `auth.refresh_tokens`, `auth.mfa_factors` and `auth.sessions` are absent entirely.

## `--env-file` was unreachable

Node v20+ implements `--env-file` itself and consumes it **even after the script path**. A bad path died with Node's own error and exit 9 rather than the script's message; a good path had Node load *every* key in the file into the environment, which is the opposite of the two-key whitelist the script applies. Renamed to `--env`, with a tripwire test.

## Verified against a live stack, not read

| | |
|---|---|
| apply → verify → probe → re-apply | clean, idempotent, zero warnings |
| reads | `public.*`, `readonly_auth.users`, `readonly_auth.identities` |
| denied | `auth.users`, `auth.flow_state`, `auth.refresh_tokens`, `vault` |
| writes with `SET TRANSACTION READ WRITE` | `permission denied for table user_profiles` |
| a write through the tool | SQLSTATE 25006 refusal, exit 1 |

That fourth row is the one worth noting: the write fails on **privilege** even with the read-only default explicitly overridden, so the two layers are genuinely independent rather than one of them wearing a disguise.

`pnpm run check` clean; 2223/2223 unit tests.

## `verify-readonly-role.sql` is new

15 assertions read from the catalog — `has_schema_privilege` / `has_table_privilege` — rather than trusting that a GRANT was issued. That distinction is the entire reason the first bug survived a review, so the check that matters most is the one asserting direct `auth` USAGE is **absent**. It also fails on any credential-shaped column name reaching the views, which is a guard against a future GoTrue column being copied in without thought.

## What you still have to do

The role does not exist on prod. One-time, human-run, deliberately not a migration (a login role is an operator concern, and `CREATE ROLE` in migration history re-runs on every `db:reset`):

```
psql "$POSTGRES_URL_ADMIN" -v pw="$(openssl rand -base64 24)" -f scripts/sql/readonly-role.sql
psql "$POSTGRES_URL_ADMIN" -f scripts/sql/verify-readonly-role.sql
```

Then put this in `.env.local` as `POSTGRES_URL_READONLY` — **not** in the Vercel registry, the app never reads it:

```
postgres://pinpoint_readonly.<project-ref>:<pw>@<host>:6543/postgres
```

The `.<project-ref>` suffix on the username is not optional; Supavisor routes by tenant and reads it there, so the bare role name fails authentication rather than failing to route.

https://claude.ai/code/session_01Y7NuN3hQycxcRfKDmCXJHh

### DIFF
--- a/scripts/query-readonly.mjs
+++ b/scripts/query-readonly.mjs
@@ -1,0 +1,184 @@
+#!/usr/bin/env node
+/**
+ * Run a query against a database with writes made impossible.
+ *
+ * Two independent layers, either of which is sufficient on its own:
+ *
+ *  1. **A read-only transaction.** Every statement runs inside
+ *     `BEGIN TRANSACTION READ ONLY`, which Postgres enforces server-side: any
+ *     INSERT/UPDATE/DELETE/DDL fails with 25006 no matter what the connecting
+ *     role is permitted to do. This works even against the ordinary
+ *     POSTGRES_URL, which is why the script is useful before anyone has set up
+ *     the role.
+ *  2. **A SELECT-only role.** `POSTGRES_URL_READONLY` should point at
+ *     `pinpoint_readonly` (see scripts/sql/readonly-role.sql), which holds no
+ *     write grants anywhere.
+ *
+ * The transaction is always rolled back, so even a read that acquires locks
+ * leaves nothing behind.
+ *
+ * This exists because investigating a production bug means reading production,
+ * and doing that over a service-role connection makes every read one typo away
+ * from a mutation. Removing the capability beats asking for restraint.
+ *
+ * Usage:
+ *   node scripts/query-readonly.mjs "select count(*) from user_profiles"
+ *   node scripts/query-readonly.mjs --file query.sql
+ *   node scripts/query-readonly.mjs --json "select ..."
+ *   node scripts/query-readonly.mjs --env other.env "select ..."
+ *
+ * Reading auth data: the role cannot select from `auth.users` directly — query
+ * `readonly_auth.users` / `readonly_auth.identities` instead. The reason is a
+ * Supabase platform constraint, spelled out in scripts/sql/readonly-role.sql.
+ *
+ * The env-file flag is `--env`, NOT `--env-file`: Node itself owns `--env-file`
+ * (v20+) and consumes it even when it appears after the script path, so a script
+ * flag by that name is unreachable. Worse, Node's version loads EVERY key in the
+ * file into the environment, which is the opposite of the whitelist below.
+ */
+
+import { readFileSync } from "node:fs";
+
+import postgres from "postgres";
+
+import {
+  describeTarget,
+  isPinPointProductionTarget,
+} from "./lib/db-target.mjs";
+
+function parseArgs(argv) {
+  const args = { json: false, file: null, sql: null, envFile: null };
+  for (let i = 0; i < argv.length; i++) {
+    const arg = argv[i];
+    if (arg === "--json") args.json = true;
+    else if (arg === "--file") args.file = argv[++i] ?? null;
+    // `--env`, not `--env-file` — see the header. Node would eat the latter.
+    else if (arg === "--env") args.envFile = argv[++i] ?? null;
+    else if (!arg.startsWith("--")) args.sql = arg;
+  }
+  return args;
+}
+
+/**
+ * Read POSTGRES_URL* out of a dotenv file into the environment.
+ *
+ * Exists so a connection string never has to be pasted onto a command line,
+ * where it lands in shell history and in this tool's own logs. Only the two
+ * keys this script uses are read; everything else in the file is ignored.
+ */
+function loadEnvFile(path, { required = true } = {}) {
+  let contents;
+  try {
+    contents = readFileSync(path, "utf8");
+  } catch (error) {
+    // A missing default file is not an error — the caller may have exported the
+    // variables instead. An explicitly named one that cannot be read is.
+    if (!required) return;
+    console.error(`❌ Cannot read env file ${path}: ${error.message}`);
+    process.exit(1);
+  }
+  for (const line of contents.split("\n")) {
+    const match = /^\s*(POSTGRES_URL(?:_READONLY)?)\s*=\s*(.*)$/.exec(line);
+    if (!match) continue;
+    const value = match[2].trim().replace(/^["']|["']$/g, "");
+    if (value) process.env[match[1]] = value;
+  }
+}
+
+function resolveUrl() {
+  const readonly = process.env.POSTGRES_URL_READONLY;
+  if (readonly) return { url: readonly, dedicated: true };
+
+  const fallback = process.env.POSTGRES_URL;
+  if (!fallback) {
+    console.error(
+      "❌ Neither POSTGRES_URL_READONLY nor POSTGRES_URL is defined.\n" +
+        "   Set up the read-only role: scripts/sql/readonly-role.sql"
+    );
+    process.exit(1);
+  }
+  return { url: fallback, dedicated: false };
+}
+
+const args = parseArgs(process.argv.slice(2));
+// `.env.local` by default, so the documented invocations work in a checkout
+// without anyone having to remember `--env`; absent, it is skipped rather
+// than fatal. Values in the file override exported ones, which fails in the safe
+// direction — `.env.local` is the local stack, so the accident it can cause is
+// querying local while you meant prod, never the reverse.
+loadEnvFile(args.envFile ?? ".env.local", { required: args.envFile !== null });
+const sql = args.file ? readFileSync(args.file, "utf8") : args.sql;
+
+if (!sql?.trim()) {
+  console.error(
+    "❌ No query given.\n" +
+      '   node scripts/query-readonly.mjs "select 1"\n' +
+      "   node scripts/query-readonly.mjs --file query.sql"
+  );
+  process.exit(1);
+}
+
+const { url, dedicated } = resolveUrl();
+
+// Say plainly which database this is and how it is protected. A read against
+// prod is legitimate; a read against prod that the operator THOUGHT was local
+// is how bad assumptions get made, so the target is never implicit.
+const target = isPinPointProductionTarget(url)
+  ? "PRODUCTION"
+  : "non-production";
+console.error(`→ ${target}: ${describeTarget(url)}`);
+console.error(
+  dedicated
+    ? "→ read-only role + read-only transaction"
+    : "→ read-only transaction only (POSTGRES_URL_READONLY not set — the " +
+        "connecting role CAN write; the transaction is what is stopping it)"
+);
+
+// `prepare: false` — the transaction pooler on :6543 does not support prepared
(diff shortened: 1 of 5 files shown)


scripts/query-readonly.mjs
```diff
@@ -0,0 +1,267 @@
...
+    // A missing default file is not an error — the caller may have exported the
+    // variables instead. An explicitly named one that cannot be read is.
+    if (!required) return;
+    throw new UsageError(`Cannot read env file ${path}: ${error.message}`);
+  }
+  for (const line of contents.split("\n")) {
+    const match = /^\s*(POSTGRES_URL(?:_READONLY)?)\s*=\s*(.*)$/.exec(line);
+    if (!match) continue;
+    const value = match[2].trim().replace(/^["']|["']$/g, "");
+    if (value) process.env[match[1]] = value;
```
timothyfroehlich: loadEnvFile overrides already-exported env vars with the default `.env.local` (`process.env[match[1]] = value` runs unconditionally). This is the opposite of dotenv precedence, where existing process.env wins. Concrete failure: an operator does `export POSTGRES_URL_READONLY=<prod readonly>` and runs from a worktree whose generated `.env.local` also defines `POSTGRES_URL_READONLY`/`POSTGRES_URL` → the file value silently replaces the export, and the investigation runs against **local** while the operator believes it is prod, producing confidently-wrong "prod" results. The header comment defends this as "fails in the safe direction," but that argument only covers write-safety; for a read-only investigation tool the failure that matters is *which DB the answer came from*. Prefer dotenv default precedence (only set when the key is unset), so an explicit export wins over the default file.

scripts/query-readonly.mjs
```diff
@@ -0,0 +1,267 @@
...
+
+  let exitCode = 0;
+  try {
+    const connection = await client.reserve();
+    try {
+      // Session-level read-only extends the transaction's protection across a
+      // `ROLLBACK; …` escape (see the header): autocommit statements after such
+      // a break still inherit it. Not a wall — `SET … = off` defeats it — but it
+      // closes the accidental case on the fallback path.
+      await connection.unsafe("SET SESSION default_transaction_read_only = on");
```
timothyfroehlich, line 235: Over the Supavisor **transaction** pooler (`:6543`, which is what `createScriptClient` connects to) a session-level `SET` is not reliably session-scoped: transaction-mode pooling assigns a server backend per transaction, so this `SET SESSION default_transaction_read_only = on` (its own implicit txn) can land on a different backend than the subsequent statements. The explicit `BEGIN TRANSACTION READ ONLY … ROLLBACK` is one transaction → one backend, so it is fine — but the header claim that this session default "still catches" the `ROLLBACK; <write>` autocommit escape on the FALLBACK path is unreliable, because that escaped write autocommits in a fresh transaction that may be routed to a backend without the SET. For the dedicated role the role-level `ALTER ROLE … SET default_transaction_read_only = on` does cover every backend at auth time, so only the fallback-path guarantee described in the header is weaker than stated. Worth softening the header rather than a code change.

scripts/sql/readonly-role.sql
```diff
@@ -0,0 +1,217 @@
...
+
+-- Set/rotate the password. Kept out of the DO block so the value is never
+-- interpolated into a string that could end up in pg_stat_statements.
+ALTER ROLE pinpoint_readonly WITH LOGIN PASSWORD :'pw';
+
+-- BYPASSRLS is what makes the role useful: without it, RLS policies written for
+-- end users hide most rows, and an investigation returns a confidently wrong
+-- "no rows" rather than an error. It grants visibility, never mutation — the
+-- role holds no INSERT/UPDATE/DELETE anywhere, and Supabase forbids SUPERUSER.
+ALTER ROLE pinpoint_readonly WITH BYPASSRLS NOCREATEDB NOCREATEROLE NOINHERIT;
```
timothyfroehlich, line 81: Setting `BYPASSRLS` on another role is version-gated on who may do it. On PostgreSQL 15 and earlier, **only a superuser** can set `BYPASSRLS`; Supabase’s `postgres` role is deliberately not a superuser (the auth-section comment even relies on this). On PG16+ a role that itself has `BYPASSRLS` **and** `CREATEROLE` may pass it on, which `postgres` does — so this works on PG16/17 but will fail on a PG15 project with `ERROR: must be superuser to change bypassrls attribute`. Because `ON_ERROR_STOP` is on, that aborts the whole setup (loud, not silent), leaving no role — but the documented one-liner then just doesn’t work, and `BYPASSRLS` is the role’s core capability. Worth a note pinning the minimum Supabase Postgres major version (or confirming prod is ≥16).

scripts/query-readonly.mjs
```diff
@@ -0,0 +1,267 @@
...
+  return { url: fallback, dedicated: false };
+}
+
+/**
+ * Read the query text for `--file`, turning a bad path into a friendly error
+ * rather than a raw ENOENT stack trace (every other input error here exits 1
+ * with a `❌` line, and `--file` should match).
+ */
+function readQuery(args) {
+  if (!args.file) return args.sql;
```
timothyfroehlich, line 163: Minor consistency gap with the otherwise strict parser: if both `--file q.sql` and a positional query are given, `readQuery` returns the file contents and silently ignores the positional SQL. The parser is loud about a *second* positional (throws "unexpected extra argument") precisely to avoid running a statement the operator did not intend — but `--file` + positional hits the same "which statement actually ran?" ambiguity and stays silent. Consider erroring when `args.file` and `args.sql` are both set.

scripts/query-readonly.mjs
```diff
@@ -0,0 +1,267 @@
...
+  try {
+    contents = readFileSync(path, "utf8");
+  } catch (error) {
+    // A missing default file is not an error — the caller may have exported the
+    // variables instead. An explicitly named one that cannot be read is.
+    if (!required) return;
+    throw new UsageError(`Cannot read env file ${path}: ${error.message}`);
+  }
+  for (const line of contents.split("\n")) {
+    const match = /^\s*(POSTGRES_URL(?:_READONLY)?)\s*=\s*(.*)$/.exec(line);
```
timothyfroehlich, line 130: Low: this mini-parser strips surrounding quotes but does not handle a trailing inline `# comment` or an `export ` prefix. A `.env.local` line like `POSTGRES_URL_READONLY=postgres://… # prod ro` would put ` # prod ro` into the connection string, producing a malformed URL that fails at connect time. Real dotenv files for this key are unlikely to carry inline comments, and the failure is loud, so this is a robustness nit rather than a bug — noting it since the value is a connection string where a stray char changes the target.

scripts/query-readonly.mjs
timothyfroehlich: Fixed in e19f7c26. `loadEnvFile` now sets a key only when the environment hasn't already exported it (standard dotenv precedence), so an explicit `export POSTGRES_URL_READONLY=<prod>` wins over a worktree's `.env.local` rather than being silently swapped for the local stack. Added a test proving the export beats a conflicting `--env` file. —Claude

scripts/query-readonly.mjs
timothyfroehlich: Fixed (wording) in e19f7c26. The header and the inline comment now call the session-level SET best-effort and spell out that over the :6543 transaction pooler the escaped autocommit statement can route to a different backend than the one the SET ran on, so it is unreliable — the dedicated role's write-refusal, which depends on none of this, is named as the actual boundary. No code change to the SET; it still helps on a direct connection and is harmless on the dedicated role. —Claude

scripts/sql/readonly-role.sql
timothyfroehlich: N/A to our target, but documented. pinpoint-prod runs PG17, and its `postgres` role holds `rolbypassrls`+`rolcreaterole` (verified against the mirrored local stack), so the grant succeeds. Added a comment in readonly-role.sql explaining the split — SUPERUSER-only on PG<=15, CREATEROLE+BYPASSRLS on PG16+ — and noting that a pre-16 project would abort loudly here under ON_ERROR_STOP rather than leave a half-built role. —Claude

scripts/query-readonly.mjs
timothyfroehlich: Fixed in e19f7c26. `parseArgs` now hard-errors when both `--file` and an inline query are supplied ("give a query OR --file, not both — --file would win and the inline query would be silently ignored"), matching the loud second-positional handling. Test added. —Claude

scripts/query-readonly.mjs
timothyfroehlich: Declining. Stripping a trailing `# comment` would corrupt a password that legitimately contains `#` (Postgres allows it in the userinfo), turning a robustness nit into a silent truncation of the target — worse than the failure it prevents, which is already loud at connect time. This `.env.local` is also machine-generated by scripts/worktree_setup.py with no inline comments and no `export ` prefix. Left as-is. —Claude

timothyfroehlich:
Claude review of head 6062818 — `/code-review high` — 5 findings: env-file precedence fixed, --file+inline query now rejected, read-only SET comments de-oversold, PG16+ BYPASSRLS note added, env-parser decline documented; all threads resolved

PR #1890, opened 2026-08-16
refactor(agents): cite AGENTS.md sections by title, not number (PP-z9m1)

timothyfroehlich:
Closes PP-z9m1.

## What

AGENTS.md section numbers shift whenever the file gains or loses a section, and nothing tells the citation. Three sites cited `§2.2.5`, which is not a section at all — it is item 5 of a numbered list under `§2.2`, and never had a heading.

This converts ~40 citations across 25 files to the heading title:

```
AGENTS.md §5 "Which tests to run"  ->  AGENTS.md "Which tests to run"
AGENTS.md §7                       ->  AGENTS.md "Deployment"
AGENTS.md §2.2.5                   ->  AGENTS.md "Process rules"
```

and broadens the existing `§2.1`-only gate in `scripts/check_rule_ids.py` to every `AGENTS.md §N`, so the class cannot recur.

**The gate immediately earned it.** It caught five citations in the `` `AGENTS.md` §7 `` form — backtick then space — that my own ripgrep pattern missed. Every quoted title was then verified to resolve against a real heading in AGENTS.md.

Two of my first-pass substitutions were wrong and are fixed here, both worth knowing about as a reviewer:

- A parenthesised gloss (`§7 (canonical endpoint table)`) became a fake title `AGENTS.md "canonical endpoint table"`. A quoted title matching no heading is worse than the number it replaced, because it looks authoritative.
- A blanket `§5 -> "Workflow"` was too coarse — several of those sites mean `"Review comments"` or `"Which tests to run"` specifically, and landing a reader in the wrong subsection is the failure this PR is supposed to prevent.

## Also here

**AGENTS.md's own internal self-references** (`see §4 Process safety`, `see §5 Branches`, …) are converted too. They are the same fragility, and leaving them would make the PR incoherent. The `pinpoint-design-bible` `§5`/`§17` reference on line ~237 is deliberately untouched — those are that spec's sections, not AGENTS.md's. Same for `docs/pbm-listing-redesign-refresher.md`, which numbers its own sections.

**`pinpoint-briefing` gains Group F.** At Tim's request, so the unattended nightly bead routine can hand work off: it reports on the bead it worked, flagged `nightly-report`, and the briefing reads that queue plus the `human` decision queue it builds. Note this producer does not exist yet — it is still a draft prompt outside the repo.

## Sequencing note

PP-z9m1's notes say to do this before "PP-22e4 PR 10 (AGENTS.md → stub)". That PR was superseded and never shipped — PP-22e4's close reason says the stub was deliberately not done. So the deadline is gone; this is durability work, not a race. Recorded on the bead.

## Testing

- `pnpm run check` — green
- `pnpm run check:python` — 442 passed
- New tests cover the multi-level `§2.2.5` case, a bare `§N` with no `AGENTS.md` prefix (must NOT match — the PBM spec's own numbering), and the durable `AGENTS.md "Title"` form being accepted

## Known gap

The gate bans section *numbers* but does not verify that a quoted *title* still resolves to a real heading. Renaming a heading would silently strand its citations. I verified all of them by hand this time; closing it in the gate is a follow-up worth filing if you want it.

https://claude.ai/code/session_01HshAyr2EZe5ZcSk95GfGek

### DIFF
--- a/.agents/rules/antigravity.md
+++ b/.agents/rules/antigravity.md
@@ -10,7 +10,7 @@ Antigravity is Google's CLI agent harness (currently Gemini-based) with full loc
 
 ## Core Mandates
 
-- **Read AGENTS.md**: Immediately read @AGENTS.md before following any user instructions. It carries the process rules, environment, and workflow. The implementation non-negotiables are **not** in it — `docs/NON_NEGOTIABLES.md` is the catalog, and §2.1 is now only a pointer to it (PP-22e4.4 moved the per-rule summaries into `.claude/rules/`, which only Claude Code loads). Read the catalog directly.
+- **Read AGENTS.md**: Immediately read @AGENTS.md before following any user instructions. It carries the process rules, environment, and workflow. The implementation non-negotiables are **not** in it — `docs/NON_NEGOTIABLES.md` is the catalog, and "Implementation" is now only a pointer to it (PP-22e4.4 moved the per-rule summaries into `.claude/rules/`, which only Claude Code loads). Read the catalog directly.
 - **Read the review brief**: @REVIEW.md is the canonical code-review rubric, shared with Claude Code — read it before reviewing any PR.
 - **You review when Tim asks — never on your own.** Nothing on this repo reviews on PR-open or on push; the bot reviewer that used to was retired on 2026-08-02 (PP-4ric).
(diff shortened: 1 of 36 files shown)


timothyfroehlich:
Closing — this was a side quest. It came out of a test run of the nightly-bead routine and grew well past what was asked for. The branch is gone; PP-z9m1 goes back on the backlog unclaimed.

